Issue #20, opened 2023-02-17
Section for raw config for custom plugins


faisal-memon:
Standard plugins should follow standard helm yaml syntax. To support custom plugins users have created we need a section where they can enter raw config for these plugins.

kfox1111:
Partially implemented in https://github.com/spiffe/helm-charts/pull/198

Still need the agent equiv.

Issue #16, opened 2023-02-16
Make workload attestors configurable


faisal-memon:
It may be beneficial to the user to be able to enable what workload attestors they want to use:

```
  workloadAttesters:
    k8s:
      enabled: false
    unix:
      enabled: true
```

Suggested by @kfox1111 here: https://github.com/spiffe/helm-charts/pull/14#discussion_r1107854544

PR #35, opened 2023-02-21
WIP: Plugins

kfox1111:
Rough idea of how we could enable mapping of chart values into plugin configs.

Lets discuss and see if we want to move forward with it. If so, I can do the WorkloadAttestors too along with the spire-agent side.

Partially addresses: https://github.com/spiffe/helm-charts/issues/20
Partially addresses: https://github.com/spiffe/helm-charts/issues/16


kfox1111:
Closing in favor of https://github.com/spiffe/helm-charts/pull/49

PR #46, opened 2023-02-24
Enable nested spire use case

kfox1111:
This patch enables the nested spire use case:

You can have a root spire server external to Kubernetes that is the base spiffe ca.
Then each of your kubernetes clusters can run an instance of this chart bound to the root server and all will be trusted.


kfox1111:
Closing here for now... keeping it open here:
https://github.com/kfox1111/spiffe-helm-charts/pull/2

Issue #52, opened 2023-02-26
extra init containers and volumes


kfox1111:
The agent and server needs the option to add extra mounts and init containers.

PR #67, opened 2023-02-28
Add extra initContainers, containers, volumes to agent and server

kfox1111:
With plugin support, agents and servers need more customization. This patch enables initContainers, extraContainers, extraVolumes and extraVolumeMounts to be added to those services.

Implements: https://github.com/spiffe/helm-charts/issues/52

### DIFF
--- a/charts/spire/charts/spire-agent/README.md
+++ b/charts/spire/charts/spire-agent/README.md
@@ -12,6 +12,9 @@ A Helm chart to install the SPIRE agent.
 |-----|------|---------|-------------|
 | bundleConfigMap | string | `"spire-bundle"` |  |
 | clusterName | string | `"example-cluster"` |  |
+| extraContainers | list | `[]` |  |
+| extraVolumeMounts | list | `[]` |  |
+| extraVolumes | list | `[]` |  |
 | fullnameOverride | string | `""` |  |
 | healthChecks.port | int | `9980` | override the host port used for health checking |
 | image.pullPolicy | string | `"IfNotPresent"` |  |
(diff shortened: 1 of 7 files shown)


charts/spire/charts/spire-agent/README.md
```diff
@@ -12,13 +12,17 @@ A Helm chart to install the SPIRE agent.
 |-----|------|---------|-------------|
 | bundleConfigMap | string | `"spire-bundle"` |  |
 | clusterName | string | `"example-cluster"` |  |
+| extraContainers | list | `[]` |  |
```
dennisgove, line 15: Are these acting as sidecar containers? Is `extra` a standard nomenclature for a setup like this?

charts/spire/charts/spire-agent/README.md
```diff
@@ -12,13 +12,17 @@ A Helm chart to install the SPIRE agent.
...
+| extraVolumeMounts | list | `[]` |  |
+| extraVolumes | list | `[]` |  |
```
dennisgove, line 17: The use of `extra` here feels like these items, if set, would be superfluous and not needed but I imagine if someone is setting these they're doing so because the volumes and mounts are very much needed. Is there a better term we can use other than `extra`? 

`additional` comes to mind.

charts/spire/charts/spire-server/README.md
```diff
@@ -42,12 +42,16 @@ A Helm chart to install the SPIRE server.
 | dataStorage.enabled | bool | `true` |  |
 | dataStorage.size | string | `"1Gi"` |  |
 | dataStorage.storageClass | string | `nil` |  |
+| extraContainers | list | `[]` |  |
+| extraVolumeMounts | list | `[]` |  |
+| extraVolumes | list | `[]` |  |
```
dennisgove, line 47: Same comments about the `extra` term; is there something better fitting?

charts/spire/charts/spire-agent/README.md
kfox1111: Yes.

Extra for stuff is a pretty common pattern going way way back. For example:
https://github.com/helm/charts/blob/master/stable/fluent-bit/values.yaml#L202-L207

charts/spire/charts/spire-server/README.md
dennisgove: I have no concern if this is a common convention.

dennisgove (approved):

PR #68, opened 2023-02-28
Add support for spire-server ingress

### LINKED ISSUE #61: Ingress support for spire-server
There should be an optional ingress object for the spire server so it can be exposed outside the cluster.

kfox1111:
This patch enables the spire-server to be exposed outside the cluster utilizing an ingress controller that supports ssl passthrough such as ingress-nginx. It also adds tests to ensure that an agent can reach it through an ingress.

Resolves #61

### DIFF
--- a/charts/spire/charts/spire-server/templates/ingress.yaml
+++ b/charts/spire/charts/spire-server/templates/ingress.yaml
@@ -1,4 +1,16 @@
+{{- if .Values.ingress.enabled -}}
 {{- $fullName := include "spire-server.fullname" . -}}
-{{- $namespace := include "spire-server.namespace" . }}
-{{- $labels := include "spire-server.labels" . }}
-{{ include "spire-lib.ingress-template" (dict "ingress" .Values.ingress "name" $fullName "namespace" $namespace "svcName" $fullName "port" .Values.service.port "labels" $labels) }}
+apiVersion: networking.k8s.io/v1
+kind: Ingress
+metadata:
+  name: {{ $fullName }}
+  namespace: {{ include "spire-server.namespace" . }}
+  labels:
+    {{ include "spire-server.labels" . | nindent 4}}
+  {{- with .Values.ingress.annotations }}
+  annotations:
+    {{- toYaml . | nindent 4 }}
+  {{- end }}
+spec:
+  {{ include "spire-lib.ingress-spec" (dict "ingress" .Values.ingress "svcName" $fullName "port" .Values.service.port "global" .Values.global) | nindent 2 }}
+{{- end }}
--- a/charts/spire/charts/spiffe-oidc-discovery-provider/templates/ingress.yaml
+++ b/charts/spire/charts/spiffe-oidc-discovery-provider/templates/ingress.yaml
@@ -1,4 +1,16 @@
+{{- if .Values.ingress.enabled -}}
 {{- $fullName := include "spiffe-oidc-discovery-provider.fullname" . }}
-{{- $namespace := include "spiffe-oidc-discovery-provider.namespace" . }}
-{{- $labels := include "spiffe-oidc-discovery-provider.labels" . }}
-{{ include "spire-lib.ingress-template" (dict "ingress" .Values.ingress "name" $fullName "namespace" $namespace "svcName" $fullName "port" .Values.service.port "labels" $labels) }}
+apiVersion: networking.k8s.io/v1
+kind: Ingress
+metadata:
+  name: {{ $fullName }}
+  namespace: {{ include "spiffe-oidc-discovery-provider.namespace" . }}
+  labels:
+    {{ include "spiffe-oidc-discovery-provider.labels" . | nindent 4 }}
+  {{- with .Values.ingress.annotations }}
+  annotations:
+    {{- toYaml . | nindent 4 }}
+  {{- end }}
+spec:
+  {{ include "spire-lib.ingress-spec" (dict "ingress" .Values.ingress "svcName" $fullName "port" .Values.service.port "global" .Values.global) | nindent 2 }}
+{{- end }}
--- a/charts/spire/charts/spire-server/templates/federation-ingress.yaml
+++ b/charts/spire/charts/spire-server/templates/federation-ingress.yaml
@@ -1,7 +1,18 @@
 {{- if .Values.federation.enabled }}
+{{- if .Values.federation.ingress.enabled -}}
 {{- $svcName := include "spire-server.fullname" . }}
-{{- $fullName := printf "%s-federation" $svcName }}
-{{- $namespace := include "spire-server.namespace" . }}
-{{- $labels := include "spire-server.labels" . }}
-{{ include "spire-lib.ingress-template" (dict "ingress" .Values.federation.ingress "name" $fullName "namespace" $namespace "svcName" $svcName "port" .Values.federation.bundleEndpoint.port "labels" $labels) }}
+apiVersion: networking.k8s.io/v1
+kind: Ingress
+metadata:
+  name: {{ printf "%s-federation" $svcName }}
+  namespace: {{ include "spire-server.namespace" . }}
+  labels:
+    {{ include "spire-server.labels" . | nindent 4 }}
+  {{- with .Values.ingress.annotations }}
+  annotations:
+    {{- toYaml . | nindent 4 }}
+  {{- end }}
+spec:
+  {{ include "spire-lib.ingress-spec" (dict "ingress" .Values.federation.ingress "svcName" $svcName "port" .Values.federation.bundleEndpoint.port "global" .Values.global) | nindent 2 }}
+{{- end }}
 {{- end }}
--- a/charts/spire/templates/_spire-lib.tpl
+++ b/charts/spire/templates/_spire-lib.tpl
@@ -55,55 +55,38 @@
 
 {{/* Takes in a dictionary with keys:
  * ingress - the standardized ingress object
- * name - The standardized object name
- * namespace - The namespace to create the object in
  * svcName - The service to route to
  * port - which port on the service to use
- * labels - lables to add to the ingress
 */}}
-{{ define "spire-lib.ingress-template" }}
+{{ define "spire-lib.ingress-spec" }}
 {{- $svcName := .svcName }}
 {{- $port := .port }}
-{{- if .ingress.enabled -}}
-apiVersion: networking.k8s.io/v1
-kind: Ingress
-metadata:
-  name: {{ .name }}
-  namespace: {{ .namespace }}
-  labels:
-    {{ .labels | nindent 4 }}
-  {{- with .ingress.annotations }}
-  annotations:
-    {{- toYaml . | nindent 4 }}
-  {{- end }}
-spec:
-  {{- with .ingress.className }}
-  ingressClassName: {{ . | quote }}
-  {{- end }}
-  {{- if .ingress.tls }}
-  tls:
-    {{- range .ingress.tls }}
-    - hosts:
-        {{- range .hosts }}
-        - {{ . | quote }}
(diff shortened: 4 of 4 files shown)


marcofranssen:
My main concern is how we can guarantee we don't introduce vulnerabilities from allowing to expose insecure ingress configurations. Maybe we should also consider a opiniated out of the box secure setup. I'm fine with both, but I think we need to discuss this and agree collectively on the solution design.

kfox1111:
> My main concern is how we can guarantee we don't introduce vulnerabilities from allowing to expose insecure ingress configurations. Maybe we should also consider a opiniated out of the box secure setup. I'm fine with both, but I think we need to discuss this and agree collectively on the solution design.

Hmm..... Yeah, I understand trying to save the user from themselves.... how to do it though, since that part of the k8s api was never standardized....

We could force the annotations in unless they disable them, but its still not guaranteed another ingress controller would honor them. in fact, aws seems to do it quite differently: https://kubernetes-sigs.github.io/aws-load-balancer-controller/v2.2/guide/tasks/ssl_redirect/#example-ingress-manifest

It would be a pretty ugly game of wack a mole...

One option might be to need an extra value when enabling ingress that states: IHaveReadTheInstructions:true before it allows the chart to be installed with ingress.enabled true? Then they have the opportunity to review the instructions, and if they set it without looking, its on them?

kfox1111:
gotta do this sort of thing: https://github.com/spiffe/helm-charts/pull/68/files#diff-8d33a0cb93c24fe8f2967716c7f6caef6fc2ba4854597ecf6cb7ec0d1f2c46ebR6-R7

.github/tests/spire-oidc-insecure/values.yaml
```diff
@@ -1,3 +1,21 @@
+spire-server:
+  ingress:
+    enabled: true
+    className: "nginx"
+    annotations:
+      nginx.ingress.kubernetes.io/backend-protocol: "HTTPS"
+      nginx.ingress.kubernetes.io/ssl-passthrough: "true"
```
marcofranssen, line 7:
```suggestion
      nginx.ingress.kubernetes.io/backend-protocol: "HTTPS"
      nginx.ingress.kubernetes.io/ssl-passthrough: "true"
      nginx.ingress.kubernetes.io/ssl-redirect: "true"
      nginx.ingress.kubernetes.io/force-ssl-redirect: "true"
```

If we are going to add annotations here anyhow, I would also add the 2 I suggest here. That will improve user experience if people don't specify `https` explicitly they will be auto-redirected.

.github/tests/spire-oidc-insecure/values.yaml
```diff
@@ -1,3 +1,23 @@
...
+      nginx.ingress.kubernetes.io/ssl-redirect: "true"
+      nginx.ingress.kubernetes.io/force-ssl-redirect: "true"
+    hosts:
+    - host: ingress-nginx-controller.spire-oidc-insecure
```
faisal-memon, line 11: Should this reference spire server since that is the target?

kfox1111: The ingress controller that the test is running in is in the spire-oidc-insecure namespace. so that is a valid hostname to use for testing against.

faisal-memon:
> Yeah, difficult topic, I think for this PR due to the SNI part it is less of an issue. I think we can still add the annotations I suggested here and then merge it if other maintainers agree.

> @faisal-memon what is your thoughts on this? Or should we check with the SSC?

I agree with adding the annotations. We should make this as secure as possible. Don't want someone's spire server getting compromised and pointing at the helm chart as a cause. We can add a comment that other ingress controllers might not work with this config and we only officially support nginx.

kfox1111:
> > Yeah, difficult topic, I think for this PR due to the SNI part it is less of an issue. I think we can still add the annotations I suggested here and then merge it if other maintainers agree.
> 
> > @faisal-memon what is your thoughts on this? Or should we check with the SSC?
> 
> I agree with adding the annotations. We should make this as secure as possible. Don't want someone's spire server getting compromised and pointing at the helm chart as a cause. We can add a comment that other ingress controllers might not work with this config and we only officially support nginx.

The other ingress merged without the attributes but with them commented. There is also the PR here:
https://github.com/spiffe/helm-charts/pull/136/files#diff-12da2c930a5613e4666455fd13c933a6c02e6aa109e39693a84a5b263407b944
That adds a production config for ingress-nginx so the user can't misconfigure it.

I suggest we rebase this PR on that PR and make sure there is a production example with the annotations there.

We also have a contributor using istio instead of ingress-nginx.

faisal-memon:
Thinking about this one and how it relates to the [same thing done with the oidc](https://github.com/spiffe/helm-charts/commit/cc7121e021cf8f569ced4a6dd767f9fdf5c32da9). If both are deployed will we be the deploying 2 ingress controllers?

Im wondering if a better way to do this would be to do this would be for people to use the [official ingress nginx chart](https://artifacthub.io/packages/helm/ingress-nginx/ingress-nginx) and provide the config to link with that. Similar to how we do with external databases. Then you can just have a single nginx to expose whatever ingress you want.

Im also concerned about maintaining an embedded nginx subchart and having to maintain versions, testing, api changes, etc.

faisal-memon:
> Thinking about this one and how it relates to the [same thing done with the oidc](https://github.com/spiffe/helm-charts/commit/cc7121e021cf8f569ced4a6dd767f9fdf5c32da9). If both are deployed will we be then deploying 2 ingress controllers?
> 
> Im wondering if a better way to do this would be to do this would be for people to use the [official ingress nginx chart](https://artifacthub.io/packages/helm/ingress-nginx/ingress-nginx) and provide the config to link with that. Similar to how we do with external databases. Then you can just have a single nginx to expose whatever ingress you want.
> 
> Im also concerned about maintaining an embedded nginx subchart and having to maintain versions, testing, api changes, etc.

Nm, disregard the above.  We are just deploying the ingress resource, not the controller.

charts/spire/charts/spire-server/values.yaml
```diff
@@ -96,6 +96,26 @@ federation:
     port: 8443
     address: "0.0.0.0"
 
+  ingress:
+    enabled: false
+    className: ""
```
faisal-memon, line 101: Maybe have this be `spire-federation` or something similar?

charts/spire/charts/spire-server/values.yaml
```diff
@@ -201,6 +221,24 @@ telemetry:
       namespace: ""
       labels: {}
 
+ingress:
+  enabled: false
+  className: ""
```
faisal-memon, line 226: Maybe `spire-server`?

faisal-memon:
Overall this looks good to me. There is some duplication with the 2 ingress templates? Is there any ways to condense that down with a helper?

charts/spire/charts/spire-server/values.yaml
kfox1111: For enabled?

Its hard to read in the patch, but its already spire-server.federation.ingress.enabled.

charts/spire/charts/spire-server/values.yaml
kfox1111: For enabled?

Its hard to read in the patch, but its already spire-server.ingress.enabled.

For this one, there is a slight chance it could be confused with the other ingress(s) in the chart. so 
spire-server.server.ingress.enabled could disambiguate. but may also be more confusing then its worth?

charts/spire/charts/spire-server/templates/ingress.yaml
```diff
@@ -0,0 +1,87 @@
...
+          {{- range .paths }}
+          - path: {{ .path }}
+            pathType: {{ .pathType }}
+            backend:
+              service:
+                name: {{ $fullName }}
+                port:
+                  number: {{ $fedPort }}
+          {{- end }}
+    {{- end }}
```
marcofranssen, line 86: This part can probably be a macro to be reused by both the ingresses?

charts/spire/charts/spire-server/values.yaml
marcofranssen: image

When reviewing use these buttons to see more of the context. to better judge.

charts/spire/charts/spire-server/templates/ingress.yaml
kfox1111: Yeah. I think I can make it a macro. So far each of them is pretty similar...

I'll take a stab at it.

examples/production/README.md
```diff
@@ -12,3 +12,22 @@ helm upgrade --install --namespace spire-server spire charts/spire -f values.yam
 ```
 
 See [values.yaml](./values.yaml) for more details on the chart configurations to achieve this setup.
+
+If you want to expose your spire-server outside of Kubernetes and are using ingress-nginx add
```
marcofranssen, line 16:
```suggestion
If you want to expose your spire-server outside of Kubernetes and are using ingress-nginx, add following values file when running `helm template/install/upgrade`.

```

examples/production/README.md
```diff
@@ -12,3 +12,22 @@ helm upgrade --install --namespace spire-server spire charts/spire -f values.yam
...
+```shell
+-f values-expose-spire-server-ingress-nginx.yaml
+```
+
+If you want to expose your federation endpoint outside of Kubernetes and are using ingress-nginx
+you have two options as described here:
+https://github.com/spiffe/spiffe/blob/main/standards/SPIFFE_Federation.md#52-endpoint-profiles
+
+If you chose profile https_web, use:
```
marcofranssen, line 25:
```suggestion
If you chose profile https_web, use:

```

examples/production/README.md
```diff
@@ -12,3 +12,22 @@ helm upgrade --install --namespace spire-server spire charts/spire -f values.yam
...
+```shell
+-f values-expose-federation-https-web-ingress-nginx.yaml
+```
+
+If you chose profile https_spiffe, use:
```
marcofranssen, line 30:
```suggestion
If you chose profile https_spiffe, use:

```

charts/spire/charts/spire-server/values.yaml
marcofranssen: Agree to leave it as is, we are already in the context of the spire-server chart.

marcofranssen (changes requested):
Few suggestions.

charts/spire/charts/spire-server/templates/ingress.yaml
kfox1111: Done

charts/spire/charts/spire-server/values.yaml
faisal-memon: referring specifically to the className default value

kfox1111: ClassName is an ingress property. Unrelated to this chart. Needs to be configurable depending on what ingress classes are available on the users cluster.

mrsabath:
Once the ingress support is merged, we should work on instruction documentation on helping setup ingress in major cloud providers in context of the SPIRE deployment. See the issue #293

charts/spire/charts/spire-server/values.yaml
faisal-memon: Yes, so do we want to give it a default value other than `""`?

kfox1111: No. Blank should be default. If blank, no ingress class will be specified. This allows it to function properly on clusters where the cluster-admin has specified a default ingress class for the cluster for the users. Actual class names are not standardized so we can't set it to anything in particular thats useful.

Some details here:
https://kubernetes.io/docs/concepts/services-networking/ingress/#default-ingress-class

faisal-memon (approved):

marcofranssen (approved):
LGTM

Issue #22, opened 2023-02-17
Support configurable CA and workload TTLs


faisal-memon:
Support the following SPIRE Server configurables:

- ca_ttl  - The default CA/signing key TTL | 24h
- default_x509_svid_ttl  - The default X509-SVID TTL
- default_jwt_svid_ttl -The default JWT-SVID TTL

Reference [SPIRE Server configuration](https://github.com/spiffe/spire/blob/main/doc/spire_server.md)

Issue #18, opened 2023-02-16
Support configurable CA key type


faisal-memon:
Reference configurable on spire server: https://github.com/spiffe/spire/blob/main/doc/spire_server.md#:~:text=ca_key_type,by%20jwt_key_type)

PR #70, opened 2023-03-01
Add support for custom plugins

faisal-memon:
Allow for config for custom plugins to be embedded via a multi line string.

fixes https://github.com/spiffe/helm-charts/issues/20

### DIFF
--- a/charts/spire/charts/spire-agent/templates/configmap.yaml
+++ b/charts/spire/charts/spire-agent/templates/configmap.yaml
@@ -42,6 +42,8 @@ data:
           }
       }
       {{- end }}
+
+{{ .Values.customPluginConfig | indent 6 }}
     }
 
     health_checks {
--- a/charts/spire/charts/spire-agent/README.md
+++ b/charts/spire/charts/spire-agent/README.md
@@ -12,6 +12,7 @@ A Helm chart to install the SPIRE agent.
 |-----|------|---------|-------------|
 | bundleConfigMap | string | `"spire-bundle"` |  |
 | clusterName | string | `"example-cluster"` |  |
+| customPluginConfig | string | `""` | Config for custom plugins, make sure to indent each line by 2 spaces |
 | fullnameOverride | string | `""` |  |
 | healthChecks.port | int | `9980` | override the host port used for health checking |
 | image.pullPolicy | string | `"IfNotPresent"` |  |
--- a/charts/spire/charts/spire-agent/values.yaml
+++ b/charts/spire/charts/spire-agent/values.yaml
@@ -80,6 +80,14 @@ workloadAttestors:
     # -- enables the Unix workload attestor
     enabled: false
 
+# -- Config for custom plugins, make sure to indent each line by 2 spaces
+customPluginConfig: |-
+#   NodeAttestor "my-custom-attestor" {
+#     plugin_data {
+#       my-data: "test"
+#     }
+#   }
+
 telemetry:
   prometheus:
     enabled: false
--- a/charts/spire/charts/spire-server/README.md
+++ b/charts/spire/charts/spire-server/README.md
@@ -38,6 +38,7 @@ A Helm chart to install the SPIRE server.
 | controllerManager.service.annotations | object | `{}` |  |
 | controllerManager.service.port | int | `443` |  |
 | controllerManager.service.type | string | `"ClusterIP"` |  |
+| customPluginConfig | string | `""` | Config for custom plugins, make sure to indent each line by 2 spaces |
 | dataStorage.accessMode | string | `"ReadWriteOnce"` |  |
 | dataStorage.enabled | bool | `true` |  |
 | dataStorage.size | string | `"1Gi"` |  |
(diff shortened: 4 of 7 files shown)


marcofranssen (commented):
Wouldn't it be better to do some more explicit things? 

```shell
additionalNodeAttestors:

additionalXXX:
```

This could allow us to more strongly type the values, to guide the users better.

kfox1111:
Draft this please.

Its a conflict of approach with https://github.com/spiffe/helm-charts/pull/35

kfox1111:
To be simpler, in context, we need solutions to:
https://github.com/spiffe/helm-charts/issues/22
https://github.com/spiffe/helm-charts/issues/18

along with any other settings that are hard coded.

Also, this is probably a post 1.0 thing, since the alternative approach is being pushed out past 1.0.

charts/spire/charts/spire-agent/templates/configmap.yaml
```diff
@@ -42,6 +42,8 @@ data:
           }
       }
       {{- end }}
+
+{{ .Values.customPluginConfig | indent 6 }}
```
marcofranssen, line 46:
```suggestion
      {{- .Values.customPluginConfig | indent 6 }}
```

kfox1111:
> Wouldn't it be better to do some more explicit things?
> 
> ```shell
> additionalNodeAttestors:
> 
> additionalXXX:
> ```
> 
> This could allow us to more strongly type the values, to guide the users better.

It would be better yet to strongly type the whole thing with yaml IMO.

After some conversation, I'm ok merging this as is with the understanding that it will be perfectly reasonable to revisit this issue post 1.0. And its reasonable if its a breaking change for users. We will make an overarching issue to discuss the best way to layout custom plugins in the longer run, weight the pros/cons, and come to a consensus.

faisal-memon:
@marcofranssen Was able to address your comment. The custom plugins can be in yaml rather than embedded hcl. So now this:

```yaml
customNodeAttestors:
  attestor1:
    a: true
    b: 5
    c: "custom"
    d:
      e:
        f: "true"
        g: true
```

will render to this: 
```
      NodeAttestor "attestor1"  {
        plugin_data {
          a = true
          b = 5
          c = "custom"
          "d" = {
            "e" = {
              f = "true"
              g = true
            }
          }
        }
      }
    }
```

Needs some more testing but just want to check in make sure you're ok with the general design before moving foward.

kfox1111:
That works well for simple stuff. And maybe all the node plugins... Not sure.

It doesn't work for named chunks in hcl like this kind of thing:
https://spiffe.io/docs/latest/architecture/federation/readme/#configure-trust-bundle-location-using-spiffe-authentication-https_spiffe
see the federates_with and bundle_endpoint_profile lines...

so there are some hcl representations that can't be done with yaml/json -> hcl. but could be done when spire is reading json natively.

kfox1111:
Can you please draft this since it needs work?

The approach though is a good balance between the raw blob of hcl and the yaml->json one. might hit the 80% good enough without having to switch to json...

Going to think about this approach for a bit.

faisal-memon:
> It doesn't work for named chunks in hcl like this kind of thing: https://spiffe.io/docs/latest/architecture/federation/readme/#configure-trust-bundle-location-using-spiffe-authentication-https_spiffe see the federates_with and bundle_endpoint_profile lines...
> 

I think we can catch that as the type would be `map[string]interface{}`

kfox1111:
Cant easily determine if this yaml:
```
foo:
  bar:
    other: thing
```

should be this hcl
```
foo {
  bar {
    other=thing
```
or this
```
foo "bar" {
  other=thing
```

faisal-memon:
> Cant easily determine if this yaml:
> 
> ```
> foo:
>   bar:
>     other: thing
> ```
> 
> should be this hcl
> 
> ```
> foo {
>   bar {
>     other=thing
> ```
> 
> or this
> 
> ```
> foo "bar" {
>   other=thing
> ```

What if its like a triple or quadruple nest? Kind of leaning towards just having the embedded hcl after going through this exercise. Not sure if i want to get into the general purpose yaml -> hcl/json conversion business.

kfox1111:
yeah, its gets mostly unusable from there. :/

I'm pondering how common this is:
```
foo {
  bar {
    other=thing
```

If it covers the 80%, that may be enough to just do that for now and revisit the json native if enough users want to support all the things. Need to think on that one...

kfox1111:
Ok. So had a chat with Andrew.

In HCL, these are actually equiv config in spire:
```
foo {
  bar {
    other=thing
```

```
foo "bar" {
  other=thing
```

And the former format is more compatible with hcl2. So we should probably just always map it to:
```
foo {
  bar {
    other=thing
```

marcofranssen:
Probably rebase this PR on top of #113 

My PR changes the existing config as is to yaml. Yours could build on top of that.

However also see the description on my PR. We need a release of spire 1.6.2 before we can merge it.

faisal-memon:
closing for now, will reopen when we move to json.

PR #71, opened 2023-03-01
Use quote function instead of actual quotes

faisal-memon:
Maintaining consistency using the `quote` function instead of wrapping values in quotes. The quote function works better for syntax highlighting.


marcofranssen (approved):
LGTM :rocket:

PR #77, opened 2023-03-02
Add k8s 1.26.0 (Kind) to the test matrix

marcofranssen:
Question to maintainers:

Should we keep 1.21 in the test matrix for now, or shall we drop that from the matrix? E.g. limit the tests to max 5 versions.

Officially we only support the last 3 k8s releases as stated in our README.md.

Currently our chart is perfectly compatible with `1.21`.

Personally I think we can leave `1.21` around until we build features that are incompatible with `1.21`.

See https://github.com/kubernetes-sigs/kind/releases/tag/v0.17.0 for more details.


kfox1111:
I'd benefit from a 1.21 test while it takes little effort to maintain. If it does become burdensome we can drop it?

kfox1111 (approved):

Issue #69, opened 2023-02-28
Configmap update rollout


kfox1111:
Ensure all configmaps have corresponding annotations on the pods to enable automatic rollouts.

PR #81, opened 2023-03-02
Make sure all configmap changes rollout

kfox1111:
When a change happens to a configmap, kubernetes does not automatically restart any pods that make use of it. This is confusing to users because they think they have made a change, only for the workload to ignore it until the user manually deletes the pods. This change follows the standard helm workaround of annotating the pods with a checksum of the configmaps so that a configmap change becomes visible as a pod change to kubernetes causing it to redeploy the affected workloads properly and automatically during a helm upgrade.

Fixes: https://github.com/spiffe/helm-charts/issues/69


faisal-memon:
Hi @kfox1111 Can you add some more context? What problem is this fixing?

kfox1111:
> Hi @kfox1111 Can you add some more context? What problem is this fixing?

I just updated the pr description with more context.

marcofranssen (approved):
LGTM :rocket:

PR #82, opened 2023-03-03
UpstreamAuthority cert-manager support

### LINKED ISSUE #53: Add support for cert-manager UpstreamAuthority
The spire-server can use a ca from cert-manager.

See
https://github.com/spiffe/spire/blob/main/doc/plugin_server_upstreamauthority_cert_manager.md

See following where to add:

[main/charts/spire/charts/spire-server/values.yaml#L94-L106](https://github.com/spiffe/helm-charts/blob/main/charts/spire/charts/spire-server/values.yaml?rgh-link-date=2023-03-02T10%3A01%3A09Z#L94-L106)
[main/charts/spire/charts/spire-server/templates/configmap.yaml#L61-L73](https://github.com/spiffe/helm-charts/blob/main/charts/spire/charts/spire-server/templates/configmap.yaml?rgh-link-date=2023-03-02T10%3A01%3A09Z#L61-L73)

kfox1111:
Enable utilizing cert-manager as the upstream ca for the spire-server.

resolves #53


marcofranssen:
@kfox1111 please use `resolves #53` in the description of this issue so the issue autocloses when the PR gets merged.

.github/tests/upstream-authority-cert-manager/pre-install.sh
```diff
@@ -0,0 +1,4 @@
+#!/bin/bash
```
marcofranssen, line 1:
```suggestion
#!/usr/bin/env bash

```

.github/tests/upstream-authority-cert-manager/cert_manager_ca.yaml
```diff
@@ -0,0 +1,31 @@
+---
```
marcofranssen, line 1: can we use the `kebab-style-filenaming` instead of `snake_case` like we do for other files as well.

.github/tests/upstream-authority-cert-manager/values.yaml
```diff
@@ -0,0 +1,7 @@
+spire-server:
+  upstreamAuthority:
+    disk:
+      enabled: false
```
marcofranssen, line 4:
```suggestion
```

This is the default already, so can be removed.

.github/workflows/helm-chart-ci.yaml
```diff
@@ -151,9 +151,15 @@ jobs:
...
+          [ "$VALUES" != "default" ] && kubectl create namespace "$VALUES"
+          [ -x .github/tests/$VALUES/pre-install.sh ] && .github/tests/$VALUES/pre-install.sh
           ct install --debug \
+            --namespace $VALUES \
             --target-branch ${{ github.base_ref }} \
             --exclude-deprecated \
             ${{ (matrix.values != 'default' && '--helm-extra-set-args "--values=.github/tests/$VALUES/values.yaml"') || '' }}
+          res=$?
+          [ -x .github/tests/$VALUES/post-install.sh ] && .github/tests/$VALUES/post-install.sh
+          exit $res
```
marcofranssen, line 163: I would implement this using a `trap` to make it less errorprone

https://medium.com/@dirk.avery/the-bash-trap-trap-ce6083f36700

charts/spire/charts/spire-server/values.yaml
```diff
@@ -104,6 +104,16 @@ upstreamAuthority:
...
         bundle: ""
+  certManager:
+    enabled: false
+    rbac:
+      create: true
+    issuer_name: "spire-ca"
+    issuer_kind: "Issuer"
+    issuer_group: "cert-manager.io"
+    # -- Specify to use a namespace other then the one the chart is installed into
+    namespace: ""
```
marcofranssen, line 115: Is this the namespace certmanager is installed into? Should we default this to `cert-manager`, I think that is where all cert-manager tutorials also put it.

charts/spire/charts/spire-server/values.yaml
```diff
@@ -104,6 +104,16 @@ upstreamAuthority:
         certificate: ""
         key: ""
         bundle: ""
+  certManager:
+    enabled: false
+    rbac:
+      create: true
+    issuer_name: "spire-ca"
+    issuer_kind: "Issuer"
+    issuer_group: "cert-manager.io"
```
marcofranssen, line 113: Will there be any usecase where the issuer_kind and issuer_group need to be changed? If not I would leave it out.

charts/spire/charts/spire-server/templates/roles.yaml
```diff
@@ -18,6 +18,19 @@ rules:
...
+  {{- if and .Values.upstreamAuthority.certManager.enabled .Values.upstreamAuthority.certManager.rbac.create }}
+  - apiGroups: ["cert-manager.io"]
+    resources:
+      - certificates
+      - certificaterequests
+    verbs:
+      - list
+      - get
+      - create
+      - delete
```
marcofranssen, line 30: Should it be possible to delete certificates by spire-server?

marcofranssen (changes requested):
Few small suggestions and questions. Looks pretty neat this one.

.github/tests/upstream-authority-cert-manager/values.yaml
kfox1111: oh. Didn't notice that. thanks.

charts/spire/charts/spire-server/values.yaml
kfox1111: no. its the namespace to register the cert requests into. if not set, it does so in the same namespace as spire. which is right most of the time. It was in the docs as an option though so I added it too.

charts/spire/charts/spire-server/values.yaml
kfox1111: Not sure. I can't think of one off hand, but was just making it as configurable as the plugin is in spire. I figure if they made it configurable there, they must have had a reason. But I could be convinced otherwise.

If you still want me to pull it, please let me know and I'll do so.

charts/spire/charts/spire-server/templates/roles.yaml
kfox1111: not sure... Now that we have a test suite, I could try removing it and see what happens.

charts/spire/charts/spire-server/values.yaml
marcofranssen: @faisal-memon what is your take on this? Do you have more background info on why this is configurable on the plugin?

My take would be to hardcode those values in the chart. But I might be missing the usecase of this.

charts/spire/charts/spire-server/templates/roles.yaml
marcofranssen: Probably we should verify which exact permissions the plugin requires, so we can narrow it down as far as possible, without breaking functionality.

charts/spire/charts/spire-server/values.yaml
kfox1111: oh... I know of at least one reason....

issuer_kind can be either Issuer or ClusterIssuer.

charts/spire/charts/spire-server/templates/roles.yaml
kfox1111: It has to get an intermediate ca cert from cert-manager in order to give the oidc pod a cert. So if thats all working in tests, I think we're ok.

It does seem to work without certificate permissions, so I think this has been fixed?

charts/spire/charts/spire-server/values.yaml
kfox1111: Actually, I can think of a common one that users may need after re-review:
issuer_kind: "ClusterIssuer"

.github/workflows/helm-chart-ci.yaml
kfox1111: Done

marcofranssen (approved):
LGTM 🚀

PR #83, opened 2023-03-03
Add a flag to configure the Kubernetes NodeAttestor

kfox1111:
If deploying with alternate NodeAttestors you may want to disable the Kubernetes one. Also a small bug with using a non unique name with cluster wide resources ClusterRole/ClusterRoleBinding is fixed.

### DIFF
--- a/charts/spire/charts/spire-server/README.md
+++ b/charts/spire/charts/spire-server/README.md
@@ -62,8 +62,8 @@ A Helm chart to install the SPIRE server.
 | jwtIssuer | string | `"oidc-discovery.example.org"` |  |
 | logLevel | string | `"info"` |  |
 | nameOverride | string | `""` |  |
-| nodeAttestor.k8s.enabled | bool | `true` |  |
-| nodeAttestor.k8s.rbac.create | bool | `true` |  |
+| nodeAttestor.k8s.psat.enabled | bool | `true` |  |
+| nodeAttestor.k8s.psat.rbac.create | bool | `true` |  |
 | nodeSelector | object | `{}` |  |
 | podAnnotations | object | `{}` |  |
 | podSecurityContext | object | `{}` |  |
--- a/charts/spire/charts/spire-server/README.md
+++ b/charts/spire/charts/spire-server/README.md
@@ -62,8 +62,8 @@ A Helm chart to install the SPIRE server.
 | jwtIssuer | string | `"oidc-discovery.example.org"` |  |
 | logLevel | string | `"info"` |  |
 | nameOverride | string | `""` |  |
-| nodeAttestor.k8s.psat.enabled | bool | `true` |  |
-| nodeAttestor.k8s.psat.rbac.create | bool | `true` |  |
+| nodeAttestor.k8sPsat.enabled | bool | `true` |  |
+| nodeAttestor.k8sPsat.rbac.create | bool | `true` |  |
 | nodeSelector | object | `{}` |  |
 | podAnnotations | object | `{}` |  |
 | podSecurityContext | object | `{}` |  |
--- a/charts/spire/charts/spire-server/README.md
+++ b/charts/spire/charts/spire-server/README.md
@@ -62,6 +62,8 @@ A Helm chart to install the SPIRE server.
 | jwtIssuer | string | `"oidc-discovery.example.org"` |  |
 | logLevel | string | `"info"` |  |
 | nameOverride | string | `""` |  |
+| nodeAttestor.k8s.enabled | bool | `true` |  |
+| nodeAttestor.k8s.rbac.create | bool | `true` |  |
 | nodeSelector | object | `{}` |  |
 | podAnnotations | object | `{}` |  |
 | podSecurityContext | object | `{}` |  |
--- a/charts/spire/charts/spire-server/templates/configmap.yaml
+++ b/charts/spire/charts/spire-server/templates/configmap.yaml
@@ -38,7 +38,7 @@ data:
 
       {{- with .Values.nodeAttestor.k8sPsat }}
       {{- if eq (.enabled | toString) "true" }}
-      NodeAttestor "k8sPsat" {
+      NodeAttestor "k8s_psat" {
         plugin_data {
           clusters = {
             {{ $server.Values.clusterName | quote }} = {
--- a/charts/spire/charts/spire-server/templates/configmap.yaml
+++ b/charts/spire/charts/spire-server/templates/configmap.yaml
@@ -36,9 +36,9 @@ data:
         }
       }
 
-      {{- with .Values.nodeAttestor.k8s.psat }}
+      {{- with .Values.nodeAttestor.k8sPsat }}
       {{- if eq (.enabled | toString) "true" }}
-      NodeAttestor "k8s_psat" {
+      NodeAttestor "k8sPsat" {
         plugin_data {
           clusters = {
             {{ $server.Values.clusterName | quote }} = {
--- a/charts/spire/charts/spire-server/templates/configmap.yaml
+++ b/charts/spire/charts/spire-server/templates/configmap.yaml
@@ -1,3 +1,4 @@
+{{ $release := .Release }}
 apiVersion: v1
 kind: ConfigMap
 metadata:
@@ -35,15 +36,19 @@ data:
         }
       }
 
+      {{- with .Values.nodeAttestor.k8s }}
+      {{- if eq (.enabled | toString) "true" }}
       NodeAttestor "k8s_psat" {
         plugin_data {
           clusters = {
             {{ .Values.clusterName | quote }} = {
-              service_account_allow_list = ["{{ .Release.Namespace }}:{{ .Release.Name }}-agent"]
+              service_account_allow_list = ["{{ $release.Namespace }}:{{ $release.Name }}-agent"]
             }
           }
         }
       }
+      {{- end }}
+      {{- end }}
 
       KeyManager "disk" {
         plugin_data {
--- a/charts/spire/charts/spire-server/templates/configmap.yaml
+++ b/charts/spire/charts/spire-server/templates/configmap.yaml
@@ -1,4 +1,4 @@
-{{ $dot := . }}
+{{ $server := . }}
 apiVersion: v1
 kind: ConfigMap
 metadata:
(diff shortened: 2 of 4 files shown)


kfox1111:
It can be merged at any time but not super useful without custom plugin support.

marcofranssen (commented):
I think we should not allow to disable this. I think this a good opiniated choice we can make on behalf of the users.

kfox1111:
I actually need to deploy multiple different spire servers hosted by Kubernetes that are not bound to the Kubernetes cluster for multiple use cases. I think this is going to be a more common need case then you think.

In some cases, the tpm NodeAttestor will replace the k8s one, and will have stronger security because of it.

charts/spire/charts/spire-server/templates/configmap.yaml
```diff
@@ -1,3 +1,4 @@
+{{ $release := .Release }}
```
faisal-memon, line 1: Why convert this to a variable.

kfox1111: Down below the chunk of code using the value is wrapped with a 'with'. That redefines '.' to be what is being 'with'ed.

In order to get access to the top level release object within the 'with' block, it needs to be made an explicit variable.

charts/spire/charts/spire-server/templates/roles.yaml
```diff
@@ -32,13 +32,15 @@ roleRef:
...
+
+{{- if and .Values.nodeAttestor.k8s.enabled .Values.nodeAttestor.k8s.rbac.create }}
 ---
 # ClusterRole to allow spire-server node attestor to query Token Review API
 # and to be able to push certificate bundles to a configmap
 kind: ClusterRole
 apiVersion: rbac.authorization.k8s.io/v1
 metadata:
-  name: {{ include "spire-server.fullname" . }}
+  name: {{ .Release.Namespace }}-{{ include "spire-server.fullname" . }}
```
faisal-memon, line 43: Is the name change critical to this overall PR? If not can we break that piece out?

charts/spire/charts/spire-server/templates/configmap.yaml
faisal-memon: ah yes, scoping. can we move this down to be on top of the block using it?

kfox1111: We could. General convention is to put it at the top so that if anything else needs it, it doesn't need to move again.

That ok to do or still think we should move it?

charts/spire/charts/spire-server/README.md
```diff
@@ -62,6 +62,8 @@ A Helm chart to install the SPIRE server.
 | jwtIssuer | string | `"oidc-discovery.example.org"` |  |
 | logLevel | string | `"info"` |  |
 | nameOverride | string | `""` |  |
+| nodeAttestor.k8s.enabled | bool | `true` |  |
+| nodeAttestor.k8s.rbac.create | bool | `true` |  |
```
faisal-memon, line 66: Can you add psat in here? cc @marcofranssen

kfox1111: Do we plan on supporting non psat with the template?

charts/spire/charts/spire-server/templates/roles.yaml
kfox1111: done

charts/spire/charts/spire-server/templates/configmap.yaml
kfox1111: Actually, It resolved itself on merge conflict from main. Namespace was added to the top and that's all we need. so I rebased on that.

kfox1111: Nah. that didn't work. Had to go the other way. Lets see if this clears the tests.

charts/spire/charts/spire-server/README.md
faisal-memon: Probably not sat attestor but i think its good to be unambiguous that this relates to the psat attestor.

charts/spire/charts/spire-server/templates/configmap.yaml
```diff
@@ -1,4 +1,4 @@
-{{ $namespace := .Release.Namespace }}
+{{ $dot := . }}
```
faisal-memon, line 1: Nice trick. I think a name like `$server` would be better.

kfox1111: ok.

charts/spire/charts/spire-server/README.md
kfox1111: kk

kfox1111: done

charts/spire/charts/spire-server/templates/configmap.yaml
kfox1111: done

faisal-memon:
looks good to me. @marcofranssen are you ok with the new variables?

charts/spire/charts/spire-server/values.yaml
```diff
@@ -91,6 +91,13 @@ ca_subject:
   organization: Example
   common_name: example.org
 
+nodeAttestor:
+  k8s:
+    psat:
+      enabled: true
+      rbac:
+        create: true
```
marcofranssen, line 99: I would not nest the psat part but do it like k8sPsat because that makes it easier to loop all keys once we add more nodeAttestors.

Having multiple levels of nesting complicates this.

charts/spire/charts/spire-server/templates/configmap.yaml
```diff
@@ -36,15 +36,19 @@ data:
         }
       }
 
+      {{- with .Values.nodeAttestor.k8s.psat }}
```
marcofranssen, line 39: Eventually we can loop all nodeAttestorKeys here. As mentioned below. That's why we should have only one level of nesting in values.

Makes this part simpler for us to manage, and still gives the users a nice consistent API.

marcofranssen (changes requested):
Small nitpick to allow easier templating of the nodeAttestors still to be added.

The rest looks good.

charts/spire/charts/spire-server/templates/configmap.yaml
kfox1111: kk

charts/spire/charts/spire-server/values.yaml
kfox1111: kk

charts/spire/charts/spire-server/values.yaml
```diff
@@ -91,6 +91,12 @@ ca_subject:
   organization: Example
   common_name: example.org
 
+nodeAttestor:
+  k8sPsat:
+    enabled: true
+    rbac:
+      create: true
```
marcofranssen, line 98:
```suggestion
      enabled: true
```

charts/spire/charts/spire-server/templates/roles.yaml
```diff
@@ -72,7 +74,6 @@ kind: ClusterRoleBinding
 apiVersion: rbac.authorization.k8s.io/v1
 metadata:
   name: {{ include "spire-server.fullname" . }}
-  namespace: {{ .Release.Namespace }}
```
marcofranssen, line 75: Why are we removing this?

charts/spire/charts/spire-server/README.md
```diff
@@ -62,6 +62,8 @@ A Helm chart to install the SPIRE server.
 | jwtIssuer | string | `"oidc-discovery.example.org"` |  |
 | logLevel | string | `"info"` |  |
 | nameOverride | string | `""` |  |
+| nodeAttestor.k8sPsat.enabled | bool | `true` |  |
+| nodeAttestor.k8sPsat.rbac.create | bool | `true` |  |
```
marcofranssen, line 66:
```suggestion
```

charts/spire/charts/spire-server/templates/roles.yaml
```diff
@@ -44,6 +44,8 @@ roleRef:
   kind: Role
   name: {{ include "spire-server.fullname" . }}
   apiGroup: rbac.authorization.k8s.io
+
+{{- if and .Values.nodeAttestor.k8sPsat.enabled .Values.nodeAttestor.k8sPsat.rbac.create }}
```
marcofranssen, line 48:
```suggestion
{{- if .Values.nodeAttestor.k8sPsat.enabled }}
```

charts/spire/charts/spire-server/values.yaml
marcofranssen, line 98:
```suggestion
```

marcofranssen (changes requested):
Couple of suggestions to minimize the options, I think we should just always create the RBAC.

charts/spire/charts/spire-server/templates/configmap.yaml
marcofranssen: A common convention is to call this `$root`

charts/spire/charts/spire-server/templates/configmap.yaml
```diff
@@ -1,4 +1,4 @@
-{{ $namespace := .Release.Namespace }}
+{{ $server := . }}
```
marcofranssen, line 1:
```suggestion
{{ $root := . }}
```

Is a common convention, as you refer to the root of the values. $server is misleading, as `.` refers to root of values.

charts/spire/charts/spire-server/templates/configmap.yaml
```diff
@@ -36,15 +36,19 @@ data:
...
 
+      {{- with .Values.nodeAttestor.k8sPsat }}
+      {{- if eq (.enabled | toString) "true" }}
       NodeAttestor "k8s_psat" {
         plugin_data {
           clusters = {
-            {{ .Values.clusterName | quote }} = {
-              service_account_allow_list = ["{{ .Release.Namespace }}:{{ .Release.Name }}-agent"]
+            {{ $server.Values.clusterName | quote }} = {
+              service_account_allow_list = ["{{ $server.Release.Namespace }}:{{ $server.Release.Name }}-agent"]
```
marcofranssen, line 45:
```suggestion
            {{ $root.Values.clusterName | quote }} = {
              service_account_allow_list = ["{{ $root.Release.Namespace }}:{{ $root.Release.Name }}-agent"]
```

charts/spire/charts/spire-server/templates/configmap.yaml
```diff
@@ -80,7 +84,7 @@ data:
           issuer_name = {{ .issuer_name | quote }}
           issuer_kind = {{ .issuer_kind | quote }}
           issuer_group = {{ .issuer_group | quote }}
-          namespace = {{ default $namespace .namespace | quote }}
+          namespace = {{ default $server.Release.Namespace .namespace | quote }}
```
marcofranssen, line 87:
```suggestion
          namespace = {{ default $root.Release.Namespace .namespace | quote }}
```

charts/spire/charts/spire-server/templates/roles.yaml
kfox1111: Just removing namespace. its dead code. helm does that.

dennisgove:
> I actually need to deploy multiple different spire servers hosted by Kubernetes that are not bound to the Kubernetes cluster for multiple use cases.

I have this use-case as well. I don't think it's appropriate for us to force on all users that node attestation must occur with Kubernetes just because the SPIRE Servers are running within Kubernetes.

kfox1111 (commented):
Merged suggestions

marcofranssen (approved):
LGTM 🚀

Issue #17, opened 2023-02-16
Support namespaceOverride configuration


faisal-memon:
With the HPE helm charts we configured namespace similar to the name: https://github.com/alexandrealvino/spire-helm-charts/blob/main/helm/spire/spire-server/templates/statefulset.yaml#L5

And then had a section in the helpers.tpl to allow release namespace to overridden by a `namespaceOverride` variable specified in the values.yaml: https://github.com/alexandrealvino/spire-helm-charts/blob/main/helm/spire/spire-server/values.yaml#L8-L9

The helpers.tpl has a small snippet that brings it together: https://github.com/alexandrealvino/spire-helm-charts/blob/main/helm/spire/spire-server/templates/_helpers.tpl#L26-L35

PR #84, opened 2023-03-03
Test that it is possible to lock down security of pods

kfox1111:
This patch tests that it is possible to run the chart using the Kubernetes Pod Security Standards to secure the
namespaces along with testing restricted mode on pods that support it.

### DIFF
--- a/.github/tests/lockdown/deps-values.yaml
+++ b/.github/tests/lockdown/deps-values.yaml
@@ -1,0 +1,54 @@
+global:
+  telemetry:
+    prometheus:
+      enabled: true
+
+spiffe-csi-driver:
+  enabled: false
+
+spire-agent:
+  enabled: false
+
+spiffe-oidc-discovery-provider:
+  enabled: true
+  insecureScheme:
+    enabled: true
+  podSecurityContext:
+    runAsUser: 1000
+    runAsGroup: 1000
+  securityContext:
+    allowPrivilegeEscalation: false
+    runAsNonRoot: true
+    readOnlyRootFilesystem: true
+    capabilities:
+      drop: [ALL]
+    seccompProfile:
+      type: RuntimeDefault
+
+spire-server:
+  nodeAttestor:
+    k8sPsat:
+      serviceAccountAllowList: ["lockdown:spire-agent"]
+  notifier:
+    k8sbundle:
+      namespace: lockdown
+  podSecurityContext:
+    runAsUser: 1000
+    runAsGroup: 1000
+  securityContext:
+    allowPrivilegeEscalation: false
+    runAsNonRoot: true
+    readOnlyRootFilesystem: true
+    capabilities:
+      drop: [ALL]
+    seccompProfile:
+      type: RuntimeDefault
+  controllerManager:
+    securityContext:
+      allowPrivilegeEscalation: false
+      runAsNonRoot: true
+      readOnlyRootFilesystem: true
+      capabilities:
+        drop: [ALL]
+      seccompProfile:
+        type: RuntimeDefault
--- a/.github/tests/lockdown/pre-install.sh
+++ b/.github/tests/lockdown/pre-install.sh
@@ -2,3 +2,4 @@
 kubectl label namespace "$scenario" pod-security.kubernetes.io/enforce=privileged
 kubectl create namespace "${scenario}-deps"
 kubectl label namespace "${scenario}-deps" pod-security.kubernetes.io/enforce=restricted
+helm install -n "${scenario}-deps" spire -f "${TEST_DIR}"/deps-values.yaml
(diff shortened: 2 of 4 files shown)


charts/spire/charts/spiffe-oidc-discovery-provider/values.yaml
```diff
@@ -49,7 +49,7 @@ insecureScheme:
   nginx:
     image:
       registry: docker.io
-      repository: nginx
+      repository: nginxinc/nginx-unprivileged
```
marcofranssen, line 52: Prefer to start using the chainguard images once, this one is resolved.

https://github.com/chainguard-images/images/issues/93

We use also the wait-for-it image from chainguard.

charts/spire/charts/spiffe-oidc-discovery-provider/templates/deployment.yaml
```diff
@@ -96,7 +99,7 @@ spec:
           image: {{ template "spiffe-oidc-discovery-provider.image" .Values.telemetry.prometheus.nginxExporter }}
           imagePullPolicy: {{ .Values.telemetry.prometheus.nginxExporter.image.pullPolicy }}
           args:
-          - -nginx.scrape-uri=http://127.0.0.1/stub_status
+          - -nginx.scrape-uri=http://127.0.0.1:8080/stub_status
```
marcofranssen, line 102: Can we do this in a prometheus/metrics PR?

charts/spire/charts/spiffe-oidc-discovery-provider/values.yaml
kfox1111: Sounds good. Maybe file an issue for it so we don't forget about it?

charts/spire/charts/spire-agent/templates/configmap.yaml
```diff
@@ -8,7 +8,7 @@ data:
     agent {
       data_dir = "/run/spire"
       log_level = {{ .Values.logLevel | quote }}
-      server_address = "{{ .Release.Name }}-server"
+      server_address = {{ .Values.server.address | default (printf "%s-server" .Release.Name) | quote }}
```
marcofranssen, line 11: this is harder to read then the original.

charts/spire/charts/spiffe-oidc-discovery-provider/templates/deployment.yaml
kfox1111: Not really prom related. The pr here is switching to unprivileged nginx, so it cant run at port 80 so it now runs at port 8080. The exporter then breaks as its hardcoded to assume port 80 and needs updating too. I'd prefer not to break something in one pr and fix it in another.

charts/spire/charts/spire-agent/templates/configmap.yaml
kfox1111: True, but does a lot more then the original. How do we make it more readable?

kfox1111: Think it makes sense to move to helpers?

faisal-memon:
I think its important to lock down the security of the pods. Can we revisit this once we have https://github.com/spiffe/helm-charts/issues/17 in place since thats also related to namespaces?

charts/spire/charts/spire-agent/templates/configmap.yaml
dennisgove: This seems fairly clear to me. Use `server.address` if it exists, otherwise use the default generated value.

kfox1111: Moved to helpers and merged in a separate pr.

kfox1111:
> I think its important to lock down the security of the pods. Can we revisit this once we have #17 in place since that's also related to namespaces?

I think that we should do it the other way around. I prefer to have tests in first before adding things that could break (or the same time). I can see how issue #17 could be used along with the tests to simplify things. I can also see how this pr can be the base to test that #17 works.

So, I think we should merge this first, then tweak the test while #17 is being implemented to ensure it works.

kfox1111:
Once the other 3 prs merge, then this becomes just the stuff under .github.

kfox1111:
Rebased and flattened.

kfox1111:
This one is now good to go.

.github/tests/lockdown/deps-values.yaml
```diff
@@ -0,0 +1,54 @@
+global:
```
marcofranssen, line 1: Where is this values file actually being used for a helm install? I don't get how this works. Looks like some changes are missing.

marcofranssen (changes requested):
Looks like we need to enhance the if statement in the workflow as well to run the test matrix also when files in test folder change. Currently it only runs when the chart changes.

something like ` if: chart.changed || .github/test.changed`

No way to see these changes actually work now.

.github/tests/lockdown/deps-values.yaml
kfox1111: Yeah. that line seems to have gotten dropped out during one of the merge conflicts. I've added it back.

marcofranssen (approved):
LGTM :rocket:

PR #88, opened 2023-03-04
Add post-install scripts to all tests to capture failure details in GitHub workflow summary

marcofranssen:
See https://github.com/spiffe/helm-charts/actions/runs/4808535068

for the result at the summary page.

The checks fail because I didn't update chart docs as intended, that commit bumping the chart version has been removed from this branch once, I saw it nicely executed the workflow.

Please note to merge #242 first. This PR will squash on top of that and therefore doesn't need another rebase, merge.

### DIFF
--- a/.github/tests/common.sh
+++ b/.github/tests/common.sh
@@ -1,0 +1,69 @@
+#!/usr/bin/env bash
+
+get_namespace_details () {
+cat <<EOF >>"$GITHUB_STEP_SUMMARY"
+### Namespace $1
+
+#### Events
+
+\`\`\`shell
+$(kubectl --request-timeout=30s get events --output wide --namespace "$1")
+\`\`\`
+
+#### Pods
+
+\`\`\`shell
+$(kubectl --request-timeout=30s describe pods --namespace "$1")
+\`\`\`
+
+#### Logs
+
+\`\`\`shell
+$(kubectl get pods -o name -n "$1" | while read -r line; do echo logs for "${line}"; kubectl logs -n "$1" "${line}" --all-containers=true --ignore-errors=true; done)
+\`\`\`
+
+EOF
+}
+
+k_wait () {
+  kubectl wait --for condition=available --timeout 30s --namespace "$1" "$2" "$3" | tail -n 1
+}
+
+k_rollout_status () {
+  kubectl rollout status --watch --timeout 30s --namespace "$1" "$2" "$3" | tail -n 1
+}
+
+get_spire_release_name () {
+  helm ls -A | grep '^spire' | awk '{print $1}'
+}
+
+print_spire_workload_status () {
+  local ns1
+  local ns2
+
+  ns1="$1"
+  ns2="${2:-$1}"
+
+  release_name="$(get_spire_release_name)"
+
+  cat <<EOF >>"$GITHUB_STEP_SUMMARY"
+### Spire
+
+| Namespace | Workload                                       | Status |
+| --------- | ---------------------------------------------- | ------ |
+| ${ns1}    | ${release_name}-server                         | <pre>$(k_rollout_status "${ns1}" statefulset "${release_name}-server")</pre> |
+| ${ns2}    | ${release_name}-spiffe-csi-driver              | <pre>$(k_rollout_status "${ns2}" daemonset "${release_name}-spiffe-csi-driver")</pre> |
+| ${ns2}    | ${release_name}-agent                          | <pre>$(k_rollout_status "${ns2}" daemonset "${release_name}-agent")</pre> |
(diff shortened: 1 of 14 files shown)


.github/tests/upstream-authority-cert-manager/post-install.sh
```diff
@@ -0,0 +1,28 @@
...
+EOF
+
+cat <<EOF >>"$GITHUB_STEP_SUMMARY"
+### spire
+
+| workload | Status |
+| -------- | ------ |
+| spire-server | $("${k_rollout_status[@]}" "${NS}" statefulset spire-server) |
+| spire-spiffe-csi-driver | $("${k_rollout_status[@]}"  "${NS}" daemonset spire-spiffe-csi-driver) |
+| spire-agent | $("${k_rollout_status[@]}" "${NS}" daemonset spire-agent) |
```
faisal-memon, line 27: oidc?

.github/tests/upstream-authority-cert-manager/post-install.sh
```diff
@@ -0,0 +1,28 @@
...
+k_rollout_status=(kubectl rollout status --watch --timeout 30s --namespace)
+
+cat <<EOF >>"$GITHUB_STEP_SUMMARY"
+### cert-manager
+
+| workload | Status |
+| -------- | ------ |
+| cert-manager | $("${k_wait[@]}" cert-manager deployments.apps cert-manager) |
+| cert-manager-cainjector | $("${k_wait[@]}" cert-manager deployments.apps cert-manager-cainjector) |
+| cert-manager-webhook | $("${k_wait[@]}" cert-manager deployments.apps cert-manager-webhook) |
```
faisal-memon, line 17: Not following this, we aren't deploying cert-manager.

.github/tests/upstream-authority-cert-manager/post-install.sh
marcofranssen: This test doesn't have oidc enabled.

faisal-memon:
The post-install.sh scripts are all duplicated. Can we have a common snippet somewhere thats imported?

kfox1111:
This is looking good. Should we move this stuff to common.sh too? https://github.com/spiffe/helm-charts/blob/main/.github/tests/namespace-override/post-install.sh#L8-L14

.github/tests/federation-bundle-endpoint/post-install.sh
```diff
@@ -0,0 +1,17 @@
+#!/usr/bin/env bash
+
+set -x
+
+SCRIPT=$(readlink -f "$0")
+SCRIPTPATH=$(dirname "$SCRIPT")
+NS="${SCRIPTPATH##*/}"
```
kfox1111, line 7: NS is the same thing as $scenario. Should probably just use that, or add NS up at the workflow level.

.github/tests/common.sh
```diff
@@ -0,0 +1,58 @@
...
+  cat <<EOF >>"$GITHUB_STEP_SUMMARY"
+### $1
+
+| workload                             | Status |
+| ------------------------------------ | ------ |
+| spire-server                         | <pre>$(k_rollout_status "$1" statefulset spire-server)</pre> |
+| spire-controller-manager             | <pre>$(k_rollout_status "$1" statefulset spire-controller-manager)</pre> |
+| spire-spiffe-csi-driver              | <pre>$(k_rollout_status "$1" daemonset spire-spiffe-csi-driver)</pre> |
+| spire-agent                          | <pre>$(k_rollout_status "$1" daemonset spire-agent)</pre> |
+| spire-spiffe-oidc-discovery-provider | <pre>$(k_rollout_status "$1" deployments.apps spire-spiffe-oidc-discovery-provider)</pre> |
```
kfox1111, line 46: These all use $1, assuming there is only ever one namespace. Might be more reusable to define $1 and $2 so split namespaces work too?

.github/tests/production-example/post-install.sh
```diff
@@ -5,47 +5,24 @@ set -x
 SCRIPT=$(readlink -f "$0")
 SCRIPTPATH=$(dirname "$SCRIPT")
 
-k_wait=(kubectl wait --for condition=available --timeout 30s --namespace)
-k_rollout_status=(kubectl rollout status --watch --timeout 30s --namespace)
+# shellcheck source=/dev/null
```
kfox1111, line 8: What does this do?

marcofranssen: https://www.shellcheck.net/wiki/SC1091

marcofranssen:
This PR got rebased on top of the shellcheck one #242 so I could eliminate some common mistakes.

kfox1111 (approved):
LGTM

PR #91, opened 2023-03-06
Make spire agent server address configurable

kfox1111:
If your server is not in the same namespace or cluster as the agent, you need a config option to specify where it is.


dennisgove (approved):
lgtm

PR #92, opened 2023-03-06
Make server service account allow list configurable

kfox1111:
If you don't have the agent and server in the same namespace or the same cluster, or want to bind additional clusters, you need the ability to configure the service account allow list.

### DIFF
--- a/charts/spire/charts/spire-server/templates/_helpers.tpl
+++ b/charts/spire/charts/spire-server/templates/_helpers.tpl
@@ -94,8 +94,8 @@ Create the name of the service account to use
 {{- end }}
 
 {{- define "spire-server.serviceAccountAllowedList" }}
-{{- if ne (len .Values.nodeAttestor.k8sPsat.serviceAccountsAllowed) 0 }}
-{{- .Values.nodeAttestor.k8sPsat.serviceAccountsAllowed | toJson }}
+{{- if ne (len .Values.nodeAttestor.k8sPsat.serviceAccountAllowList) 0 }}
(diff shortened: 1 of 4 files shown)


dennisgove (approved):
This looks fine to me. 

A minor nit, but could we keep the value name inline with the plugin setting name? `nodeAttestor.k8sPsat.serviceAccountAllowList` instead of `nodeAttestor.k8sPsat.serviceAccountsAllowed`.

kfox1111:
> This looks fine to me.
> 
> A minor nit, but could we keep the value name inline with the plugin setting name? `nodeAttestor.k8sPsat.serviceAccountAllowList` instead of `nodeAttestor.k8sPsat.serviceAccountsAllowed`.

Oh, good idea. Fixing...

charts/spire/charts/spire-server/templates/_helpers.tpl
```diff
@@ -92,3 +92,11 @@ Create the name of the service account to use
 {{- define "spire-k8s-workload-registrar.fullname" -}}
 {{ include "spire-server.fullname" . | trimSuffix "-server" }}-k8s-workload-registrar
 {{- end }}
+
+{{- define "spire-server.serviceAccountAllowedList" }}
+{{- if ne (len .Values.nodeAttestor.k8sPsat.serviceAccountAllowList) 0 }}
+{{- .Values.nodeAttestor.k8sPsat.serviceAccountAllowList | toJson }}
+{{- else }}
+[{{ printf "%s:%s-agent" .Release.Namespace .Release.Name | quote }}]
```
marcofranssen, line 100: This introduces a bug of assigning a single value as opposed to a list of service accounts with a single element.

PR #93, opened 2023-03-06
Make the namespace the bundle is dropped into configurable

kfox1111:
When the server and agent are not in the same namespace, the bundle needs to be uploadable in the agent's namespace.

### DIFF
--- a/charts/spire/charts/spire-server/templates/roles.yaml
+++ b/charts/spire/charts/spire-server/templates/roles.yaml
@@ -3,13 +3,22 @@
 kind: Role
 apiVersion: rbac.authorization.k8s.io/v1
 metadata:
-  name: {{ include "spire-server.fullname" . }}
+  name: {{ include "spire-server.fullname" . }}-pods
   namespace: {{ .Release.Namespace }}
 rules:
   # allow "get" access to pods (to resolve selectors for PSAT attestation)
   - apiGroups: [""]
     resources: [pods]
     verbs: [get]
+    # allow access to "get" and "patch" the spire-bundle ConfigMap (for SPIRE
+    # agent bootstrapping, see the spire-bundle ConfigMap below)
+---
+kind: Role
+apiVersion: rbac.authorization.k8s.io/v1
+metadata:
+  name: {{ include "spire-server.fullname" . }}-bundle
+  namespace: {{ .Values.notifier.k8sbundle.namespace | default .Release.Namespace }}
+rules:
     # allow access to "get" and "patch" the spire-bundle ConfigMap (for SPIRE
     # agent bootstrapping, see the spire-bundle ConfigMap below)
   - apiGroups: [""]
@@ -34,7 +43,7 @@ rules:
 kind: RoleBinding
 apiVersion: rbac.authorization.k8s.io/v1
 metadata:
-  name: {{ include "spire-server.fullname" . }}
+  name: {{ include "spire-server.fullname" . }}-pods
   namespace: {{ .Release.Namespace }}
 subjects:
   - kind: ServiceAccount
@@ -43,6 +52,20 @@ subjects:
 roleRef:
   kind: Role
   name: {{ include "spire-server.fullname" . }}
+  apiGroup: rbac.authorization.k8s.io
+---
+kind: RoleBinding
+apiVersion: rbac.authorization.k8s.io/v1
+metadata:
+  name: {{ include "spire-server.fullname" . }}-bundle
+  namespace: {{ .Values.notifier.k8sbundle.namespace | default .Release.Namespace }}
+subjects:
+  - kind: ServiceAccount
+    name: {{ include "spire-server.serviceAccountName" . }}
+    namespace: {{ .Release.Namespace }}
+roleRef:
+  kind: Role
+  name: {{ include "spire-server.fullname" . }}-bundle
   apiGroup: rbac.authorization.k8s.io
 ---
 # ClusterRole to allow spire-server node attestor to query Token Review API
--- a/charts/spire/charts/spire-server/values.yaml
+++ b/charts/spire/charts/spire-server/values.yaml
@@ -115,6 +115,10 @@ upstreamAuthority:
     namespace: ""
     kube_config_file: ""
 
+notifier:
+  k8sbundle:
+    namespace: ""
+
 controllerManager:
   enabled: false
 
--- a/charts/spire/charts/spire-server/README.md
+++ b/charts/spire/charts/spire-server/README.md
@@ -63,6 +63,7 @@ A Helm chart to install the SPIRE server.
 | logLevel | string | `"info"` |  |
 | nameOverride | string | `""` |  |
 | nodeSelector | object | `{}` |  |
+| notifier.k8sbundle.namespace | string | `""` |  |
 | podAnnotations | object | `{}` |  |
 | podSecurityContext | object | `{}` |  |
 | replicaCount | int | `1` | SPIRE server currently runs with a sqlite database. Scaling to multiple instances will not work until we use an external database. |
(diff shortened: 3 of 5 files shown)


charts/spire/charts/spire-server/templates/roles.yaml
```diff
@@ -12,6 +12,15 @@ rules:
...
     # agent bootstrapping, see the spire-bundle ConfigMap below)
+---
+kind: Role
+apiVersion: rbac.authorization.k8s.io/v1
+metadata:
+  name: {{ include "spire-server.fullname" . }}-bundle
+  namespace: {{ .Values.notifier.k8sbundle.namespace | default .Release.Namespace }}
+rules:
+    # allow access to "get" and "patch" the spire-bundle ConfigMap (for SPIRE
+    # agent bootstrapping, see the spire-bundle ConfigMap below)
```
kfox1111, line 23: The previous role/rolebinding is split into two because the role/rolebinding only provides permissions into the namespace the role/rolebinding is created in. In this case, the ability to deal with the configmap is in the other namespace so needs separate objects.

charts/spire/charts/spire-server/templates/roles.yaml
```diff
@@ -3,7 +3,7 @@
 kind: Role
 apiVersion: rbac.authorization.k8s.io/v1
 metadata:
-  name: {{ include "spire-server.fullname" . }}
+  name: {{ include "spire-server.fullname" . }}-pods
```
marcofranssen, line 6: Why changing this role name? Arn't roles per definition used for pods?

kfox1111: The previous role/rolebinding is split into two because the role/rolebinding only provides permissions into the namespace the role/rolebinding is created in. In this case, the ability to deal with the configmap is in the other namespace so needs separate objects.

Roles can impart permissions to all object types, not just pods.

kfox1111: Actually, I think there is a cleaner way to do this now that you make me look at it a little closer.....

the pod permissions in the rule are actually dead code. Its only able to look up pods in the local namespace with it. Its the cluster role /binding later in the file that actually lets it get the pods cluster wide.

kfox1111: Except, now that upstream certmanager is a thing, that still needs to be in the spire-server namespace....

Will have to rework this a bit.

charts/spire/charts/spire-server/templates/roles.yaml
```diff
@@ -1,15 +1,13 @@
...
 metadata:
-  name: {{ include "spire-server.fullname" . }}
-  namespace: {{ .Release.Namespace }}
+  name: {{ include "spire-server.fullname" . }}-bundle
+  namespace: {{ .Values.notifier.k8sbundle.namespace | default .Release.Namespace }}
 rules:
-  # allow "get" access to pods (to resolve selectors for PSAT attestation)
-  - apiGroups: [""]
-    resources: [pods]
-    verbs: [get]
```
kfox1111, line 12: The role code here is removed, cause it never worked. It is a Role, so only applies to the spire-server's namespace. Its the cluster role below that lets the pods get PSAT'ed for the entire cluster. That remains and is functional.

charts/spire/charts/spire-server/templates/roles.yaml
kfox1111: K. reworked it to fix the dead pod code and incorporate the merged cert-manager code that merged kind of poorly.

faisal-memon:
Does it have to be in the agents namespace or can it be in the servers namespace? I think its easier if its the latter.

charts/spire/charts/spire-server/templates/roles.yaml
marcofranssen: Nit pick, can we move this into a separate PR. For the sake of traceability and release notes.

Our CONTRIBUTING.md also states that PRs should have a single subject. This change does not align with that PR title.

kfox1111:
> Does it have to be in the agents namespace or can it be in the servers namespace? I think its easier if its the latter.

Only used by the agent. It has to be there.

charts/spire/charts/spire-server/templates/roles.yaml
kfox1111: sigh

nit picking the nit pick a little bit. Its going to take a lot more effort to add another role just to remove it again.We're early in the project and we're slowing things way down with nitpicking the really small stuff.

I can kind of live with this though.

But, one of the things I'm really objecting to is the little feature patches having to merge before the tests that ensure they work. That is BAD imo. I broke this off a working pr with tests ensuring it works. now there are changes being asked for and there is no tests to ensure the changes work. So, I think the current process needs fixing.

kfox1111: https://github.com/spiffe/helm-charts/pull/104

kfox1111: done.

charts/spire/charts/spire-server/templates/roles.yaml
```diff
@@ -1,10 +1,12 @@
-# ClusterRole to allow spire-server node attestor to query Token Review API
-# and to be able to push certificate bundles to a configmap
+# ClusterRole to allow spire-server node attestor to query Token Review API,
+# Role to be able to push certificate bundles to a configmap, and Role to
+# be able to manage cert requests with Cert-Manager
```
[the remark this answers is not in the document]
kfox1111: The comment is accurate. There is a ClusterRole (not Role) that does the pod stuff for Token Review, a Role for bundles, and an role for cert-manager.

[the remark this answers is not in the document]
kfox1111: Its kind of hard to see it in the diff view. but if you go here, you can see the while file:
https://github.com/spiffe/helm-charts/blob/dcc255c6049c9b60097eee8ec4e3dc69296e2c41/charts/spire/charts/spire-server/templates/roles.yaml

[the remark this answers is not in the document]
faisal-memon: Yes the ClusterRole comment is duplicated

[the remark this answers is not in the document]
faisal-memon: Yea thats the one. My guess is it was a copy+paste error. Do you mind deleting the duplicate comment and moving the new one below the seperator so its clear that it applies to that block?

charts/spire/charts/spire-server/templates/roles.yaml
```diff
@@ -25,20 +34,34 @@ rules:
...
+---
+kind: RoleBinding
+apiVersion: rbac.authorization.k8s.io/v1
+metadata:
+  name: {{ include "spire-server.fullname" . }}-bundle
+  namespace: {{ .Values.notifier.k8sbundle.namespace | default .Release.Namespace }}
+subjects:
+  - kind: ServiceAccount
+    name: {{ include "spire-server.serviceAccountName" . }}
+    namespace: {{ .Release.Namespace }}
```
faisal-memon, line 61: Does need to be `{{ .Values.notifier.k8sbundle.namespace | default .Release.Namespace }}`?

faisal-memon: nm, i get it. I didnt know you could apply RoleBinding across namespaces.

kfox1111: yeah. the rolebinding must be in the namespace where the role is and the permissions will be restricted to. but, the service account its bound to can be in any namespace.

faisal-memon:
Change looks good to me.

charts/spire/charts/spire-server/values.yaml
```diff
@@ -115,6 +115,10 @@ upstreamAuthority:
     namespace: ""
     kube_config_file: ""
 
+notifier:
+  k8sbundle:
+    namespace: ""
```
faisal-memon, line 120:
```suggestion
    # -- Namespace to push the bundle into, if blank will default to SPIRE Server namespace
    namespace: ""
```

kfox1111:
Ok. Had to rework this and flatten.

faisal-memon (approved):

PR #94, opened 2023-03-06
Add values for server TTL configurables

### LINKED ISSUE #22: Support configurable CA and workload TTLs

marcofranssen:
Resolves #22

### DIFF
--- a/charts/spire/charts/spire-server/README.md
+++ b/charts/spire/charts/spire-server/README.md
@@ -23,6 +23,7 @@ A Helm chart to install the SPIRE server.
 | autoscaling.minReplicas | int | `1` |  |
 | autoscaling.targetCPUUtilizationPercentage | int | `80` |  |
 | bundleConfigMap | string | `"spire-server"` |  |
+| caTTL | string | `"24h"` |  |
 | ca_subject.common_name | string | `"example.org"` |  |
 | ca_subject.country | string | `"NL"` |  |
 | ca_subject.organization | string | `"Example"` |  |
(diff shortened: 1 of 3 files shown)


charts/spire/charts/spire-server/README.md
```diff
@@ -60,6 +61,7 @@ A Helm chart to install the SPIRE server.
 | imagePullSecrets | list | `[]` |  |
 | initContainers | list | `[]` |  |
 | jwtIssuer | string | `"oidc-discovery.example.org"` |  |
+| jwtSvidTTL | string | `"1h"` |  |
```
faisal-memon, line 64: Should we have the default in front like the spire server does? They would be next to each other in the list if so.

marcofranssen: Thought about that, but then started thinking there isn't some setting as a non-default. Both are fine for me. Just found the default thing a bit weird from Helm perspective, but I might be missing some reasoning behind that default.

kfox1111:
One of the things I was playing with, in the yaml config pr, was to move these sorts of settings under a config: section in values. It enables the whole config section to be imported directly into the server {} block without having to template out every single option. Can we put the settings in values under a config section so we don't have to decide to exclude that kind of implementation?

charts/spire/charts/spire-server/README.md
kfox1111: +1 for keeping the option name the same I think?

default is because the TTL can be overridden at the per entry level in the spire db. So its the default TTL when no override is set.

faisal-memon (approved):

PR #95, opened 2023-03-06
Allow to configure spire-server CA key type

### LINKED ISSUE #18: Support configurable CA key type

marcofranssen:
Resolves #18


kfox1111:
One of the things I was playing with, in the yaml config pr, was to move these sorts of settings under a config: section in values. It enables the whole config section to be imported directly into the server {} block without having to template out every single option. Can we put the settings in values under a config section so we don't have to decide to exclude that kind of implementation?

faisal-memon (approved):

PR #97, opened 2023-03-07
Add option to enable federation on spire-server

petercable:
Adds an option to enable the federation bundle endpoint in `server.conf` along with matching entries in the `Service` and `StatefulSet`.

### DIFF
--- a/charts/spire/charts/spire-server/templates/configmap.yaml
+++ b/charts/spire/charts/spire-server/templates/configmap.yaml
@@ -29,6 +29,17 @@ data:
         common_name = {{ .common_name | quote }},
         {{- end }}
       }
+
+      {{- with .Values.federation }}
+      {{- if eq (.enabled | toString) "true" }}
+      federation {
+        bundle_endpoint {
+          address = "{{ .bundleEndpoint.address }}"
+          port = {{ .bundleEndpoint.port }}
+        }
+      }
+      {{- end }}
+      {{- end }}
     }
 
     plugins {
--- a/charts/spire/charts/spire-server/templates/service.yaml
+++ b/charts/spire/charts/spire-server/templates/service.yaml
@@ -16,5 +16,13 @@ spec:
       port: {{ .Values.service.port }}
       targetPort: grpc
       protocol: TCP
+    {{- with .Values.federation }}
+    {{- if eq (.enabled | toString) "true" }}
+    - name: federation
+      port: {{ .bundleEndpoint.port }}
+      targetPort: federation
+      protocol: TCP
+    {{- end }}
+    {{- end }}
   selector:
     {{- include "spire-server.selectorLabels" . | nindent 4 }}
--- a/charts/spire/charts/spire-server/templates/tests/test-connection.yaml
+++ b/charts/spire/charts/spire-server/templates/tests/test-connection.yaml
@@ -17,4 +17,12 @@ spec:
       args: ['-zvw3', '{{ include "spire-server.fullname" . }}', '{{ .Values.service.port }}']
       securityContext:
         {{- toYaml .Values.securityContext | nindent 8 }}
+    {{- if eq (.Values.federation.enabled | toString) "true" }}
+    - name: wget-federation-bundle-endpoint
+      image: busybox
+      command: ['wget']
+      args: ['--no-check-certificate', '-O', '/dev/null', 'https://{{ include "spire-server.fullname" . }}.{{ include "spire-server.namespace" . }}.svc.cluster.local:{{ .Values.federation.bundleEndpoint.port }}']
+      securityContext:
+        {{- toYaml .Values.securityContext | nindent 8 }}
+    {{- end }}
   restartPolicy: Never
--- a/charts/spire/charts/spire-server/values.yaml
+++ b/charts/spire/charts/spire-server/values.yaml
@@ -87,6 +87,12 @@ trustDomain: example.org
 
 bundleConfigMap: spire-server
 
+federation:
+  enabled: false
+  bundleEndpoint:
+    port: 8443
+    address: "0.0.0.0"
+
 ca_subject:
   country: NL
   organization: Example
--- a/.github/tests/federation-bundle-endpoint/values.yaml
+++ b/.github/tests/federation-bundle-endpoint/values.yaml
@@ -1,0 +1,3 @@
+spire-server:
+  federation:
+    enabled: true
--- a/charts/spire/charts/spire-server/README.md
+++ b/charts/spire/charts/spire-server/README.md
@@ -56,6 +56,9 @@ A Helm chart to install the SPIRE server.
 | extraContainers | list | `[]` |  |
 | extraVolumeMounts | list | `[]` |  |
 | extraVolumes | list | `[]` |  |
+| federation.bundleEndpoint.address | string | `"0.0.0.0"` |  |
+| federation.bundleEndpoint.port | int | `8443` |  |
+| federation.enabled | bool | `false` |  |
 | fullnameOverride | string | `""` |  |
 | image.pullPolicy | string | `"IfNotPresent"` |  |
 | image.registry | string | `"ghcr.io"` |  |
--- a/charts/spire/charts/spire-server/templates/statefulset.yaml
+++ b/charts/spire/charts/spire-server/templates/statefulset.yaml
@@ -57,6 +57,13 @@ spec:
               protocol: TCP
             - containerPort: 8080
               name: healthz
+            {{- with .Values.federation }}
+            {{- if eq (.enabled | toString) "true" }}
+            - name: federation
+              containerPort: {{ .bundleEndpoint.port }}
+              protocol: TCP
+            {{- end }}
+            {{- end }}
             {{- if (dig "telemetry" "prometheus" "enabled" .Values.telemetry.prometheus.enabled .Values.global) }}
             - containerPort: 9988
               name: prom


kfox1111:
Thanks for the PR.

I'm not sure if it needs more then one federation section for multiple servers in the federation? Just glancing at https://spiffe.io/docs/latest/architecture/federation/readme/ it looks like there is a bunch of other config that may be needed to make it work as well.

We also should probably consider how this works along with the ingress pr since it needs exposing outside the cluster.

petercable:
> I'm not sure if it needs more then one federation section for multiple servers in the federation?

This just starts the federation endpoint for this cluster. To enable federation with other clusters we can create a `ClusterFederatedTrustDomain` custom resource (https://github.com/spiffe/spire-controller-manager/blob/main/docs/clusterfederatedtrustdomain-crd.md), then the spire-controller-manager pod will tell spire-server to fetch the federation trust bundles from the remote servers.

> We also should probably consider how this works along with the ingress pr since it needs exposing outside the cluster.

I think it might make sense to include this as an option in the ingress PR but I think the dependency should go the other way. Not everyone using federation is going to want to use an `Ingress` to plumb this to the outside world (for instance, I've got this working with our existing istio gateway).

Currently I am doing all of this with kustomize postRenderers but doing so requires me to rewrite the entire `server.conf` config map since I can't just patch a few lines into it:

```
  postRenderers:
    - kustomize:
        patchesJson6902:
          - target:
              kind: StatefulSet
              name: spire-server
            patch:
              - op: add
                path: /spec/template/spec/hostNetwork
                value: true
              - op: add
                path: /spec/template/spec/dnsPolicy
                value: ClusterFirst
              - op: add
                path: /spec/template/spec/containers/0/ports/-
                value:
                  name: http
                  hostPort: 10443
                  containerPort: 10443
                  protocol: TCP
          - target:
              kind: Service
              name: spire-server
            patch:
              - op: add
                path: /spec/ports/-
                value:
                  name: http
                  protocol: TCP
                  port: 10443
                  targetPort: 10443
          - target:
              kind: ConfigMap
              name: spire-server
            patch:
              - op: replace
                path: /data/server.conf
                value: |
                  server {
                    bind_address = "0.0.0.0"
                    bind_port = "8081"
                    socket_path = "/run/spire/server-sockets/spire-server.sock"
                    trust_domain = "example.org"
                    data_dir = "/run/spire/data"
                    log_level = "info"
                    # AWS requires the use of RSA.  EC cryptography is not supported
                    ca_key_type = "rsa-2048"

                    jwt_issuer = "oidc-discovery.example.org"

                    default_x509_svid_ttl = "1h"
                    default_jwt_svid_ttl = "1h"
                    ca_subject = {
                      country = ["US"],
                      organization = ["Example"],
                      common_name = "example.org",
                    }

                    federation {
                      bundle_endpoint {
                          address = "0.0.0.0"
                          port = 10443
                      }
                    }
                  }

                  plugins {
                    DataStore "sql" {
                      plugin_data {
                        database_type = "sqlite3"
                        connection_string = "/run/spire/data/datastore.sqlite3"
                      }
                    }

                    NodeAttestor "k8s_psat" {
                      plugin_data {
                        clusters = {
                          "dc-01" = {
                            service_account_allow_list = ["spire-system:spire-agent"]
                          }
                        }
                      }
                    }

                    KeyManager "disk" {
                      plugin_data {
                        keys_path = "/run/spire/data/keys.json"
                      }
                    }

                    Notifier "k8sbundle" {
                      plugin_data {
                        namespace = "spire-system"
                        config_map = "spire-bundle"
                      }
                    }
                  }

                  health_checks {
                    listener_enabled = true
                    bind_address = "0.0.0.0"
                    bind_port = "8080"
                    live_path = "/live"
                    ready_path = "/ready"
                  }
```

Unfortunately this renders a big chunk of the helm chart useless, since I have to manually specify some things that should come from `values.yaml` in this patch.

kfox1111:
Got it. Thanks for the explanation. Configuring it through the CR's makes sense...

It looks like its https, how do you deal with the initial untrusted certificates without acme or some other config?

For the port type, can you mark them as 'federation' rather then http? That would make it a bit more clear what its for.

We've been discussing having the config file in json. That would let you patch it easier..... though we'd rather you not have to do that.

petercable:
> It looks like its https, how do you deal with the initial untrusted certificates without acme or some other config?

In our case we put this behind regular PKI on one of our existing load balancers.

> For the port type, can you mark them as 'federation' rather then http? That would make it a bit more clear what its for.

Yep.

kfox1111:
> > It looks like its https, how do you deal with the initial untrusted certificates without acme or some other config?
> 
> In our case we put this behind regular PKI on one of our existing load balancers.

ah. Just ignoring the pki between the load balancer and spire server. Gotcha.

kfox1111:
I think this is probably about ready to merge.

Two more things I think.
1. DCO is broken. That would need to be fixed. See the Details link on that test for details how to fix.
2. I'm thinking the setting should probably be federation.bundleEndpoint instead of federationBundleEndpoint?

petercable:
> I think this is probably about ready to merge.
> 
> Two more things I think.
> 
> 1. DCO is broken. That would need to be fixed. See the Details link on that test for details how to fix.
> 2. I'm thinking the setting should probably be federation.bundleEndpoint instead of federationBundleEndpoint?

Done and done.

charts/spire/charts/spire-server/templates/configmap.yaml
```diff
@@ -28,6 +28,16 @@ data:
         common_name = {{ .common_name | quote }},
         {{- end }}
       }
+      {{- with .Values.federation.bundleEndpoint }}
```
faisal-memon, line 31:
```suggestion

      {{- with .Values.federation.bundleEndpoint }}
```

add a new line to space it from the previous block

faisal-memon:
Docs need to be updated, can you run `./helm-docs.sh` in the root of the repo?

charts/spire/charts/spire-server/values.yaml
```diff
@@ -86,6 +86,12 @@ trustDomain: example.org
 
 bundleConfigMap: spire-server
 
+federation:
+  bundleEndpoint:
+    enabled: true
```
marcofranssen, line 91:
```suggestion
    enabled: false
```

I think false is a better default.

Please also add a test scenario in `.github/tests/federation` where federation is enabled.

As well a test in `charts/spire/charts/spire-server/templates/tests` to test the federation endpoint.

charts/spire/charts/spire-server/templates/service.yaml
```diff
@@ -16,5 +16,13 @@ spec:
       port: {{ .Values.service.port }}
       targetPort: grpc
       protocol: TCP
+    {{- with .Values.federation.bundleEndpoint }}
+    {{- if eq (.enabled | toString) "true" }}
+    - name: federation
+      port: {{ .port }}
+      targetPort: http
```
marcofranssen, line 23:
```suggestion
      targetPort: federation
```

marcofranssen (changes requested):
Small bug I could notice, please also add some tests as suggested.

petercable:
I haven't had time to look at the tests yet today so I went ahead and pushed the other pending changes, I will take a look at adding tests as soon as my schedule allows.

marcofranssen:
> I haven't had time to look at the tests yet today so I went ahead and pushed the other pending changes, I will take a look at adding tests as soon as my schedule allows.

Please consider rebasing this branch see the `Update branch` arrow down buttong on this PR.

We have made couple of changes on tests.

petercable:
well, that is not what I intended to do 😆

petercable:
Looking at `charts/spire/charts/spire-server/templates/tests/test-connection.yaml` it seems like it would be straightforward to add a similar test for the bundle endpoint but it's not clear to me how I ensure that it only runs against an installation with the bundle endpoint installed. Any tips?

kfox1111:
Something like:
https://github.com/spiffe/helm-charts/blob/main/charts/spire/charts/spiffe-oidc-discovery-provider/templates/tests/test-connection.yaml#L25-L30
But with a condition around it so it only runs if .Values.federation.bundleEndpoint.enabled

kfox1111:
an if like that looks like: https://github.com/spiffe/helm-charts/blob/main/charts/spire/charts/spiffe-oidc-discovery-provider/templates/deployment.yaml#L73

kfox1111:
lgtm. lets see how the tests run.

charts/spire/charts/spire-server/templates/tests/test-connection.yaml
```diff
@@ -16,4 +16,12 @@ spec:
       args: ['-zvw3', '{{ include "spire-server.fullname" . }}', '{{ .Values.service.port }}']
       securityContext:
         {{- toYaml .Values.securityContext | nindent 8 }}
+    {{- if eq (Values.federation.bundleEndpoint.enabled | toString) "true" }}
+    - name: wget-federation-bundle-endpoint
+      image: busybox
+      command: ['wget']
+      args: ['--no-check-certificate', '-O', '/dev/null', 'https://{{ include "spire-server.fullname" . }}.{{ .Release.Namespace }}.svc.cluster.local:{{ .Values.federation.bundleEndpoint.port }}']
```
petercable, line 23: it turns out `--no-check-certificate` isn't required because busybox wget doesn't implement certificate verification. Should I leave it here anyway in case this gets implemented or the image is switched out?

kfox1111: Yeah, I'd leave it. Its a nice breadcrumb to know about I think.

marcofranssen:
Test are failing, server fails to boot.

petercable:
> Test are failing, server fails to boot.

Yeah, let me see if I can get this running locally in minikube, figure out what is going on.

petercable:
> got: FLOAT

I see, it wants quotes around `0.0.0.0`. Updated.

charts/spire/charts/spire-server/values.yaml
```diff
@@ -87,6 +87,12 @@ trustDomain: example.org
 
 bundleConfigMap: spire-server
 
+federation:
+  bundleEndpoint:
+    enabled: false
```
marcofranssen, line 92: Should we lift this one level up?

So federation.enabled?

marcofranssen (changes requested):
One small suggestion for a more logical API.

charts/spire/charts/spire-server/values.yaml
kfox1111: Good question.... what else might be enabled if federation is enabled?

charts/spire/charts/spire-server/values.yaml
```diff
@@ -87,6 +87,12 @@ trustDomain: example.org
 
 bundleConfigMap: spire-server
 
+federation:
+  enabled: false
+  bundleEndpoint:
+    port: 8443
+    address: "0.0.0.0"
```
marcofranssen, line 94:
```suggestion
  enabled: false
  bundleEndpoint:
    port: 8443
    address: "0.0.0.0"
  federatesWith:
    trustDomain: broker.example.org
    bundleEndpoint:
      url: ""
  ingress:
    enabled: false
```

marcofranssen (changes requested):
I have made some inline suggestions to add the missing parts 🤞 if all of that is complete enough.

We will also need to add a workload of type `Job`, at least if that works out to bootstrap the federation. (A Job runs as a one time completion thing. Although I'm not sure if we can use that to trigger the spire-server commands, an alternative there would be to run the init parts in a initContainer within spire-server statefulset, but the implementation has to be ephemeral, because when the Pod restarts or gets rebalanced to another node it will also run the initContainer again.

https://spiffe.io/docs/latest/architecture/federation/readme/#get-the-bootstrap-trust-bundle

Last but not least to test the actual federation you will need a test setup using the pre-install.sh like we are doing [here](https://github.com/spiffe/helm-charts/pull/117/files#diff-9585f1add83d43d58c30c0713fbf245cbdf0cc7635398f66186f319399edda25R2-R4) to create another spire-server installation to federate with.

e.g. pre-install.sh does following:

- create the `broker` spire-server

e.g. values.yaml installs the stock-market spire-server like in the example, that federates with this broker spire-server.

charts/spire/charts/spire-server/templates/tests/test-connection.yaml
```diff
@@ -17,4 +17,12 @@ spec:
       args: ['-zvw3', '{{ include "spire-server.fullname" . }}', '{{ .Values.service.port }}']
       securityContext:
         {{- toYaml .Values.securityContext | nindent 8 }}
+    {{- if eq (.Values.federation.enabled | toString) "true" }}
+    - name: wget-federation-bundle-endpoint
+      image: busybox
+      command: ['wget']
+      args: ['--no-check-certificate', '-O', '/dev/null', 'https://{{ include "spire-server.fullname" . }}.{{ .Release.Namespace }}.svc.cluster.local:{{ .Values.federation.bundleEndpoint.port }}']
```
marcofranssen, line 24: A recent merge on upstream added the ability to override namespaces, below suggestion reuses that logic via this Macro to ensure this also works when the namespace is overridden.

```suggestion
      args: ['--no-check-certificate', '-O', '/dev/null', 'https://{{ include "spire-server.fullname" . }}.{{ include "spire-server.namespace" . }}.svc.cluster.local:{{ .Values.federation.bundleEndpoint.port }}']
```

charts/spire/charts/spire-server/values.yaml
petercable: please see this previous discussion:

https://github.com/spiffe/helm-charts/pull/97#issuecomment-1458401788

There are better ways to configure `federatesWith` in kubernetes. The scope of this change is just to enable the bundle endpoint.

charts/spire/charts/spire-server/templates/configmap.yaml
```diff
@@ -29,6 +29,17 @@ data:
...
         {{- end }}
       }
+
+      {{- with .Values.federation }}
+      {{- if eq (.enabled | toString) "true" }}
+      federation {
+        bundle_endpoint {
+          address = "{{ .bundleEndpoint.address }}"
+          port = {{ .bundleEndpoint.port }}
+        }
```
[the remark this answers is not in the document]
petercable: Ingress is just one possible way to expose this federation bundle endpoint. I'm not using it nor do I have a way to reliably test it.

I can try to add an ingress but I can't say when I will have time to do so nor can I guarantee that it works. Could we instead create an issue to add ingress for the bundle endpoint and someone can scratch that particular itch when they need it?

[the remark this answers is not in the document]
kfox1111: @marcofranssen the federates with part is going to be more complicated. as it can also be a list...

He's also using istio so that can be exposed differently then ingress. I think the pr is as small as possible with the minimum implementation for some use cases. But needs further ones to really flesh out all of Federation. Should we merge the bit here and do follow-ons?

charts/spire/charts/spire-server/values.yaml
kfox1111: If we get the general plugin config stuff in place, then federatesWith could be added that way?

charts/spire/charts/spire-server/templates/tests/test-connection.yaml
kfox1111: This is a good suggestion.

marcofranssen:
Could you resolve the DCO @petercable?

```shell
git checkout main
git pull upstream main
git checkout add-federation-bundle-endpoint
git rebase - --signoff
git push --force-with-lease
```

kfox1111:
lgtm. what do you think @marcofranssen?

marcofranssen (approved):
LGTM :rocket:

PR #101, opened 2023-03-08
Remove some duplication on chart-testing CI

marcofranssen:


kfox1111:
Yeah, committing this would break things. .github/tests/upstream-authority-cert-manager/pre-install.sh  needs to be fixed in this pr to follow the changes.

marcofranssen:
Ah, good catch will fix tomorrow.

kfox1111 (approved):

PR #104, opened 2023-03-08
Remove dead role code

kfox1111:
The cluster role does the same thing, but at the cluster level where it belongs. The extra role code does nothing so we remove it here.


faisal-memon:
The cluster role is gated for only if psat is enabled. If psat is disabled will this cause a problem?

kfox1111:
The permission is only used by the psat plugin near as I can tell. Until we have general plugin support, we don't support anything other then the psat plugin either. If it was needed by some third party plugin, its highly likely they would need the cluster variant instead of the local only one.

faisal-memon (approved):

PR #105, opened 2023-03-08
Add tmp mount so that server can run locked down

kfox1111:
This pr adds a tmp mount to the spire server pod so that it can run with read only root.


faisal-memon:
The `spire-server-socket` is under `/tmp` and mounted `emptyDir`. Is there something else with `/tmp` thats shared between server and controller manager?

kfox1111:
controller manager writes temporary files directly under /tmp. it doesn't work when the container is marked readonly.

faisal-memon:
Thanks for the explanation. Do we need both mounts or can we consolidate to just the `/tmp` one?

kfox1111:
None of the files controller-manager needs to write out temporarily need to be shared with the other container. So for security reasons, its best to keep them separate.

PR #106, opened 2023-03-08
Fix the tests so they can run locked down.

kfox1111:
The tests need to be able to run with locked down settings to run in a locked down namespace. This patch adds that capability.


kfox1111:
closing in favor of: https://github.com/spiffe/helm-charts/pull/111

PR #111, opened 2023-03-11
Fix the tests so they can run locked down.

kfox1111:
The tests need to be able to run with locked down settings to run in a locked down namespace. This patch adds that capability.

### DIFF
--- a/charts/spire/charts/spiffe-oidc-discovery-provider/templates/tests/test-connection.yaml
+++ b/charts/spire/charts/spiffe-oidc-discovery-provider/templates/tests/test-connection.yaml
@@ -25,7 +25,7 @@ spec:
     - name: wget-service-name-namespace-svc-cluster-local
       image: busybox
       command: ['wget']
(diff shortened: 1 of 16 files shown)


charts/spire/charts/spiffe-oidc-discovery-provider/templates/tests/test-connection.yaml
```diff
@@ -7,17 +7,25 @@ metadata:
...
     "helm.sh/hook": test
 spec:
+  securityContext:
+    {{- toYaml .Values.podSecurityContext | nindent 4 }}
   containers:
     - name: wget-service-name
       image: busybox
       command: ['wget']
-      args: ['{{ include "spiffe-oidc-discovery-provider.fullname" . }}:{{ .Values.service.port }}/.well-known/openid-configuration']
+      args: ['-O', '/dev/null', '{{ include "spiffe-oidc-discovery-provider.fullname" . }}:{{ .Values.service.port }}/.well-known/openid-configuration']
```
marcofranssen, line 16: Why hide the output? That makes it hard to debug the tests isn't it?

kfox1111: Its not hiding the output, its writing the file it normally would to the current working directory to dev/null instead. That way the root filesystem can be read only.

marcofranssen (approved):

PR #114, opened 2023-03-12
Make the agent socket configurable

kfox1111:
This patch makes the agent socket configurable.

### DIFF
--- a/charts/spire/charts/spiffe-csi-driver/README.md
+++ b/charts/spire/charts/spiffe-csi-driver/README.md
@@ -13,6 +13,7 @@ A Helm chart to install the SPIFFE CSI driver.
 
 | Key | Type | Default | Description |
 |-----|------|---------|-------------|
+| agentSocketPath | string | `"/run/spire/agent-sockets/spire-agent.sock"` |  |
 | fullnameOverride | string | `""` |  |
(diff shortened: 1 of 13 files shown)


marcofranssen (commented):
Subcharts can reuse the helpers from the rootChart. With this change we either have a unused root helper that can be removed or we should see if we can reuse that one to have less duplication.

faisal-memon:
Why does this need to be configurable?

kfox1111:
> Subcharts can reuse the helpers from the rootChart. With this change we either have a unused root helper that can be removed or we should see if we can reuse that one to have less duplication.

I just removed the unused helper. Good catch.

kfox1111:
> Why does this need to be configurable?

Its a hostpath. Not all distros put their stuff in the same locations.

marcofranssen:
Haven't we changed this recently all to empty dir volumes? For spire server we also didn't make this configurable. Within the container the path doesn't really matter right.

kfox1111:
We did it on the server because it is only used within the pod.

The agent is different. It has to export it out to the host, so that the csi-driver can import it, and reexport it to all the pods using the csi driver.

marcofranssen:
Oh right. So this goed hand in hand with the csi PR

marcofranssen (approved):

PR #115, opened 2023-03-12
Make csi driver configurable to be able to run multiple instances

kfox1111:
This patch makes it possible to configure the name of the csi driver.

### DIFF
--- a/charts/spire/charts/spiffe-csi-driver/values.yaml
+++ b/charts/spire/charts/spiffe-csi-driver/values.yaml
@@ -18,6 +18,9 @@ resources: {}
   #   cpu: 100m
   #   memory: 64Mi
 
+healthChecks:
+    port: 9809
+
 imagePullSecrets: []
 nameOverride: ""
 namespaceOverride: ""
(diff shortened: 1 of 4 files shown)


charts/spire/charts/spiffe-csi-driver/values.yaml
```diff
@@ -1,3 +1,6 @@
+# -- Set the csi driver name deployed to Kubernetes.
+pluginName: csi.spiffe.io
```
faisal-memon, line 2: Should this be `{{ .Values.pluginName }}`?

marcofranssen: No in values file this should be the real value. Templates only go inside the templates folder.

charts/spire/charts/spiffe-csi-driver/values.yaml
```diff
@@ -15,6 +18,9 @@ resources: {}
   #   cpu: 100m
   #   memory: 64Mi
 
+healthChecks:
+    port: 9809
```
marcofranssen, line 22:
```suggestion
  port: 9809
```

marcofranssen (changes requested):
Small nitpick regarding yaml indentation

Issue #54, opened 2023-02-26
Nested support


kfox1111:
Some organizations will have a root spire server and then want to bind their K8s clusters to it. The chart should support having an upstream spire server.

PR #117, opened 2023-03-13
Support Nested Spire with External Agent

kfox1111:
This patch enables the nested spire use case:

You can have a root spire server external to Kubernetes that is the base spiffe ca. Then each of your kubernetes clusters can run an instance of this chart bound to the root server and all will be trusted.

fixes https://github.com/spiffe/helm-charts/issues/54, https://github.com/spiffe/helm-charts/issues/422


.github/tests/nested/pre-install.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/bash
+SCRIPT_DIR=$( cd -- "$( dirname -- "${BASH_SOURCE[0]}" )" &> /dev/null && pwd )
+kubectl create namespace "${scenario}-deps"
+helm install -n "${scenario}-deps" spire charts/spire --wait -f $SCRIPT_DIR/deps-values.yaml
+kubectl get all -n "${scenario}-deps"
+kubectl get nodes -o go-template='{{range .items}}{{printf "%s\n" .metadata.uid}}{{end}}' | while read line; do
+	kubectl exec -t spire-server-0 -n "${scenario}-deps" -- spire-server entry create -spiffeID spiffe://example.org/example-cluster/nested-spire -parentID spiffe://example.org/spire/agent/k8s_psat/example-cluster/$line -selector k8s:pod-label:app.kubernetes.io/name:server -downstream
+done
+
+helm install -n "${scenario}" spire-upstream-driver charts/spire --wait -f $SCRIPT_DIR/deps2-values.yaml
```
marcofranssen, line 10: Can we rename deps2 values to better describe what it is used for? 

E.g. upstream-server-values.yaml

Find it quite hard to follow what deps and deps2 represents.

charts/spire/charts/spire-server/templates/configmap.yaml
```diff
@@ -80,6 +80,18 @@ data:
       {{- end }}
       {{- end }}
 
+      {{- with .Values.upstreamAuthority.spire }}
+      {{- if eq (.enabled | toString) "true" }}
+      UpstreamAuthority "spire" {
+        plugin_data {
+          server_address      = {{ .server.address | quote }}
+          server_port         = {{ .server.port }}
+          workload_api_socket = "/run/spire/upstream_agent/spire-agent.sock"
```
marcofranssen, line 89: Should we hardcode this or also make variable? Looks like it should be a variable as it depends how you deploy the agent.

marcofranssen (changes requested):
I find it very hard to follow this PR. Why would you want to nest spire servers on the same cluster? This PR looks like making it possible to nest on the same cluster, but I wonder why you would like to go that route with al complexity involved in this chart. 

Could you elaborate on the usecases why you want to nested servers in a single cluster.

kfox1111:
Each server may be configured very differently. One is the root server which should be tightly locked down. While the second is for the k8s cluster.

The pr is not primarily about having 2 at once, but does enable that use case.

Its more about being able to test that having an upstream spire for your nested spire deployment works.

charts/spire/charts/spire-server/templates/configmap.yaml
kfox1111: We could make it variable to match the upstream csi driver deployment, yeah. In theory the upstream driver deployment is only used for this one workload though, so changing it in the upstream / making it configurable here doesn't make too much sense to me?

kfox1111:
Here's another use case: https://github.com/spiffe/spire-controller-manager/issues/76

.github/tests/nested/deps-values.yaml
marcofranssen, line 1: Rename file to `spire-server-upstream-values.yaml` if I get the idea if this install right?

.github/tests/nested/values.yaml
```diff
@@ -0,0 +1,16 @@
...
+  enabled: true
+  upstreamAuthority:
+    spire:
+      enabled: true
+      server:
+        address: spire-server.nested-deps
+  controllerManager:
+    enabled: true
+    identities:
+      spiffeIDTemplate: spiffe://{{ .TrustDomain }}/k8s/example-cluster/ns/{{ .PodMeta.Namespace }}/sa/{{ .PodSpec.ServiceAccountName }}
```
marcofranssen, line 11: Do we have to override this? Is the default value not good enough?

charts/spire/charts/spire-server/templates/configmap.yaml
marcofranssen: Probably I'm confused due to the 3 helm installs.

.github/tests/nested/pre-install.sh
kfox1111: I'll rework those bits.

.github/tests/nested/deps-values.yaml
kfox1111: or spire-root-server-values.yaml to use the terminology in their documentation?

.github/tests/nested/deps2-values.yaml
```diff
@@ -0,0 +1,21 @@
...
+  bundleConfigMap: spire-bundle-upstream
+  healthChecks:
+    port: 9981
+
+spiffe-csi-driver:
+  enabled: true
+  pluginName: upstream.csi.spiffe.io
+  agentSocketPath: /run/spire/agent-sockets-upstream/spire-agent.sock
+  healthChecks:
+    port: 9810
```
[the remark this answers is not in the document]
kfox1111: Because the test is trying to reproduce what someone might actually try to deploy on their systems. IE, root server (somewhere else) and a fully functional nested spire server (upstream agent, nested-spire-server, downstream agent, + the rest of the plumbing needed to make it work). Currently you can't deploy the upstream agent with the nested-spire-server so its a separate deployment for that piece. Follow-on pr that allows that to be merged together.

.github/tests/nested/values.yaml
kfox1111: We dont have to override it, no. Was testing it, as when you want to use nested spire, your likely to want to have multiple k8s clusters and each would hae to have a unique cluster name like example-cluster in its template. It could be removed. But its kind of useful documentation too, so not sure where it should go.

kfox1111: We may also want to template that out by default so whatever the clustername is set to, will automatically get it set in the template so this would pretty much never need an override?

.github/tests/nested/spire-root-server-values.yaml
```diff
@@ -0,0 +1,18 @@
+spire-server:
+  controllerManager:
+    enabled: false
+  nodeAttestor:
+    k8sPsat:
+      serviceAccountAllowList:
+      - "nested:spire-agent-upstream"
```
marcofranssen, line 7:
```suggestion
        - nested:spire-agent-upstream
```

.github/tests/nested/values.yaml
```diff
@@ -0,0 +1,16 @@
...
+  enabled: true
+  upstreamAuthority:
+    spire:
+      enabled: true
+      server:
+        address: spire-server.nested-root-server
+  controllerManager:
+    enabled: true
+    identities:
+      spiffeIDTemplate: spiffe://{{ .TrustDomain }}/k8s/{{ .ClusterName }}/ns/{{ .PodMeta.Namespace }}/sa/{{ .PodSpec.ServiceAccountName }}
```
marcofranssen, line 11: This might be the default if I'm not mistaking 🤔

marcofranssen (changes requested):

.github/tests/nested/values.yaml
kfox1111: The default does not have .ClusterName in it. Otherwise the same.

We might consider changing the default to .ClusterName to make it easier from going from basic to federation/tested later smoother?

kfox1111: Through chatting with various folks and looking at some other projects defaults, I think this template is still probably best for nested and not for other cases.

kfox1111:
Rebased and flattened.

edwbuck:
@kfox1111 I understand this is a "larger request" that is being continually broken down into smaller requests. 

@marcofranssen @faisal-memon I would normally call for this to be closed, as it is being handled in other merge requests, but we started down this path of using this as the "main body" of items to "carve off our next mergeable piece" so let's follow that pattern until this is down to the "last bite" size.

edwbuck:
@faisal-memon asked

> I find it very hard to follow this PR. Why would you want to nest spire servers on the same cluster? 

I did a very large nested SPIRE deployment.  I eventually nested SPIRE servers on the same cluster.  There are advantages to doing so:

- The root cluster is managed as one cluster, and is simplified to never serve workload IDs directly
- The agents of the root cluster are managed in the children cluster(s).  Each child thus has "upstream configuration information" necessary to configure the root-cluster-agents, which are co-deployed in the secondary cluster's "server" PODs.
- By dividing the clusters up in a tiered deployment to have "parent clusters" and "child clusters", one can then deploy one parent and one child into a single k8s cluster to service workload ID requests in the same cluster.  Initially this seems odd, but it has certain advantages:
  - The failure behavior of all workload ID issuing clusters is identical, allowing for aligned procedures on them.
  - The monitoring of the root cluster can be held in higher priority, as it will be known that any root cluster failure is not trivial (scoped to a workload id request) but is critical (scoped to an intermediate CA issuance)
  - The logic in the helm templates for the clusters is greatly simplified, as one only has "one kind" of root cluster, and a child cluster that can be configured one of two ways.  The child cluster being configured one of two ways is unavoidable, but the "one kind" of root cluster prevents logic for agents / workload helm template updates from destabilizing the root cluster.

faisal-memon:
> @faisal-memon asked
> 
> > I find it very hard to follow this PR. Why would you want to nest spire servers on the same cluster?

Actually that was @marcofranssen, though I share the same concerns. I would like to take in just the config needed to support a nested deployment and not do the embedded deployment. Instead you can redeploy the chart with a different release name, nameoverrides, etc. and necessary config to get the same effect. 

Deploying the upstream agent and csi driver within a single chart seems out of scope to me. Similar to deploying databases or ingress controllers.

kfox1111:
I think the current plan has https://github.com/spiffe/helm-charts/pull/260 merging before this as a prereq.

faisal-memon:
> I think the current plan has #260 merging before this as a prereq.

We dont need experimental flag or the code to deploy an upstream spire server/agent in a single chart. That to me is out of scope, like deploying a database. Just the Upstream Authority code is sufficient.

kfox1111:
so, just this patch 'charts/spire/charts/spire-server/templates/configmap.yaml' here:
https://github.com/spiffe/helm-charts/pull/117/files#diff-95a921e948f7af4dcca64089a1d722399e1ab49bf5b42f420c6ee4f42c3cdd25

In its own pr?

faisal-memon:
> What's the plan for this PR? I would like to see this get merged

Hey @matthewdevenny. Thanks for chiming in on this one. Its been in limbo for a while. Do you have a nested spire deployment within the same cluster?

matthewdevenny:
I’m working on a nested spire architecture where I have a centralized root spire server running on a management k8s cluster and then N edge clusters with a nested spire server on each cluster. I do think the nested agent could just be an additional container on the nested server pod rather than a complete daemonset though. 

 I would also like to see some of the other ideas @kfox1111 is pitching in other PRs for complete control over the configuration files. In my use case I’m also using the BoxBoat fork of the spire-tpm-plugin for node attestation.

kfox1111:
@matthewdevenny sounds like your trying to do basically the same thing I am.

We can't just add one upstream container to the downstream csi driver / agent daemonsets as the other support containers are only able to talk to one driver at a time. So we would basically have to duplicate all of the containers in each pod of the daemonset. That may be ok to do but would be less simple then just deploying the driver/agent a second time in separate daemonsets I think. Is there a drawback to having an extra set of daemonsets?

It also would not allow you to deploy the agents/drivers with separate configurations. For example, I'm thinking about deploying the upstream agent/driver and spire-server only on the control plane nodes with a nodeselector/toleration, but allowing the downstream agent/driver onto all nodes. Not sure how we could do such a deployment without a separate daemonset. Is that a feature you would want to use as well?

faisal-memon:
> I’m working on a nested spire architecture where I have a centralized root spire server running on a management k8s cluster and then N edge clusters with a nested spire server on each cluster. I do think the nested agent could just be an additional container on the nested server pod rather than a complete daemonset though.
> 

Thanks @matthewdevenny. Having the nested agent inside the spire server pod is an interesting idea. Would that eliminate the need for the extra csi driver?

kfox1111:
Oh. I misread and didn't notice it was the spire-server that would get the upstream agent. Sorry. Thats an interesting idea.

That could be made to work...

It has some benefits:
* less resources needed
* doesn't need a csi driver at all, can put the agent right in the spire-server pod

But would have the following drawbacks
* Config combines some of whats needed for the agent into the server pod too. This complicates config quite a bit.
* duplicate code between agent and spire-server. could somewhat be mitigated by moving the code out to spire-lib
* Would require much looser restrictions on the spire-server pod (PodSecurityStandards) so that it could reach the host for things like tpms. This would loosen up the security quite a bit.
* May need net=host to properly get host identity that the agent may need. Also would require antiaffinity so no two pods land on the same host so two agents don't get the same identity

I think for my own deployment, I'd prefer the separate upstream csi driver / agent in order to keep the tight security.

But, these ideas are not an either or thing. We could support both options too if there was an interest to.

matthewdevenny:
I'm in the early stages of figuring out how I want to deploy the nested spire server... 

As far as the security goes for the spire-server with the nested-agent running alongside - I think you would only need to have those additional permissions set on the nested-agent container securityContext. The overall podSecurityContext could remain the same. I'm also using a device plugin to expose the tpm to the spire-agent, rather than mounting the device directly on the pod see: https://github.com/boxboat/k8s-tpm-device

kfox1111:
The https://github.com/boxboat/k8s-tpm-device device plugin is an interesting way of going about it....

That would indeed lessen the permissions needed for the PSS. But I think it would still need to run as root?, which would disallow Restricted.

I think you'd also still want to only run one per host and have a stable identity so you can register the entries on the root server. So, net=host with antiaffinity or daemonset support for spire-server would still likely be a good idea too?

faisal-memon:
Thanks @kfox1111 for the breakdown. How do you handle the anti affinity issue with a separate daemonset?

kfox1111:
> Thanks @kfox1111 for the breakdown. How do you handle the anti affinity issue with a separate daemonset?

Daemonsets are naturally anti-affinity (one pod per host) by design

faisal-memon:
> > Thanks @kfox1111 for the breakdown. How do you handle the anti affinity issue with a separate daemonset?
> 
> Daemonsets are naturally anti-affinity (one pod per host) by design

Yes, though in this cause you would have 2 separate ones. Can you can still get into the scenario where there are 2 agents and the server running on the same node?

kfox1111:
two agents and one server is the way this pr is designed. one upstream agent pod, one server, one downstream agent pod.

workloads need to talk to a downstream agent pod. the server needs to be able to talk to an upstream agent.

if you wanted to combine the upstream agent and spire-server into the same pod, then it shouldn't run two spire-server pods on the same host or you would get multiple upstream spire-agents on the same host.

faisal-memon:
@kfox1111 Thanks for clarifying. There's definitely a collision. Can possibly use PodAffinity to avoid it: https://platform9.com/kb/kubernetes/how-to-avoid-scheduling-more-that-one-replicas-of-a-pod-on-same

In any case Im fine with either approach. @matthewdevenny Do you have a preference?

matthewdevenny:
The only reason I was thinking to combine them into the server pod was because it seemed like a waste of resources to run a root agent on nodes where it is not serving any purpose. 

That being said, I could go either way - the agent container is pretty lightweight but if resources are at a premium (think edge compute) then consolidation might make the most sense. The anti-affinity of the `spire-server` in the case of multiple replicas is probably going to be something you would want to do in either solution.

kfox1111:
> The only reason I was thinking to combine them into the server pod was because it seemed like a waste of resources to run a root agent on nodes where it is not serving any purpose.

That could be limited by restricting the spire-agent/csi driver daemonsets to a smaller set of nodes, but your right. It will always take a bit more resource then embedding it within the spire-server.

> That being said, I could go either way - the agent container is pretty lightweight but if resources are at a premium (think edge compute) then consolidation might make the most sense. The anti-affinity of the `spire-server` in the case of multiple replicas is probably going to be something you would want to do in either solution.

Yeah. I can definitely see the benefit of that approach. Anti-affinity wouldn't necessarily be needed if you had multiple spire-servers on one node with a shared spire-agent.

The separate daemonsets is a lot less work for us to maintain, in that we just reuse the existing charts a second time without changes to the charts. With the embeded spire-agent in the server, we would need to template out the spire-agent configmap and containerspec in a way we could share it with both charts. So, would be a lot more work to do.

Is that something that reviewers would support?

kfox1111:
Ok, I think everything has been refereshed.

charts/spire/values.yaml
```diff
@@ -41,10 +41,30 @@ spire-agent:
   enabled: true
   nameOverride: agent
 
+upstream:
+  # -- enable upstream csi driver and agent for use with nested spire.
+  enabled: false
```
faisal-memon, line 46: I think the enabled flag should be moved to the `upstream-spire-agent` and `upstream-csi-driver` sections, or move the csi driver and agent under `upstream:`

```
upstream:
   enabled:false
   spire-agent: ...
   spiffe-csi-driver: ...
```

charts/spire/README.md
```diff
@@ -122,6 +124,15 @@ Now you can interact with the Spire agent socket from your own application. The
...
 | tornjak-frontend.enabled | bool | `false` | Enables deployment of Tornjak frontend/UI (Not for production) |
+| upstream-spiffe-csi-driver.agentSocketPath | string | `"/run/spire/agent-sockets-upstream/spire-agent.sock"` |  |
+| upstream-spiffe-csi-driver.healthChecks.port | int | `9810` |  |
+| upstream-spiffe-csi-driver.pluginName | string | `"upstream.csi.spiffe.io"` |  |
+| upstream-spire-agent.bundleConfigMap | string | `"spire-bundle-upstream"` |  |
+| upstream-spire-agent.healthChecks.port | int | `9981` |  |
+| upstream-spire-agent.nameOverride | string | `"agent-upstream"` |  |
+| upstream-spire-agent.serviceAccount.name | string | `"spire-agent-upstream"` |  |
+| upstream-spire-agent.socketPath | string | `"/run/spire/agent-sockets-upstream/spire-agent.sock"` |  |
+| upstream.enabled | bool | `false` | enable upstream csi driver and agent for use with nested spire. |
```
faisal-memon, line 135: This section seems mostly duplicated with the updated values in the root yaml.  Is that a helm docs thing?

kfox1111: Yes, helm docs thing.

faisal-memon (approved):

charts/spire/charts/spire-server/templates/statefulset.yaml
```diff
@@ -269,6 +274,12 @@ spec:
           secret:
             secretName: {{ include "spire-server.upstream-ca-secret" . }}
         {{- end }}
+        {{- if eq (.Values.upstreamAuthority.spire.enabled | toString) "true" }}
```
edwbuck, line 277: This part should be controlled on if CSI drivers are being used, not if nested clusters are used

charts/spire/charts/spire-server/templates/statefulset.yaml
```diff
@@ -125,6 +125,11 @@ spec:
               mountPath: /run/spire/upstream_ca
               readOnly: false
             {{ end }}
+            {{- if eq (.Values.upstreamAuthority.spire.enabled | toString) "true" }}
```
edwbuck, line 128: This would fail some conventions that people might use to say that something is not enabled.

For example, as a boolean value, zero values evaulate to false, and non-zero values evaluate to true.  This means that

| false, 0, ""

evaluate to false while

| true, (any number), (any non-empty string)

evaluates to true.

This means that our "check" for true is stronger than the checks the template package that Helm is built on top of does.  This could lead to the unexpected behavior of

| upstreamAuthority.spire.enabled = "True" 

failing because we are doing a string comparison, and "True" != "true" yields the Boolean false.  When if we didn't do string comparison, the logic would be "True" is non-zero, yielding the Boolean value true

Strongly consider using Boolean evaluation instead of string matching.

charts/spire/charts/spire-server/values.yaml
```diff
@@ -209,6 +209,7 @@ upstreamAuthority:
       renewBefore: ""
   spire:
     enabled: false
+    upstreamDriver: upstream.csi.spiffe.io
```
edwbuck, line 212: Here's the value I would expect there to be no default value.  That would mean it would not exist in the values.yaml file here; but a _helpers.tpl would check it exists when the "enabled" flag is set to true.

This forces the value to be provided in a values.yaml file manually, or via the cli manually, preventing any kind of "default" value.

charts/spire/values.yaml
```diff
@@ -41,10 +41,33 @@ spire-agent:
   enabled: true
   nameOverride: agent
 
+upstream-spire-agent:
```
edwbuck, line 44: It seems to me that by having the agent and the upstream CA defined in different places, that we are not designing a user interface that concentrates the concerns into the same location.

If I am using nested spire, I should be able to specify the sensible shared items for the upstream agent and upstream CA items under a locally grouped item, the "upstream system".  Minimally I would expect them both to be enabled simultaneously and disabled simultaneously.

examples/nested/run-tests.sh
```diff
@@ -0,0 +1,51 @@
...
+"${helm_install[@]}" spire charts/spire \
+  --namespace spire-root-server \
+  --values "${DEPS}/spire-root-server-values.yaml" \
+  --wait
+
+kubectl get nodes -o go-template='{{range .items}}{{printf "%s\n" .metadata.uid}}{{end}}' | while read -r line; do
+  kubectl exec -t spire-server-0 -n "spire-root-server" -- spire-server entry create -spiffeID spiffe://example.org/example-cluster/nested-spire -parentID "spiffe://example.org/spire/agent/k8s_psat/example-cluster/$line" -selector k8s:pod-label:app.kubernetes.io/name:server -downstream
+done
+
+"${helm_install[@]}" --namespace "${ns}" --values "${SCRIPTPATH}/values.yaml,${SCRIPTPATH}/../production/values.yaml" \
```
edwbuck, line 42: De-template this.  It makes it unclear that this is a helm upgrade (upsert) command, and makes it doubly unclear that this is the child cluster.  

In addition, de-template the namespace "ns" parameter as we are dealing with two namespaces.  If it must be held in a variable, it should be "child_namespace".  Additionally, if we are holding the namespace in a variable for the child, we MUST hold the parent namespace in a "parent_namespace" variable to match, and ideally would set both of those constants together near the top of the file.

I'm not 100% convinced that the need to template the namespace is by holding it in a variable is valuable; as the decision to do so creates issues, such as the issue of whether we wish to create a test that has two child clusters, in which case child_namespace might better be an array, and that then raises the question of how to make multiple arrays coordinate (Associate arrays with the cluster name as the key?  Indexed arrays?)

examples/nested/run-tests.sh
```diff
@@ -0,0 +1,51 @@
...
+  --wait spire charts/spire
+helm test --namespace "${ns}" spire
+
+print_helm_releases
+print_spire_workload_status "${ns}"
```
edwbuck, line 47: Does this need some sort of extension to handle parent cluster reporting?  Possibly in a different bash function?

examples/nested/run-tests.sh
```diff
@@ -0,0 +1,51 @@
...
+
+if [[ "$1" -ne 0 ]]; then
+  get_namespace_details "${ns}"
```
edwbuck, line 50: We should get namespace details for the parent too, perhaps in a get_parent_namespace_details function (as it might differ from that of a typical child cluster)

examples/nested/values.yaml
```diff
@@ -0,0 +1,24 @@
+spire-server:
+  enabled: true
+  upstreamAuthority:
+    spire:
+      enabled: true
+      server:
+        address: spire-server.spire-root-server
```
edwbuck, line 7: This value is duplicated on line 24, leading to the ability to get it wrong in one or both places.  This is the example that proves the design of the YAML values is wrong.

examples/nested/values.yaml
```diff
@@ -0,0 +1,24 @@
+spire-server:
+  enabled: true
+  upstreamAuthority:
+    spire:
+      enabled: true
```
edwbuck, line 5: This value is duplicated in line 22, leading to the ability to get it wrong in one or both places.  This is the example that proves the design of the YAML values is wrong.

examples/nested/values.yaml
```diff
@@ -0,0 +1,24 @@
...
+    identities:
+      spiffeIDTemplate: spiffe://{{ .TrustDomain }}/k8s/{{ .ClusterName }}/ns/{{ .PodMeta.Namespace }}/sa/{{ .PodSpec.ServiceAccountName }}
+
+spiffe-oidc-discovery-provider:
+  enabled: true
+  insecureScheme:
+    enabled: true
+
+upstream-spiffe-csi-driver:
+  enabled: true
```
edwbuck, line 19: This may need some thought to properly handle enablement / disablement based on the upstream spire CA being enabled.  We don't want to advertise that it can be enabled in the values.yaml when the upstream spire CA is disabled, which indicates that a control flag designed this way is wrong, as it allows mismatches of settings that lead to nonsensical deployments.

I would do a Boolean truth table to be constructed, indicating the valid settings and then the values.yaml to be modified to have settings that only permit the valid combinations to be reached.

examples/production/values.yaml
```diff
@@ -46,6 +46,11 @@ spiffe-csi-driver:
   namespaceOverride: spire-system
   priorityClassName: system-node-critical
 
+
+upstream-spiffe-csi-driver:
+  namespaceOverride: spire-system
```
edwbuck, line 51: This implies an override, which implies a value that is being changed.  There is little to no value in indicating that it overrides something, as Helm is a declarative tool.  We should just "declare" the namespace, and not a policy of how to apply it.

examples/production/values.yaml
```diff
@@ -69,6 +74,24 @@ spire-agent:
 
   logLevel: info
 
+upstream-spire-agent:
+  namespaceOverride: spire-system
```
edwbuck, line 78: Again with the overrides....

edwbuck (changes requested):
Many small comments.  Not every one needs to be honored, but either attempts to honor them or explanations on why they can't be honored will speed the review.

charts/spire/Chart.yaml
```diff
@@ -30,10 +30,20 @@ dependencies:
...
+  - name: spire-agent
+    alias: upstream-spire-agent
+    condition: upstream-spire-agent.enabled
+    repository: file://./charts/spire-agent
+    version: 0.1.0
   - name: spiffe-csi-driver
     condition: spiffe-csi-driver.enabled
     repository: file://./charts/spiffe-csi-driver
     version: 0.1.0
+  - name: spiffe-csi-driver
```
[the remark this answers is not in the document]
edwbuck: I received a sufficient explanation for the addition of the CSI driver.

charts/spire/README.md
```diff
@@ -468,5 +482,95 @@ Now you can interact with the Spire agent socket from your own application. The
 | tornjak-frontend.startupProbe.timeoutSeconds | int | `5` | Timeout seconds for startupProbe |
 | tornjak-frontend.tolerations | list | `[]` |  |
 | tornjak-frontend.topologySpreadConstraints | list | `[]` |  |
+| upstream-spiffe-csi-driver.agentSocketPath | string | `"/run/spire/agent-sockets/spire-agent.sock"` | The unix socket path to the spire-agent |
```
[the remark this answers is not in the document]
edwbuck: There is a different PR and effort to clean up the documentation of the entries

examples/production/values.yaml
edwbuck: Please break this out into an issue on the appropriate helm chart and don't attempt to handle this in this PR

examples/production/values.yaml
edwbuck: This implies an override, which implies a value that is being changed. There is little to no value in indicating that it overrides something, as Helm is a declarative tool. We should just "declare" the namespace, and not a policy of how to apply it.

Please break this out into an issue on the appropriate helm chart and don't attempt to handle this in this PR

charts/spire/values.yaml
```diff
@@ -41,10 +41,33 @@ spire-agent:
...
+  # -- enable upstream agent for use with nested spire.
+  enabled: false
```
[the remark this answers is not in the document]
kfox1111: Cant do that with nested charts. They only support booleans, not tests. Helm limitation.

charts/spire/values.yaml
```diff
@@ -41,10 +41,33 @@ spire-agent:
...
+    port: 9981
+  telemetry:
+    prometheus:
+      port: 9989
+
 spiffe-csi-driver:
   # -- Enables deployment of CSI driver
   enabled: true
 
+upstream-spiffe-csi-driver:
```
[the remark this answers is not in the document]
kfox1111: In our conversation, we came to the realization that having support for both an external agent and an internal one in the spire-server pod are both needed use cases. This pr is only focusing on the former. It is orthogonal to the later.

examples/nested/values.yaml
kfox1111: Has to be done this way due to limitations in how helm chart nesting works. Helm isn't flexible enough to do that unfortunately.

examples/production/values.yaml
kfox1111: I'm not creating the namespaceOverride here. I'm simply using it, appropriately. Changing the name of namespaceOverride to be something else is out of scope for this PR and should be done as a separate issue.

examples/production/values.yaml
kfox1111: I'm not creating the namespaceOverride here. I'm simply using it, appropriately. Changing the name of namespaceOverride to be something else is out of scope for this PR and should be done as a separate issue.

examples/nested/run-tests.sh
```diff
@@ -0,0 +1,51 @@
...
+  helm uninstall --namespace "${ns}" spire 2>/dev/null || true
+  kubectl delete ns "${ns}" 2>/dev/null || true
+
+  helm uninstall --namespace mysql spire-root-server 2>/dev/null || true
+  kubectl delete ns spire-root-server 2>/dev/null || true
+}
+
+trap 'trap - SIGTERM && teardown' SIGINT SIGTERM EXIT
+
+kubectl create namespace spire-system 2>/dev/null || true
```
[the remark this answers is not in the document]
kfox1111: This is a common pattern throughout the chart. I actually mostly just copied this file and tweaked a few things. We probably need an issue to fix these sorts of things more globally. that being said, I'll fix this one as you request.

examples/production/values.yaml
```diff
@@ -46,6 +46,11 @@ spiffe-csi-driver:
   namespaceOverride: spire-system
   priorityClassName: system-node-critical
 
+
```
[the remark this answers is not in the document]
kfox1111: because the server needs to talk to the agent associated with the upstream spire server. since its in a separate pod, it needs its own csi driver. Its needed for security reasons.

charts/spire/charts/spire-server/templates/pre-delete-hook.yaml
```diff
@@ -0,0 +1,75 @@
...
+      restartPolicy: Never
+      serviceAccountName: {{ include "spire-server.serviceAccountName" . }}-pre-delete
+      securityContext:
+        {{- toYaml .Values.podSecurityContext | nindent 8 }}
+      containers:
+      - name: pre-delete-job
+        securityContext:
+          {{- toYaml .Values.securityContext | nindent 10 }}
+        image: {{ template "spire-lib.kubectl-image" (dict "appVersion" $.Chart.AppVersion "image" .Values.tools.kubectl.image "global" .Values.global "KubeVersion" .Capabilities.KubeVersion.Version) }}
+        args:
```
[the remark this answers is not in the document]
kfox1111: helm install from a helm delete would be really hard to do reliably since you don't have access to the chart (in case of offline clusters) and there are likely locking issues involved. Will work on the error message route.

examples/production/values.yaml
edwbuck: **Please break this out into an issue on the appropriate helm chart and don't attempt to handle this in this PR**

examples/production/values.yaml
edwbuck: **Please break this out into an issue on the appropriate helm chart and don't attempt to handle this in this PR**

examples/nested/run-tests.sh
```diff
@@ -0,0 +1,51 @@
...
+kubectl label namespace spire-system pod-security.kubernetes.io/enforce=privileged || true
+kubectl create namespace "${ns}" 2>/dev/null || true
+kubectl label namespace "${ns}" pod-security.kubernetes.io/enforce=restricted || true
+
+"${helm_install[@]}" spire charts/spire \
```
edwbuck, line 33: Avoid hiding the command inside of variables.  The options and flags that are passed become silent and no hint is given to the reader at the point where it is used that they exist.

If you want to ensure that a particular option is present, or might be present then 

```
helm install ${NAMESPACE:+--namespace} ${NAMESPACE}
```

reads much better.

examples/production/values.yaml
kfox1111: I'm not creating the namespaceOverride here. I'm simply using it, appropriately. Its required for proper deployment.

Changing the name of namespaceOverride to be something else is out of scope for this PR and should be done as a separate issue.

If you would like to rename the namespaceOverride value, please file an issue for it.

examples/production/values.yaml
kfox1111: I'm not creating the namespaceOverride here. I'm simply using it, appropriately. Its required for proper deployment.

Changing the name of namespaceOverride to be something else is out of scope for this PR and should be done as a separate issue.

If you would like to rename the namespaceOverride value, please file an issue for it.

charts/spire/charts/spire-server/templates/pre-delete-hook.yaml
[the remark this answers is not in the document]
kfox1111: I tried to think of a way to do either of these. Can't think of a reliable way. The problem already exists in tree. I filed an issue and I think we shouldn't let this block us since we have it other places. https://github.com/spiffe/helm-charts/issues/421

examples/nested/run-tests.sh
kfox1111: This is a common pattern throughout the chart. I actually mostly just copied this file and tweaked a few things. We probably need an issue to fix these sorts of things more globally. that being said, I'll fix this one as you request.

charts/spire/charts/spire-server/README.md
```diff
@@ -238,5 +238,6 @@ In order to run Tornjak with simple HTTP Connection only, make sure you don't cr
 | upstreamAuthority.spire.enabled | bool | `false` |  |
 | upstreamAuthority.spire.server.address | string | `""` |  |
 | upstreamAuthority.spire.server.port | int | `8081` |  |
+| upstreamAuthority.spire.upstreamDriver | string | `"upstream.csi.spiffe.io"` |  |
```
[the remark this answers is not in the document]
kfox1111: Done

examples/nested/run-tests.sh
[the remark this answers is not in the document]
kfox1111: Done

charts/spire/charts/spire-server/templates/statefulset.yaml
kfox1111: Done

charts/spire/charts/spire-server/templates/statefulset.yaml
kfox1111: Overcome by differences in values now. Done.

charts/spire/charts/spire-server/values.yaml
kfox1111: Done

examples/nested/run-tests.sh
kfox1111: Done.

examples/nested/run-tests.sh
kfox1111: Done

examples/nested/run-tests.sh
kfox1111: Done

examples/nested/values.yaml
kfox1111: Whether to have an upstream spire authority config can be separate from deploying it with this chart.

examples/nested/values.yaml
kfox1111: Done

examples/nested/values.yaml
kfox1111: Done

charts/spire/values.yaml
kfox1111: Updated to be enabled simultaneously.

kfox1111:
@edwbuck 

I think I incorporated all the review items I could. Should be good for a re-review.

faisal-memon:
@edwbuck Do you have time to look at this again?

project/issues_due_to_nesting.md
```diff
@@ -0,0 +1 @@
+Nested charts only support enablement by boolean, not by complex tests such as string compares.
```
marcofranssen, line 1: What does this mean? Shouldn't this move into the FAQ, preferably with a more explanatory description, I have absolutely no clue what this statement means and why we need a dedicated MD file for this.

kfox1111: Charts.yaml -> dependency -> condition property can only reference a boolean value, not a complex test such as
```
if eq .Values.type "secondary"
```

kfox1111: The FAQ is intended for users. This file seems to be intended for contributors.

edwbuck (approved):
With the changes, this is as good as it is going to get until we un-nest our core application.

project/issues_due_to_nesting.md
marcofranssen: Ok getting the point this is contributors driven. Discussed during the meeting, Find it hard to grasp the intent/meaning of that single sentence. Was hoping it could be improved so it is more clear to any reader what we mean by this single sentence in this isolated file that doesn't provide additional context.

kfox1111: We can fix it, but lets do that in a followon, this has been WAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAY too long in review already.

marcofranssen: Not a fan of fixing unclear things in followup PRs, the whole point of a review is to adress them in the PR, but trusting you on creating that doc improvement PR as followup.

kfox1111: https://github.com/spiffe/helm-charts/pull/450 filed as promised.

PR #118, opened 2023-03-13
Update nested spire to work with extra built in driver/agent

kfox1111:
This pr updates the nested spire work so that a separate deployment
of the chart just for upstream support isn't needed.


charts/spire/values.yaml
```diff
@@ -19,17 +19,38 @@ spire-server:
   controllerManager:
     enabled: true
 
-spire-agent:
+agent-downstream:
```
marcofranssen, line 22: I prefer not to rename this.

charts/spire/values.yaml
```diff
@@ -19,17 +19,38 @@ spire-server:
...
+  clusterName: *clusterName
+  trustDomain: *trustDomain
+
+  socketPath: /run/spire/agent-sockets-upstream/spire-agent.sock
+  serviceAccount:
+    name: spire-agent-upstream
+  healthChecks:
+    port: 9981
+
+driver-downstream:
```
marcofranssen, line 44: Also here I prefer the old name.

charts/spire/values.yaml
```diff
@@ -19,17 +19,38 @@ spire-server:
...
+agent-downstream:
   enabled: true
   nameOverride: agent
   bundleConfigMap: *bundleConfigMap
 
   clusterName: *clusterName
   trustDomain: *trustDomain
 
-spiffe-csi-driver:
+agent-upstream:
```
marcofranssen, line 30:
```suggestion
upstream-spire-agent:
```

charts/spire/values.yaml
```diff
@@ -19,17 +19,38 @@ spire-server:
...
+  socketPath: /run/spire/agent-sockets-upstream/spire-agent.sock
+  serviceAccount:
+    name: spire-agent-upstream
+  healthChecks:
+    port: 9981
+
+driver-downstream:
   enabled: true
 
+driver-upstream:
```
marcofranssen, line 47: upstream-spiffe-csi-driver:

examples/production/values.yaml
```diff
@@ -47,12 +47,12 @@ spire-server:
       seccompProfile:
         type: RuntimeDefault
 
-spiffe-csi-driver:
+driver-downstream:
```
marcofranssen, line 50:
```suggestion
spiffe-csi-driver:
```

examples/production/values.yaml
```diff
@@ -47,12 +47,12 @@ spire-server:
...
   enabled: true
   namespaceOverride: spire-system
   priorityClassName: system-node-critical
 
-spire-agent:
+agent-downstream:
```
marcofranssen, line 55:
```suggestion
spire-agent:
```

charts/spire/Chart.yaml
```diff
@@ -23,11 +23,23 @@ dependencies:
     repository: file://./charts/spire-server
     version: 0.1.0
   - name: spire-agent
-    condition: spire-agent.enabled
+    alias: agent-downstream
+    condition: agent-downstream.enabled
```
marcofranssen, line 27:
```suggestion
    condition: spire-agent.enabled
```

charts/spire/Chart.yaml
```diff
@@ -23,11 +23,23 @@ dependencies:
...
     repository: file://./charts/spire-agent
     version: 0.1.0
+  - name: spire-agent
+    alias: agent-upstream
+    condition: agent-upstream.enabled
+    repository: file://./charts/spire-agent
+    version: 0.1.0
```
marcofranssen, line 34:
```suggestion
  - name: spire-agent
    alias: upstream-spire-agent
    condition: upstream-spire-agent.enabled
    repository: file://./charts/spire-agent
    version: 0.1.0
```

charts/spire/Chart.yaml
```diff
@@ -23,11 +23,23 @@ dependencies:
...
+  - name: spiffe-csi-driver
+    alias: driver-downstream
+    condition: driver-downstream.enabled
```
marcofranssen, line 37:
```suggestion
    condition: spiffe-csi-driver.enabled
```

charts/spire/Chart.yaml
```diff
@@ -23,11 +23,23 @@ dependencies:
...
+    repository: file://./charts/spiffe-csi-driver
+    version: 0.1.0
   - name: spiffe-csi-driver
-    condition: spiffe-csi-driver.enabled
+    alias: driver-upstream
+    condition: driver-upstream.enabled
```
marcofranssen, line 42:
```suggestion
    alias: upstream-spiffe-csi-driver
    condition: upstream-spiffe-csi-driver.enabled
```

marcofranssen (changes requested):
Couple of suggestions on naming. Basically I would not call anything downstream as that suggests it's something outside of the context of this chart install.

Downstream IMHO is inside the context of the current chart install and therefore doesn't require any additional naming.

For upstream it does make sense, I really like this approach.

One question in general? Why using the alias instead of the name? Isn't specifying the name as `upstream-spire-agent` sufficient?

charts/spire/values.yaml
kfox1111: so, there would be agent and agent-upstream?

kfox1111:
> Couple of suggestions on naming. Basically I would not call anything downstream as that suggests it's something outside of the context of this chart install.

kk


> Downstream IMHO is inside the context of the current chart install and therefore doesn't require any additional naming.
> 
> For upstream it does make sense, I really like this approach.
> 
> One question in general? Why using the alias instead of the name? Isn't specifying the name as `upstream-spire-agent` sufficient?

The name is the name of the chart to include. so has to be spire-agent to reuse that chart. alias is what to name it from the perspective of the values file.

kfox1111:
Rebased and flattened.

charts/spire/values.yaml
```diff
@@ -27,9 +27,32 @@ spire-agent:
...
+  bundleConfigMap: spire-bundle-upstream
+
+  clusterName: *clusterName
+  trustDomain: *trustDomain
+
+  socketPath: /run/spire/agent-sockets-upstream/spire-agent.sock
+  serviceAccount:
+    name: spire-agent-upstream
+  healthChecks:
+    port: 9981
```
marcofranssen, line 45:
```suggestion
upstream:
  # -- enable upstream csi driver and agent for use with nested spire.
  enabled: false

  spire-agent:
    nameOverride: agent-upstream
    bundleConfigMap: spire-bundle-upstream

    clusterName: *clusterName
    trustDomain: *trustDomain

    socketPath: /run/spire/agent-sockets-upstream/spire-agent.sock
    serviceAccount:
      name: spire-agent-upstream
    healthChecks:
      port: 9981

  spiffe-csi-driver:
    pluginName: upstream.csi.spiffe.io
    agentSocketPath: /run/spire/agent-sockets-upstream/spire-agent.sock
    healthChecks:
      port: 9810
```

charts/spire/values.yaml
```diff
@@ -27,9 +27,32 @@ spire-agent:
...
+
 spiffe-csi-driver:
   enabled: true
 
+upstream-spiffe-csi-driver:
+  pluginName: upstream.csi.spiffe.io
+  agentSocketPath: /run/spire/agent-sockets-upstream/spire-agent.sock
+  healthChecks:
+    port: 9810
+
```
marcofranssen, line 55:
```suggestion
```

marcofranssen (changes requested):
2 small suggestions to improve the API? What do you think?

charts/spire/values.yaml
kfox1111: I'd love to do the api like that, but there is a problem. You can only nest charts one level per nested chart.

so to have it like that, we'd need an intermediate chart named 'upstream' under charts/spire/charts/
I think it would work, but would look just a little funny on the filesystem.

Do you have any issues with that arrangement? If not, I'll prototype it out.

marcofranssen (approved):
LGTM

PR #119, opened 2023-03-14
Improve Chart API

marcofranssen:
Because we are already in the context of spire-agent the API looks more logical to not have another 'agent' part in the name.

Furthermore to make it more clear the oidc provider only requires the name of the socket as opposed to the entire path like in the other charts I made that more explicit in the name of the value.


kfox1111 (approved):
Looks like a good improvement to me.

PR #120, opened 2023-03-14
Add documentation how to use Spire in own workloads

marcofranssen:


charts/spire/README.md
```diff
@@ -32,6 +32,43 @@ spec:
...
+ apiVersion: v1
+ kind: Pod
+ metadata:
+   name: my-app
+ spec:
+   containers:
+     - name: my-app
+       image: "my-app:latest"
+       imagePullPolicy: Always
++      env:
```
kfox1111, line 49: The ENV bit is hard to describe. It has no effect on any app I know of. So could be confusing to users when it doesn't "work"

kfox1111 (commented):
Generally looks good. one comment inline. Not sure what to do with it.

charts/spire/README.md
marcofranssen: Agree removed. This is specific to an app that uses this env var.
I also didn't use it in this example.

https://github.com/philips-labs/spiffe-vault/blob/main/charts/spiffe-vault/templates/spiffe-vault.yaml

kfox1111 (approved):
lgtm

Issue #109, opened 2023-03-10
Kubelet dir configurable


kfox1111:
Some distro's put kubelets dir somewhere else. Need to make it configurable. CSI driver in particular.

PR #123, opened 2023-03-14
Make kubelet path configurable

kfox1111:
Some distro's do not keep their kubelet path at /var/lib/kubelet

fixes: https://github.com/spiffe/helm-charts/issues/109


faisal-memon (approved):

faisal-memon:
Does this need to be configurable anywhere else?

kfox1111:
I grepped for /var/lib/kubelet. only cases I could find.

faisal-memon:
> I grepped for /var/lib/kubelet. only cases I could find.

Thanks for checking.

marcofranssen (approved):
LGTM 🚀

Issue #72, opened 2023-03-01
Upgrading an existing installation of spire helm chart fails on spire-controller-manager


marcofranssen:
This could be an installation order thing where spire-controller-manager is replaced while the CRD is being deployed at the same time. Subsequent run to perform the upgrade makes the install succeed.

```shell
$ helm  upgrade -n spire-system --install --values ../../../philips-labs/helm-charts/.local/spire-values.yaml spire ./charts/spire
Error: UPGRADE FAILED: cannot patch "spire-controller-manager-service-account-based" with kind ClusterSPIFFEID: Internal error occurred: failed calling webhook "vclusterspiffeid.kb.io": failed to call webhook: Post "https://spire-controller-manager-webhook.spire-system.svc:443/validate-spire-spiffe-io-v1alpha1-clusterspiffeid?timeout=10s": no endpoints available for service "spire-controller-manager-webhook"
$ helm  upgrade -n spire-system --install --values ../../../philips-labs/helm-charts/.local/spire-values.yaml spire ./charts/spire
Release "spire" has been upgraded. Happy Helming!
NAME: spire
LAST DEPLOYED: Wed Mar  1 11:32:51 2023
NAMESPACE: spire-system
STATUS: deployed
REVISION: 3
NOTES:
Installed spire…
```

How to reproduce.

```shell
$ helm -n spire-system uninstall spire
$ git checkout release
$ helm upgrade -n spire-system --install --values ../../../philips-labs/helm-charts/.local/spire-values.yaml spire ./charts/spire
Release "spire" has been installed. Happy Helming!
NAME: spire
LAST DEPLOYED: Wed Mar  1 11:32:51 2023
NAMESPACE: spire-system
STATUS: deployed
REVISION: 1
NOTES:
Installed spire…
$ git checkout main
$ helm upgrade -n spire-system --install --values ../../../philips-labs/helm-charts/.local/spire-values.yaml spire ./charts/spire
Error: UPGRADE FAILED: cannot patch "spire-controller-manager-service-account-based" with kind ClusterSPIFFEID: Internal error occurred: failed calling webhook "vclusterspiffeid.kb.io": failed to call webhook: Post "https://spire-controller-manager-webhook.spire-system.svc:443/validate-spire-spiffe-io-v1alpha1-clusterspiffeid?timeout=10s": no endpoints available for service "spire-controller-manager-webhook"
[...]

PR #124, opened 2023-03-14
Make webhook fail policy configurable

kfox1111:
This patch makes the webhook fail policy configurable.


faisal-memon:
Im a little hesitant on this one. Its kind of dangerous to have a crd with no protection. To resolve https://github.com/spiffe/helm-charts/issues/72 we can use the pre-upgrade hook to change it to ignore and minimize the window where the webhook is disabled.

kfox1111:
The hooks are clever, but can have unintended side effects.

helm hooks can cause pain with ci/cd systems. I try to avoid them. For example, some will just run:
helm upgrade --install -n spire spire charts/spire -f site-values.yaml

anytime anything changes in the config git repo. with the hooks, it will slow the apply down substantially because the 2 hooks always runs whether anything changes or not causing the webhook to switch back and forth between ignore and fail with each run even while nothing else changes.

Gatekeeper, which may be even more security sensitive then this chart, even defaults to ignore and they do support HA webhooks out of the box:
https://github.com/open-policy-agent/gatekeeper/blob/master/charts/gatekeeper/values.yaml#L10
https://github.com/open-policy-agent/gatekeeper/blob/master/charts/gatekeeper/values.yaml#L1

faisal-memon:
Ok with making the fail policy configurable, but lets keep the default as fail. People with more complex use cases can set it to ignore.

kfox1111:
what about leaving it as ignore for now so upgrades work, until we get the ha webhook in place, then change it back to fail?

kfox1111:
Never intended this to be a permanent thing. Just a workaround for now.

faisal-memon:
> what about leaving it as ignore for now so upgrades work, until we get the ha webhook in place, then change it back to fail?

Got it. Im working on the pre-upgrade hook today. Got sidetracked a bit yesterday. That should fix the upgrade issue.

kfox1111:
Probably need a combination of approaches then.  The pre-upgrade hook is probably ok for some folks, but I'll want to turn it off on my cluster and use this pr to set it to ignore to reduce churn.

So, if that webhook were disable possible to disable, I'll modify this pr to be default fail.

faisal-memon (approved):

PR #127, opened 2023-03-15
Replace lockdown test with production-example test

marcofranssen:
In general you could deploy spire in a namespace spire-public for the privileged containers
and override spire-server to run in a locked down namespace called spire-system
for this test case the used namespace is named lockdown see .github/workflows/helm-chart-ci.yaml:169

See https://github.com/spiffe/helm-charts/actions/runs/4471132629 for the succeeding workflow run.


kfox1111:
There is a couple things going on here....

1. testing. Trying to not have too many tests. We need a lockdown test, and we need a namespaceOverride test. Right now, the two are combined to test the install with a 3 namespace check. We could break out lockdown so it only has two namespaces,  but would have to add a whole nother set of test to test the 3 namespace config, so not really saving anything there.

2. naming. Namings hard. but, my logic:
* spire-csi-driver & spire-agent run with hostPaths. IE, they are intimately tied to the system that its running on. So, spire-system namespace makes sense to me. So is putting privilege on the spire-system namespace.
* spire-server and spiffe-oidc-server are webservers. so putting them in spire-server has a certain logic to it.

kfox1111:
also, the spire-server (stateful) service should be in the non overridden namespace by default so that its less likely to cause breakage if someone wants to go from nonlocked down to locked down. Moving out the csi-driver/agent is stateless/easy. moving the server is hard. So if we do want to test the locked down 2 namespace variant, the test is currently backwards I think.

marcofranssen:
I'm not following your comments I think. What I thought was the idea based on your original implementation is the following.

Components per namespace:

- spire-system (privileged)
  - spire-server
  - spire-controller-manager
- spire-public (non-privileged)
  - spire-agent
  - spiffe-oidc-discovery-provider
  - spiffe-csi-driver

Basically the only thing I changed is to use the `lockdown` namespace as the `non-privileged` that is what the ci workflow uses to deploy the things. In general this PR only renamed that namespace from original test implementation to actually use the one used in CI. That caused the breaking.

Could you please confirm this was intention? Or explain in the bulleted form as I did what was the intention so I can follow your thinking.

kfox1111:
Close but backwards. privilege is needed to touch the host. so csi-driver and agent need privilege. The rest can be locked down hard and be non-privileged.

* spire-system (privileged)
  * spire-agent
  * spiffe-csi-driver
* lockdown/spire-public/spire-server (non-privileged)
  * spire-server
  * spire-controller-manager
  * spiffe-oidc-discovery-provider

For your change, it would work if you s/spire-server/lockdown/. the example/privileged/values.yaml also assumes namespaces spire-server/spire-system in its config, so you'd have to override some of the stuff that references spire-server as well to refer to lockdown instead in .github/tests/lockdown/values.yaml

marcofranssen:
Finally fixed the test. This PR builds on top of #143

.github/tests/run-test.sh
```diff
@@ -0,0 +1,29 @@
+#!/usr/bin/env bash
```
kfox1111, line 1: How does this get run?

kfox1111: A future pr with that sounds like a good idea. :)

.github/tests/production-example/pre-install.sh
```diff
@@ -0,0 +1,6 @@
+#!/usr/bin/env bash
+
+kubectl create namespace spire-system
+kubectl label namespace spire-system pod-security.kubernetes.io/enforce=privileged
+kubectl create namespace spire-server
+kubectl label namespace spire-server pod-security.kubernetes.io/enforce=restricted
```
kfox1111, line 6: Since you are doing an install.sh, this stuff could just be moved there.

kfox1111:
This needs to wait until https://github.com/spiffe/helm-charts/pull/134 to maintain history. I'd also like to see it pass the tests.

kfox1111:
Ok, I think probably just the pre-install.sh stuff merging into install.sh is the last thing to do and this looks good to me.

.github/tests/production-example/pre-install.sh
marcofranssen: Prefer to keep it here, in case future test setup refactors, this makes it very clear.

.github/tests/production-example/post-install.sh
```diff
@@ -0,0 +1,51 @@
...
+\`\`\`
+
+#### Logs
+
+\`\`\`shell
+$(kubectl get pods -o name -n "$1" | while read -r line; do echo logs for "${line}"; kubectl logs -n "$1" "${line}" --all-containers=true --ignore-errors=true; done)
+\`\`\`
+
+EOF
+}
```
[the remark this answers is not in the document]
marcofranssen: That is what I was planning on that other PR to bring this feature everywhere. (The draft one we discussed earlier)

marcofranssen:
> Ok, I think probably just the pre-install.sh stuff merging into install.sh is the last thing to do and this looks good to me.

Prefer to keep it in pre-install, so install.sh is purely used to replace the install command (the way it is used in the pipeline).

kfox1111 (approved):

PR #128, opened 2023-03-15
Add hooks to set failurepolicy to ignore on webhook

faisal-memon:
Adds a set of 3 hooks to address upgrade failure issues. Workflow:

- Set failurePolicy on the webhook to `Ignore` in the template yaml
- Post-install hook sets it to `{{ .Values.controllerManager.validatingWebhookConfiguration.failurePolicy }}`
- Pre-upgrade sets it to `Ignore`
- Post-upgrade sets it back to `{{ .Values.controllerManager.validatingWebhookConfiguration.failurePolicy }}`

I thought this was the most deterministic way to workaround this issue. 

We could potentially have only the pre upgrade hook if we ensure that the clusterspiffeid is always installed/upgraded before the webhook. Right now this is the case as helm determines install order by filesystem order and `controller-manager-cluster-ids.yaml` comes before `controller-manager-webhook.yaml`. Relying on file names seems error prone to me though. What do you think?

fixes https://github.com/spiffe/helm-charts/issues/72

### DIFF
--- a/charts/spire/charts/spire-server/templates/controller-manager-webhook.yaml
+++ b/charts/spire/charts/spire-server/templates/controller-manager-webhook.yaml
@@ -24,7 +24,7 @@ webhooks:
         name: {{ include "spire-controller-manager.fullname" . }}-webhook
         namespace: {{ include "spire-server.namespace" . }}
         path: /validate-spire-spiffe-io-v1alpha1-clusterspiffeid
-    failurePolicy: {{ .Values.controllerManager.validatingWebhookConfiguration.failurePolicy }}
+    failurePolicy: Ignore
     name: vclusterspiffeid.kb.io
     rules:
       - apiGroups: ["spire.spiffe.io"]
--- a/charts/spire/charts/spire-server/templates/controller-manager-webhook.yaml
+++ b/charts/spire/charts/spire-server/templates/controller-manager-webhook.yaml
@@ -24,7 +24,7 @@ webhooks:
         name: {{ include "spire-controller-manager.fullname" . }}-webhook
         namespace: {{ include "spire-server.namespace" . }}
         path: /validate-spire-spiffe-io-v1alpha1-clusterspiffeid
-    failurePolicy: Ignore
+    failurePolicy: Ignore # Actual value to be set by post install/upgrade hooks
     name: vclusterspiffeid.kb.io
     rules:
       - apiGroups: ["spire.spiffe.io"]
--- a/charts/spire/charts/spire-server/templates/controller-manager-webhook.yaml
+++ b/charts/spire/charts/spire-server/templates/controller-manager-webhook.yaml
@@ -10,7 +10,7 @@ webhooks:
         name: {{ include "spire-controller-manager.fullname" . }}-webhook
         namespace: {{ include "spire-server.namespace" . }}
         path: /validate-spire-spiffe-io-v1alpha1-clusterfederatedtrustdomain
-    failurePolicy: {{ .Values.controllerManager.validatingWebhookConfiguration.failurePolicy }}
+    failurePolicy: Ignore # Actual value to be set by post install/upgrade hooks
     name: vclusterfederatedtrustdomain.kb.io
     rules:
       - apiGroups: ["spire.spiffe.io"]
--- a/charts/spire/charts/spire-server/templates/post-install-hook.yaml
+++ b/charts/spire/charts/spire-server/templates/post-install-hook.yaml
@@ -59,7 +59,7 @@ spec:
       serviceAccountName: {{ include "spire-server.serviceAccountName" . }}-post-install
       containers:
       - name: post-install-job
-        image: bitnami/kubectl
+        image: {{ template "spire-server.image" (dict "appVersion" $.Chart.AppVersion "image" .Values.controllerManager.validatingWebhookConfiguration.upgradeHook.image) }}
         command:
         - /bin/sh
         - -c
--- a/charts/spire/charts/spire-server/templates/post-install-hook.yaml
+++ b/charts/spire/charts/spire-server/templates/post-install-hook.yaml
@@ -68,6 +68,18 @@ spec:
         - /bin/sh
         - -c
         - |-
-          kubectl patch validatingwebhookconfiguration {{ include "spire-controller-manager.fullname" . }}-webhook --type='strategic' -p '{"webhooks":[{"name":"vclusterspiffeid.kb.io","failurePolicy":"{{ .Values.controllerManager.validatingWebhookConfiguration.failurePolicy }}"}]}'
+          kubectl patch validatingwebhookconfiguration {{ include "spire-controller-manager.fullname" . }}-webhook --type='strategic' -p '
+          {
+            "webhooks":[
+              {
+                "name":"vclusterspiffeid.kb.io",
+                "failurePolicy":"{{ .Values.controllerManager.validatingWebhookConfiguration.failurePolicy }}"
+              },
+              {
+                "name":"vclusterfederatedtrustdomain.kb.io",
+                "failurePolicy":"{{ .Values.controllerManager.validatingWebhookConfiguration.failurePolicy }}"
+              }
+            ]
+          }'
 {{- end }}
 {{- end }}
--- a/charts/spire/charts/spire-server/templates/post-install-hook.yaml
+++ b/charts/spire/charts/spire-server/templates/post-install-hook.yaml
@@ -64,6 +64,6 @@ spec:
         - /bin/sh
         - -c
         - |-
-          kubectl patch validatingwebhookconfiguration {{ include "spire-controller-manager.fullname" . }}-webhook --type='strategic' -p "{\"webhooks\":[{\"name\":\"vclusterspiffeid.kb.io\",\"failurePolicy\":\"{{ .Values.controllerManager.validatingWebhookConfiguration.failurePolicy }}\"}]}"
+          kubectl patch validatingwebhookconfiguration {{ include "spire-controller-manager.fullname" . }}-webhook --type='strategic' -p '{"webhooks":[{"name":"vclusterspiffeid.kb.io","failurePolicy":"{{ .Values.controllerManager.validatingWebhookConfiguration.failurePolicy }}"}]}'
 {{- end }}
 {{- end }}
--- a/charts/spire/charts/spire-server/templates/post-install-hook.yaml
+++ b/charts/spire/charts/spire-server/templates/post-install-hook.yaml
@@ -21,6 +21,7 @@ metadata:
 rules:
   - apiGroups: ["admissionregistration.k8s.io"]
     resources: ["validatingwebhookconfigurations"]
+    resourceNames: [{{ printf "%s-webhook"  (include "spire-controller-manager.fullname" .) | quote }}]
     verbs: ["get", "patch"]
 ---
 kind: ClusterRoleBinding
--- a/charts/spire/charts/spire-server/templates/post-install-hook.yaml
+++ b/charts/spire/charts/spire-server/templates/post-install-hook.yaml
@@ -1,0 +1,66 @@
+{{- if eq (.Values.controllerManager.enabled | toString) "true" }}
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  name: {{ include "spire-server.serviceAccountName" . }}-post-install
+  namespace: {{ include "spire-server.namespace" . }}
+  labels:
+    {{- include "spire-server.labels" . | nindent 4 }}
+  annotations:
+    "helm.sh/hook": post-install
+    "helm.sh/hook-delete-policy": hook-succeeded, hook-failed
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: {{ include "spire-server.fullname" . }}-post-install
+  annotations:
+    "helm.sh/hook": post-install
+    "helm.sh/hook-delete-policy": hook-succeeded, hook-failed
+rules:
+  - apiGroups: ["admissionregistration.k8s.io"]
+    resources: ["validatingwebhookconfigurations"]
+    verbs: ["get", "list", "patch"]
+---
+kind: ClusterRoleBinding
+apiVersion: rbac.authorization.k8s.io/v1
+metadata:
+  name: {{ include "spire-server.fullname" . }}-post-install
+  annotations:
+    "helm.sh/hook": post-install
+    "helm.sh/hook-delete-policy": hook-succeeded, hook-failed
+subjects:
+  - kind: ServiceAccount
+    name: {{ include "spire-server.serviceAccountName" . }}-post-install
+    namespace: {{ include "spire-server.namespace" . }}
+roleRef:
+  kind: ClusterRole
+  name: {{ include "spire-server.fullname" . }}-post-install
+  apiGroup: rbac.authorization.k8s.io
+---
+apiVersion: batch/v1
+kind: Job
+metadata:
+  name: {{ include "spire-server.fullname" . }}-post-install
+  namespace: {{ include "spire-server.namespace" . }}
+  labels:
+    {{- include "spire-server.labels" . | nindent 4 }}
+  annotations:
+    "helm.sh/hook": post-install
+    "helm.sh/hook-delete-policy": hook-succeeded, hook-failed
+spec:
+  template:
+    metadata:
+      name: {{ include "spire-server.fullname" . }}-post-install
+    spec:
+      restartPolicy: Never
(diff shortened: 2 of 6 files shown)


charts/spire/charts/spire-server/templates/pre-upgrade-hook.yaml
```diff
@@ -0,0 +1,66 @@
+{{- if eq (.Values.controllerManager.enabled | toString) "true" }}
```
kfox1111, line 1: same

charts/spire/charts/spire-server/templates/post-install-hook.yaml
```diff
@@ -0,0 +1,66 @@
...
+      serviceAccountName: {{ include "spire-server.serviceAccountName" . }}-post-install
+      containers:
+      - name: post-install-job
+        image: bitnami/kubectl
+        command:
+        - /bin/sh
+        - -c
+        - |
+          kubectl patch validatingwebhookconfiguration {{ include "spire-controller-manager.fullname" . }}-webhook --type='strategic' -p "{\"webhooks\":[{\"name\":\"vclusterspiffeid.kb.io\",\"failurePolicy\":\"{{ .Values.controllerManager.validatingWebhookConfiguration.failurePolicy }}\"}]}"
+{{- end }}
```
kfox1111, line 66: Needs security context bits to run in a locked down namespace.

charts/spire/charts/spire-server/templates/post-upgrade-hook.yaml
```diff
@@ -0,0 +1,66 @@
+{{- if eq (.Values.controllerManager.enabled | toString) "true" }}
```
kfox1111, line 1: Each of the other hook files has the same issues.

charts/spire/charts/spire-server/templates/post-install-hook.yaml
```diff
@@ -0,0 +1,66 @@
...
+      restartPolicy: Never
+      serviceAccountName: {{ include "spire-server.serviceAccountName" . }}-post-install
+      containers:
+      - name: post-install-job
+        image: bitnami/kubectl
+        command:
+        - /bin/sh
+        - -c
+        - |
+          kubectl patch validatingwebhookconfiguration {{ include "spire-controller-manager.fullname" . }}-webhook --type='strategic' -p "{\"webhooks\":[{\"name\":\"vclusterspiffeid.kb.io\",\"failurePolicy\":\"{{ .Values.controllerManager.validatingWebhookConfiguration.failurePolicy }}\"}]}"
```
kfox1111, line 65: There are two webhooks with fail on them.... This is only hitting the one out of the box, but will fail upgrade/install on the other if anyone adds a cr of that type.

kfox1111 (dismissed):

charts/spire/charts/spire-server/templates/post-install-hook.yaml
```diff
@@ -0,0 +1,66 @@
+{{- if eq (.Values.controllerManager.enabled | toString) "true" }}
```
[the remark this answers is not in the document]
marcofranssen: why? If the workaround works for now, then that is the fix until the controller manager is fixed.

I don't like to have tons and tons of options.

It either works or it doesn't. Havong a value for every single bit is a bit overkill I think.

charts/spire/charts/spire-server/templates/post-install-hook.yaml
```diff
@@ -0,0 +1,66 @@
...
+spec:
+  template:
+    metadata:
+      name: {{ include "spire-server.fullname" . }}-post-install
+    spec:
+      restartPolicy: Never
+      serviceAccountName: {{ include "spire-server.serviceAccountName" . }}-post-install
+      containers:
+      - name: post-install-job
+        image: bitnami/kubectl
```
marcofranssen, line 60: Please us the macro function we have in place for setting the images.

charts/spire/charts/spire-server/templates/post-install-hook.yaml
marcofranssen, line 65:
```suggestion
          kubectl patch validatingwebhookconfiguration {{ include "spire-controller-manager.fullname" . }}-webhook --type='strategic' -p '{"webhooks":[{"name":"vclusterspiffeid.kb.io","failurePolicy":"{{ .Values.controllerManager.validatingWebhookConfiguration.failurePolicy }}"}]}'
```

charts/spire/charts/spire-server/templates/post-upgrade-hook.yaml
marcofranssen: I really wouldn't make this configurable, this is just a necessary fix until the the controller-manager-webhook fixes it.

charts/spire/charts/spire-server/templates/pre-upgrade-hook.yaml
```diff
@@ -0,0 +1,66 @@
...
+    spec:
+      restartPolicy: Never
+      serviceAccountName: {{ include "spire-server.serviceAccountName" . }}-pre-upgrade
+      containers:
+      - name: post-install-job
+        image: bitnami/kubectl
+        command:
+        - /bin/sh
+        - -c
+        - |
```
marcofranssen, line 64:
```suggestion
        - |-
```

charts/spire/charts/spire-server/templates/post-upgrade-hook.yaml
```diff
@@ -0,0 +1,66 @@
...
+    spec:
+      restartPolicy: Never
+      serviceAccountName: {{ include "spire-server.serviceAccountName" . }}-post-upgrade 
+      containers:
+      - name: post-upgrade-job
+        image: bitnami/kubectl
+        command:
+        - /bin/sh
+        - -c
+        - |
```
marcofranssen, line 64:
```suggestion
        - |-
```

This changes the yaml to a single line string despite being spanned over multiple lines which is what you want.

marcofranssen (changes requested):
Few small things, don't think we need to make these helm hooks configurable. Either the fix works or it doesn't, no need to bring this to the user facing API IMHO.

charts/spire/charts/spire-server/templates/post-install-hook.yaml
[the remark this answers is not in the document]
kfox1111: As previously discussed, the workaround always runs, whether the chart runs or not. It can cause some issues with ci/cd systems. I do not want it running every time I do a helm upgrade. It needs an option to disable it. helm has a flag to skip hooks, but since the resource itself is forcing a change, then the flag doesnt work properly either. So a value is needed.

charts/spire/charts/spire-server/templates/post-upgrade-hook.yaml
kfox1111: If we're willing to compromise on a workaround to keep it simple because its temporary, we can also just always set the fail policy to ignore until the fix is out and that's much much simpler then all of these hooks.

charts/spire/charts/spire-server/templates/post-install-hook.yaml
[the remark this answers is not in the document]
faisal-memon: I wrapped the hook in `{{- if eq .Values.controllerManager.validatingWebhookConfiguration.failurePolicy "Fail" }}` so they won't run if the FailurePolicy is Ignore. That should cover your use case without adding more flags.

charts/spire/charts/spire-server/templates/post-install-hook.yaml
```diff
@@ -0,0 +1,66 @@
...
+kind: ClusterRole
+metadata:
+  name: {{ include "spire-server.fullname" . }}-post-install
+  annotations:
+    "helm.sh/hook": post-install
+    "helm.sh/hook-delete-policy": hook-succeeded, hook-failed
+rules:
+  - apiGroups: ["admissionregistration.k8s.io"]
+    resources: ["validatingwebhookconfigurations"]
+    verbs: ["get", "list", "patch"]
```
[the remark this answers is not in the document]
faisal-memon: Removed the `list` and added the `resourceNames`.

charts/spire/charts/spire-server/templates/post-install-hook.yaml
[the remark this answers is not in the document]
marcofranssen: Especially in CD I need to run it. Without this fix the upgrade just fails. Still don't get the point on having this configurable? I guess I need to take some time testing to understand this part.

charts/spire/charts/spire-server/templates/post-upgrade-hook.yaml
marcofranssen: IMHO this is the real fix. At least after the upgrade there who had the behavior I want.

kfox1111: Patching files with hooks that helm manages while the system is running is a hack. We're just debating which hack is best at this point. I'm glad the behavior works for you. I don't think the behavior works for me. So I just want the option to opt out.

faisal-memon: This is covered with an additional configurable.

kfox1111: Not sure I follow. Do you mean you think it can be done as is with one of the configurables, or that you will add the additional configurable?

faisal-memon: Sorry, typo. *without*. If you set fail policy to ignore the hooks wont run

charts/spire/charts/spire-server/templates/post-install-hook.yaml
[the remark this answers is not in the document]
faisal-memon: Doesnt work without the get. I think under the hood kubctl does a read modify write sequence.

charts/spire/charts/spire-server/templates/post-install-hook.yaml
faisal-memon: i did that, but that does expose additional config options in values.yaml. What do you think of that? https://github.com/spiffe/helm-charts/pull/128/commits/448d8a6b7f46f862a5a031d87d3ee6ba71c2e3c3#diff-2970cff31c683ad68e57af57ddbb9acf98765a7898efdeff84c8532c260e1dcbR185-R191

charts/spire/charts/spire-server/templates/post-install-hook.yaml
faisal-memon: I only see the controller manager one. whats the other one?

charts/spire/charts/spire-server/templates/post-install-hook.yaml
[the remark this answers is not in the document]
kfox1111: @faisal-memon ah, thats cleaver. +1.

kfox1111:
This is looking pretty good now. Just need the securityContext bits and then I think it will work.

kfox1111:
See the two securityContext bits here as an example:
https://github.com/spiffe/helm-charts/blob/main/charts/spire/charts/spiffe-oidc-discovery-provider/templates/tests/test-connection.yaml#L11-L19

charts/spire/charts/spire-server/templates/post-install-hook.yaml
kfox1111: The webhook for clusterfederatedtrustdomains. https://github.com/spiffe/helm-charts/blob/main/charts/spire/charts/spire-server/templates/controller-manager-webhook.yaml#L13

I'm thinking maybe the upgrade will also fail if the user has one specified?

charts/spire/charts/spire-server/templates/post-install-hook.yaml
kfox1111: lgtm

charts/spire/charts/spire-server/templates/controller-manager-webhook.yaml
```diff
@@ -10,7 +10,7 @@ webhooks:
         name: {{ include "spire-controller-manager.fullname" . }}-webhook
         namespace: {{ include "spire-server.namespace" . }}
         path: /validate-spire-spiffe-io-v1alpha1-clusterfederatedtrustdomain
-    failurePolicy: {{ .Values.controllerManager.validatingWebhookConfiguration.failurePolicy }}
+    failurePolicy: Ignore # Actual value to be set by post install/upgrade hooks
```
kfox1111, line 13: Will need to set  the other webhook to ignore in this file too.

charts/spire/charts/spire-server/templates/post-install-hook.yaml
faisal-memon: oh good catch. Made this change: https://github.com/spiffe/helm-charts/pull/128/commits/7453f36cbd7334c9fa1a3f1ac58d07462bc366f2

kfox1111:
And run ./helm-docs.sh

charts/spire/charts/spire-server/templates/controller-manager-webhook.yaml
faisal-memon: Should be covering both now. Did I miss something?

kfox1111: Oh. I think I had an older version of the file. This is looking ok to me now.

charts/spire/charts/spire-server/templates/post-install-hook.yaml
```diff
@@ -0,0 +1,85 @@
...
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  name: {{ include "spire-server.serviceAccountName" . }}-post-install
+  namespace: {{ include "spire-server.namespace" . }}
+  labels:
+    {{- include "spire-server.labels" . | nindent 4 }}
+  annotations:
+    "helm.sh/hook": post-install
+    "helm.sh/hook-delete-policy": hook-succeeded, hook-failed
```
kfox1111, line 12: hmm.. what happens on hook-failure? if it deletes the job, no one might notice it has failed and the webhook is still Ignore? or does the helm upgrade fail and its ok cause the user will notice?

kfox1111:
I was thinking about this some more. For those that don't want to use the hook, we can simply do any upgrade twice:
```
# Do the actual upgrade with ignore set
helm upgrade --install spire chart/spire -f myvalues.yaml  \
 --set spire-server controllerManager.validatingWebhookConfiguration.failurePolicy=Ignore

# Set the fail flag back.
helm upgrade --install spire chart/spire -f myvalues.yaml \ 
--set spire-server controllerManager.validatingWebhookConfiguration.failurePolicy=Fail
```

And it basically will just switch the flag back and forth without other changes if nothing else changed.

But, with the flag doing double duty this doesn't work. So, could we get it added as a separate flag to disable the jobs?

marcofranssen:
That won't work with argocd. Those kind of tools will need a single install that simply succeeds.

What is the exact problem with the hooks?

faisal-memon:
@kfox1111 You can use the helm `--no-hooks` flag to disable them.

faisal-memon:
I did some research today. Seem like this is the [install order](https://github.com/helm/helm/blob/12f1bc0acdeb675a8c50a78462ed3917fb7b2e37/pkg/releaseutil/kind_sorter.go#L31-L67). Custom resources are [installed last](https://github.com/helm/helm/blob/12f1bc0acdeb675a8c50a78462ed3917fb7b2e37/pkg/releaseutil/kind_sorter.go#L149).

charts/spire/charts/spire-server/values.yaml
```diff
@@ -182,6 +182,13 @@ controllerManager:
 
   validatingWebhookConfiguration:
     failurePolicy: Fail
+    upgradeHook:
+      image:
+        registry: "bitnami"
+        repository: kubectl
+        pullPolicy: IfNotPresent
+        # Overrides the image tag whose default is the chart appVersion.
+        version: "latest"
```
marcofranssen, line 191: https://www.chainguard.dev/unchained/chainguard-image-now-available-for-kubectl
https://github.com/chainguard-images/images/tree/main/images/kubectl

```suggestion
        registry: cgr.dev
        repository: chainguard/kubectl
        pullPolicy: IfNotPresent
        version: latest
```

marcofranssen (approved):
LGTM 🚀 

Tested on my cluster and it successfully upgraded the chart.

kfox1111:
Still really dislike that this got forced through.... is a single flag to disable all this stuff really that bad?

faisal-memon:
> Still really dislike that this got forced through.... is a single flag to disable all this stuff really that bad?

Did you try the `--no-hooks` suggestion? That should address your concern.

PR #130, opened 2023-03-16
Enhance the production example

marcofranssen:
To be double checked with latest release. https://github.com/spiffe/helm-charts/releases/tag/spire-0.5.0

### DIFF
--- a/examples/production/values.yaml
+++ b/examples/production/values.yaml
@@ -3,24 +3,6 @@ global:
     prometheus:
       enabled: true
 
-spiffe-oidc-discovery-provider:
-  enabled: true
-  insecureScheme:
-    enabled: true
-  podSecurityContext:
-    runAsUser: 1000
-    runAsGroup: 1000
-    fsGroup: 1000
-    fsGroupChangePolicy: OnRootMismatch
-  securityContext:
-    allowPrivilegeEscalation: false
-    runAsNonRoot: true
-    readOnlyRootFilesystem: true
-    capabilities:
-      drop: [ALL]
-    seccompProfile:
-      type: RuntimeDefault
-
 spire-server:
   nodeAttestor:
     k8sPsat:
@@ -41,6 +23,12 @@ spire-server:
       drop: [ALL]
     seccompProfile:
       type: RuntimeDefault
+
+  dataStorage:
+    enabled: true
+
+  logLevel: info
+
   controllerManager:
     securityContext:
       allowPrivilegeEscalation: false
@@ -50,6 +38,11 @@ spire-server:
         drop: [ALL]
       seccompProfile:
         type: RuntimeDefault
+
+    ignoreNamespaces:
+      - kube-system
+      - kube-public
+      - local-path-storage
 
 spiffe-csi-driver:
   enabled: true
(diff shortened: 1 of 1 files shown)


examples/production/values.yaml
```diff
@@ -37,7 +21,43 @@ spire-server:
       drop: [ALL]
     seccompProfile:
       type: RuntimeDefault
+
+  topologySpreadConstraints:
+    - maxSkew: 1
+      topologyKey: topology.kubernetes.io/zone
```
kfox1111, line 27: I don't think a lot of clusters have zone defined, and DoNotSchedule together might break things?

examples/production/values.yaml
```diff
@@ -37,7 +21,43 @@ spire-server:
...
+    enabled: true
+    size: 10Gi
+    accessMode: ReadWriteOnce
+    storageClass: gp3-encrypted
+
+  jwtIssuer: &jwtIssuer oidc-spire.eu-west-1-marco-test.my-domain.com
+  ca_subject:
+    country: "NL"
+    organization: "Marco Home Office"
+    common_name: spire.eu-west-1-marco-test.my-domain.com
```
kfox1111, line 52: Should probably leave the jwtIssuer, ca_subject, clusterName and trustDomain alone?

examples/production/values.yaml
```diff
@@ -47,10 +67,39 @@ spire-server:
...
+
+  resources:
+    requests:
+      cpu: 100m
+      memory: 32Mi
+    limits:
+      cpu: 200m
+      memory: 128Mi
+
+  nodeDriverRegistrar:
```
kfox1111, line 95: This chunk should be removed. we dont have it.

examples/production/values.yaml
```diff
@@ -60,3 +109,69 @@ spire-agent:
...
+  resources:
+    requests:
+      cpu: 100m
+      memory: 64Mi
+    limits:
+      cpu: 500m
+      memory: 512Mi
+
+  clusterName: *clusterName
+  trustDomain: *trustDomain
```
kfox1111, line 124: Shouldn't have clustername/domain here.

examples/production/values.yaml
```diff
@@ -60,3 +109,69 @@ spire-agent:
...
+
+  resources:
+    requests:
+      cpu: 100m
+      memory: 64Mi
+    limits:
+      cpu: 200m
+      memory: 256Mi
+
+  config:
```
kfox1111, line 168: From here down should probably drop out too.

examples/production/values.yaml
marcofranssen: We do have that see the daemonset. There is a second container.

kfox1111: Doh. I misread that and was thinking about the other k8s sync thing. Sorry.

examples/production/values.yaml
```diff
@@ -41,7 +23,17 @@ spire-server:
       drop: [ALL]
     seccompProfile:
       type: RuntimeDefault
+
+  dataStorage:
+    enabled: true
+    size: 10Gi
```
kfox1111, line 29: The db only seems to be used for a very small number of entries, even on a big cluster. 10Gi seems overkill.
1Gi? 
I'd recommend even smaller, but I've seen provisioning break on volumes < 1Gi.

examples/production/values.yaml
```diff
@@ -41,7 +23,17 @@ spire-server:
...
+    accessMode: ReadWriteOnce
+
+  logLevel: info
+
   controllerManager:
+    enabled: true
```
kfox1111, line 35: Its already defaulted to enabled.

kfox1111 (changes requested):
Mostly looking good.

examples/production/values.yaml
```diff
@@ -41,7 +23,17 @@ spire-server:
       drop: [ALL]
     seccompProfile:
       type: RuntimeDefault
+
+  dataStorage:
```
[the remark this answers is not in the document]
marcofranssen: yup requires changes on the chart as well

kfox1111 (approved):
lgtm

PR #132, opened 2023-03-17
Allow advanced users to bring their own configmaps

kfox1111:
So that we don't have to support every single config option available, add the option for the advanced end user to skip the chart managing configmaps and allow the user to provide their own.


kfox1111:
Closing in favor of json config.

PR #133, opened 2023-03-17
Remove k8s 1.21 from test matrix + small syntax error fix

marcofranssen:
- Remove k8s 1.21 from test matrix
- Fix syntax
- Improve readability test step

Officialy according to readme we support last three k8s versions.
Testing last 5 versions should be sufficient.


kfox1111:
Its unlikely that we'll introduce a bug that makes this not work as far back as 1.21. but its kind of nice to test it just to be sure. If we're only supporting 3, maybe we do those 3 and 1.21 just out of curiosity if it works that far back. if it works on 1.21 its highly likely it will work on 1.22 as well. If we're going to remove one, should we remove 1.22 instead?

The rest looks good to me.

faisal-memon (approved):

PR #134, opened 2023-03-17
Fork the lockdown test to two tests as it is doing the work of 2

kfox1111:
This patch makes a second copy of the existing lockdown test
because it is doing double duty. In follow on patches we will
make each test do one thing only.

### DIFF
--- a/.github/tests/lockdown2/.env
+++ b/.github/tests/lockdown2/.env
@@ -1,0 +1,1 @@
+export EXTRA_HELM_ARGS="--values=examples/production/values.yaml"
--- a/.github/tests/lockdown2/post-install.sh
+++ b/.github/tests/lockdown2/post-install.sh
@@ -1,0 +1,48 @@
+#!/usr/bin/env bash
+
+set -x
+
+SCRIPT=$(readlink -f "$0")
+SCRIPTPATH=$(dirname "$SCRIPT")
+
+k_wait=(kubectl wait --for condition=available --timeout 30s --namespace)
+k_rollout_status=(kubectl rollout status --watch --timeout 30s --namespace)
+
+cat <<EOF >>"$GITHUB_STEP_SUMMARY"
+### spire
+| workload | Status |
+| -------- | ------ |
+| spire-server | $("${k_rollout_status[@]}" spire-server statefulset spire-server) |
+| spire-spiffe-csi-driver | $("${k_rollout_status[@]}" spire-system daemonset spire-spiffe-csi-driver) |
+| spire-agent | $("${k_rollout_status[@]}" spire-system daemonset spire-agent) |
+| spire-spiffe-oidc-discovery-provider | $("${k_wait[@]}" spire-server deployments.apps spire-spiffe-oidc-discovery-provider) |
+EOF
+
+if [ $1 -ne 0 ]; then
+  echo
+  echo '```'
+  echo '==> Events of namespace spire-server'
+  echo '........................................................................................................................'
+  echo '>>> kubectl --request-timeout=30s get events --output wide --namespace spire-server'
+  kubectl --request-timeout=30s get events --output wide --namespace spire-server
+  echo '........................................................................................................................'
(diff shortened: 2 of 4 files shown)


.github/tests/namespace-override/pre-install.sh
```diff
@@ -0,0 +1,5 @@
+#!/bin/bash
+kubectl create namespace "spire-system"
+kubectl label namespace "spire-system" pod-security.kubernetes.io/enforce=privileged
+kubectl create namespace "spire-server"
+kubectl label namespace "spire-server" pod-security.kubernetes.io/enforce=restricted
```
marcofranssen, line 5:
```suggestion
kubectl create namespace spire-system
kubectl label namespace spire-system pod-security.kubernetes.io/enforce=privileged
kubectl create namespace spire-server
kubectl label namespace spire-server pod-security.kubernetes.io/enforce=restricted
```

.github/tests/namespace-override/post-install.sh
```diff
@@ -0,0 +1,48 @@
...
+k_rollout_status=(kubectl rollout status --watch --timeout 30s --namespace)
+
+cat <<EOF >>"$GITHUB_STEP_SUMMARY"
+### spire
+| workload | Status |
+| -------- | ------ |
+| spire-server | $("${k_rollout_status[@]}" spire-server statefulset spire-server) |
+| spire-spiffe-csi-driver | $("${k_rollout_status[@]}" spire-system daemonset spire-spiffe-csi-driver) |
+| spire-agent | $("${k_rollout_status[@]}" spire-system daemonset spire-agent) |
+| spire-spiffe-oidc-discovery-provider | $("${k_wait[@]}" spire-server deployments.apps spire-spiffe-oidc-discovery-provider) |
```
marcofranssen, line 18:
```suggestion
| spire-server | $("${k_rollout_status[@]}" spire-server statefulset spire-server) |
| spire-spiffe-oidc-discovery-provider | $("${k_wait[@]}" spire-server deployments.apps spire-spiffe-oidc-discovery-provider) |
| spire-spiffe-csi-driver | $("${k_rollout_status[@]}" spire-system daemonset spire-spiffe-csi-driver) |
| spire-agent | $("${k_rollout_status[@]}" spire-system daemonset spire-agent) |
```

marcofranssen (changes requested):
Could you add a temporary commit to run the tests? Also few nit-picky suggestions in there.

.github/tests/namespace-override/pre-install.sh
kfox1111: Done

kfox1111:
This is ready to go, but needs the version bump reverted. It was put in place to ensure all the changes work.

.github/tests/namespace-override/pre-install.sh
```diff
@@ -0,0 +1,5 @@
+#!/bin/bash
```
marcofranssen, line 1:
```suggestion
#!/usr/bin/env bash

```

.github/tests/namespace-override/.env
```diff
@@ -0,0 +1 @@
+export EXTRA_HELM_ARGS="--values=examples/production/values.yaml"
```
marcofranssen, line 1: This should not be part of this test. I would expect just the namespace override.

marcofranssen (dismissed):
Can we rename this PR to just

`Replace lockdown test with namespace-override test`?

#127 replaces the lockdown test with just testing the production example (which is already locked down).

I even debate the value of testing these 2 namespace overrides. Feels a bit like this PR is obsolete when we merge #127

kfox1111:
Trying to maintain history. the existing lockdown test is doing 2 things. testing production values, and namespace-override testing. To follow the established principal of one pr doing just one thing, we need to fork the test into the two so its obvious where the test came from. We then can continue to enhance both test separately in separate pr's.

faisal-memon (approved):

PR #136, opened 2023-03-17
ingress-nginx production tests and spiffe-oidc-discovery-provider example

kfox1111:
This pr adds a production addon values file with best practices for using ingress-nginx with spire-oidc-discovery-provider and tests all existing ingresses to ensure they work properly.


.github/tests/production-ingress-nginx/install.sh
```diff
@@ -0,0 +1,29 @@
+#!/bin/bash -e
```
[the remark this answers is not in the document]
kfox1111: I can pull out the set, but its much more common for bash to exist at /bin then it is for /usr/bin/env to exist. So this makes it less portable. Do we really want to do this?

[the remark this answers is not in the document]
kfox1111: I did some looking and every place I could find had both env and bash.... so I may be showing my age. It used to be true bash was more common then env. Modern distros seem to have both.

kfox1111:
Rebased

.github/tests/production-ingress-nginx/post-install.sh
```diff
@@ -0,0 +1,39 @@
+#!/usr/bin/env bash
+
+set -x
+
+SCRIPT=$(readlink -f "$0")
+SCRIPTPATH=$(dirname "$SCRIPT")
+
+k_wait=(kubectl wait --for condition=available --timeout 30s --namespace)
+k_rollout_status=(kubectl rollout status --watch --timeout 30s --namespace)
```
marcofranssen, line 9: Maybe do the same fix as in your recent PR. To align the style now we are touching it newly

.github/tests/production-ingress-nginx/post-install.sh
```diff
@@ -0,0 +1,39 @@
...
+
+cat <<EOF >>"$GITHUB_STEP_SUMMARY"
+### spire
+| workload | Status |
+| -------- | ------ |
+| spire-server | $("${k_rollout_status[@]}" spire-server statefulset spire-server) |
+| spire-spiffe-oidc-discovery-provider | $("${k_wait[@]}" spire-server deployments.apps spire-spiffe-oidc-discovery-provider) |
+| spire-spiffe-csi-driver | $("${k_rollout_status[@]}" spire-system daemonset spire-spiffe-csi-driver) |
+| spire-agent | $("${k_rollout_status[@]}" spire-system daemonset spire-agent) |
```
marcofranssen, line 18:
```suggestion
| spire-server | <pre>$("${k_rollout_status[@]}" spire-server statefulset spire-server)</pre> |
| spire-spiffe-oidc-discovery-provider | <pre>$("${k_wait[@]}" spire-server deployments.apps spire-spiffe-oidc-discovery-provider)</pre> |
| spire-spiffe-csi-driver | <pre>$("${k_rollout_status[@]}" spire-system daemonset spire-spiffe-csi-driver)</pre> |
| spire-agent | <pre>$("${k_rollout_status[@]}" spire-system daemonset spire-agent)</pre> |
```

.github/tests/production-ingress-nginx/testcert.yaml
```diff
@@ -0,0 +1,44 @@
...
+  ca:
+    secretName: root-secret
+---
+apiVersion: cert-manager.io/v1
+kind: Certificate
+metadata:
+  name: oidc
+spec:
+  dnsNames:
+  - oidc-discovery.example.org
```
marcofranssen, line 39:
```suggestion
    - oidc-discovery.example.org
```

charts/spire/charts/spiffe-oidc-discovery-provider/values.yaml
```diff
@@ -77,8 +77,8 @@ insecureScheme:
 config:
   logLevel: info
   domains:
-    - localhost
     - oidc-discovery.example.org
+    - localhost
```
marcofranssen, line 81: Why was this swap of domains required?

examples/production/README.md
```diff
@@ -8,7 +8,14 @@ kubectl label namespace "spire-system" pod-security.kubernetes.io/enforce=privil
 kubectl create namespace "spire-server"
 kubectl label namespace "spire-server" pod-security.kubernetes.io/enforce=restricted
 
+```
```
marcofranssen, line 11:
```suggestion
```shell
```

examples/production/README.md
```diff
@@ -8,7 +8,14 @@ kubectl label namespace "spire-system" pod-security.kubernetes.io/enforce=privil
...
 helm upgrade --install --namespace spire-server spire charts/spire -f values.yaml
 ```
 
+If your using ingress-nginx and want to expose spire outside the cluster, add the
+following to the end of the helm upgrade example:
+```
```
marcofranssen, line 17:
```suggestion

```shell
```

charts/spire/charts/spiffe-oidc-discovery-provider/values.yaml
kfox1111: The test is only testing the first entry in the list... localhost wont resolve to anything useful within the test pod cause it has its own 127.0.0.1 separate from the discovery pod. localhost being in the list is somewhat useful for testing the helm chart within minikube using port-forwarding.

charts/spire/charts/spiffe-oidc-discovery-provider/templates/tests/test-connection.yaml
```diff
@@ -30,11 +30,36 @@ spec:
...
+        {{- toYaml .Values.tests.hostAliases | nindent 8 }}
+      {{- end }}
       securityContext:
         {{- toYaml .Values.securityContext | nindent 8 }}
     {{- end }}
+  {{- if .Values.tests.tls.customCA }}
+  volumes:
+  - name: custom-ca
+    secret:
+      secretName: {{ .Values.tests.tls.customCA }}
```
marcofranssen, line 63:
```suggestion
    - name: custom-ca
      secret:
        secretName: {{ .Values.tests.tls.customCA }}
```

charts/spire/charts/spiffe-oidc-discovery-provider/templates/tests/test-connection.yaml
```diff
@@ -30,11 +30,36 @@ spec:
...
+      command: ['curl']
+      {{- if .Values.tests.tls.customCA }}
+      args: ['--cacert', '/ca/ca.crt', 'https://{{ index .Values.config.domains 0 }}/.well-known/openid-configuration']
+      {{- else }}
+      args: ['https://{{ index .Values.config.domains 0 }}/.well-known/openid-configuration']
+      {{- end }}
+      {{- if .Values.tests.tls.customCA }}
+      volumeMounts:
+      - name: custom-ca
+        mountPoint: /ca
```
marcofranssen, line 45:
```suggestion
        - name: custom-ca
          mountPoint: /ca
```

charts/spire/charts/spiffe-oidc-discovery-provider/values.yaml
```diff
@@ -157,3 +157,8 @@ ingress:
   #  - secretName: chart-example-tls
   #    hosts:
   #      - oidc-discovery.example.org
+
+tests:
+  hostAliases: []
+  tls:
+    customCA: ""
```
marcofranssen, line 164: Doubt a bit of exposing the test things on the public facing API.

examples/production/values-export-ingress-nginx.yaml
```diff
@@ -0,0 +1,27 @@
+spiffe-oidc-discovery-provider:
+  tests:
+    tls:
+      enabled: true
```
marcofranssen, line 4: Don't like to expose the testing logic on public API.

Is there a way around this not to do this?

marcofranssen (changes requested):
2 nitpicks on indents, to use same indent style as the rest of the chart.

Don't like the testing logic being exposed in public API (values.yaml).

examples/production/values-export-ingress-nginx.yaml
kfox1111: The user may have setup things such that tls testing will work (proper certificate (acme) or custom ca loaded in properly). Or a lot of people won't bother and the test will fail without proper config. Even some of our tests don't test it fully. This is the first test that actually does.

charts/spire/charts/spiffe-oidc-discovery-provider/values.yaml
kfox1111: If the user is testing with their own CA (a lot of companies run their own private ones), this is the only way to ensure it works.

examples/production/values-export-ingress-nginx.yaml
marcofranssen: Should we bother testing the working of TLS? can't we just ignore TLS verification. In other words how much outside of our own scope are we going to test.

kfox1111: How do we know our recommendations actually work and we dont break them with PR's? Testing is critical to support IMO. We're going to be recommending some things in the spire-server ingress that will need proper testing too.

I'm good skipping tests of spire itself, as the spire team should be testing those. But things added specifically by the helm chart, that isn't part of spire should probably be tested to ensure it all works together.

Also, the end user may want to test in their QA environment that the particular combination of settings / their cluster all works together to successfully produce a secure endpoint.

faisal-memon:
See https://github.com/spiffe/helm-charts/pull/68#issuecomment-1547034080

I dont think we should be deploying ingress-nginx. Thats best left to the [official ingress-nginx chart](https://artifacthub.io/packages/helm/ingress-nginx/ingress-nginx). Suggest to close this one off.

kfox1111:
Going to rework this a bit now that the other ingress objects merged.

edwbuck:
@kfox1111 Is the rework complete?  This one is getting a bit old to remain in PR state.  If the core of it came across in other items, great! (but please leave a comment confirming that here).  If not, I would like an update so we can close this PR with the desired changes, or properly abandon it.

kfox1111:
Its ready and awaiting review.

marcofranssen (dismissed):

faisal-memon (approved):

kfox1111:
Its now incompatible with the test framework that was rewritten. This pr needs work to get it into the new style.

examples/production/values-export-ingress-nginx.yaml
kfox1111: k8s has a feature in development for that, https://github.com/kubernetes/enhancements/issues/3257 but is not released yet so we can't rely on it. No idea how to solve it without that, other then what I'm doing here. Suggestions welcome.

marcofranssen:
@edwbuck are you OK on how the feedback was addressed? Can we move this one forward?

edwbuck (approved):

PR #138, opened 2023-03-18
Add maturity tag

faisal-memon:
See here for more info: https://github.com/spiffe/spiffe/blob/main/MATURITY.md

### DIFF
--- a/README.md
+++ b/README.md
@@ -3,6 +3,7 @@
 # SPIFFE Helm Charts
 
 [![Apache 2.0 License](https://img.shields.io/github/license/spiffe/helm-charts?style=for-the-badge)](https://opensource.org/licenses/Apache-2.0)
+[![Development Phase](https://github.com/spiffe/spiffe/blob/main/.img/maturity/dev.svg)](https://github.com/spiffe/spiffe/blob/main/MATURITY.md#development)
 
 A suite of [Helm Charts](https://helm.sh/docs) for standardized installations of SPIRE components in Kubernetes environments.
 


README.md
```diff
@@ -3,6 +3,7 @@
 # SPIFFE Helm Charts
 
 [![Apache 2.0 License](https://img.shields.io/github/license/spiffe/helm-charts?style=for-the-badge)](https://opensource.org/licenses/Apache-2.0)
+[![Development Phase](https://github.com/spiffe/spiffe/blob/main/.img/maturity/dev.svg)](https://github.com/spiffe/spiffe/blob/main/MATURITY.md#development)
```
kfox1111, line 6: I think this needs to be in README.md.gotmpl

faisal-memon: I dont think we have one for the main readme

kfox1111: https://github.com/spiffe/helm-charts/blob/main/charts/spire/README.md.gotmpl ?

faisal-memon: That's in a sub chart. We want this to be in the root of the repo.

marcofranssen (changes requested):
Use same badge styling as the other badge.

README.md
kfox1111: Oh. Chart vs root readme. Your right. sorry.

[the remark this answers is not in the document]
faisal-memon: i like the consistency but it cuts off the spire logo.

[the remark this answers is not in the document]
faisal-memon: I updated the license badge to match the spiffe one. That was easier than base 64 encoding.

marcofranssen (approved):

PR #140, opened 2023-03-18
Remove @dennisgove from CODEOWNERS

dennisgove:
I continue to support this project but I'm not able to provide it the level of time and focus necessary to realistically remain a listed codeowner. In fairness to other CODEOWNERs I think it appropriate to remove myself from the list.


faisal-memon (approved):
Thanks @dennisgove for your contributions.

Issue #31, opened 2023-02-21
Tornjak support


faisal-memon:

faisal-memon:
Can this be closed as resolved now?

faisal-memon:
> Should we close it, or keep it as a placeholder for ingress & auth that still need done? (things to make it not experimental)

Wasnt sure if anything major needed to be done. Can keep this one open or open specific issues for the remaining items. Either is fine with me.

PR #144, opened 2023-03-20
Introduction of Tornjak to SPIRE Server helm charts

mrsabath:
This PR introduces a simplified version of [Tornjak](https://github.com/spiffe/tornjak)  to support UI and SPIRE control plane. 

It extends the `/charts/spire-server` sub-chart by injecting *tornjak* container to the *spire-server* pod. Tornjak image consists of Tornjak API (Tornjak Backend), Tornjak UI (Frontend), and database for storing Tornjak specific information. 

Tornjak Backend communicates with SPIRE API via SPIRE Server socket. 

Since Tornjak UI (Frontend) is a React code that renders in a browser, it needs to communicate with Tornjak APIs, and that requires communication ports to be open, either via Ingress (in Cloud deployment) or via port forwarding in local deployments (e.g. kind, minikube etc). Typically port 10000 is used for HTTP connection to Backend, and port 3000 for the HTTP connection to Frontend. 

End user management, TLS, and mTLS connections will be addressed by future PRs. 

This PR resolves issue #31


charts/spire/charts/spire-server/templates/statefulset.yaml
```diff
@@ -137,6 +137,45 @@ spec:
...
+            httpGet:
+              scheme: HTTP
+              port: 3000  
+            failureThreshold: 6
+            initialDelaySeconds: 60
+            periodSeconds: 30
+            successThreshold: 1
+            timeoutSeconds: 10
+          env:
+          {{- if .Values.tornjak.config.frontend }}
```
kfox1111, line 156: If the frontend is just a webserver that calls the backend, is there any reason it shouldn't be its own deployment or maybe even chart?

charts/spire/charts/spire-server/templates/tornjak-config.tpl
```diff
@@ -0,0 +1,38 @@
+{{- if eq (.Values.tornjak.enabled | toString) "true" }}
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: tornjak-config
+  namespace: {{ .Values.namespace }}
```
kfox1111, line 6: Should use the namespace macro, since it contains more override logic.

charts/spire/charts/spire-server/templates/tornjak-config.tpl
```diff
@@ -0,0 +1,38 @@
+{{- if eq (.Values.tornjak.enabled | toString) "true" }}
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: tornjak-config
```
kfox1111, line 5: Name should be templated.

charts/spire/charts/spire-server/templates/tornjak-config.tpl
```diff
@@ -0,0 +1,38 @@
...
+        plugin_data {
+          drivername = "{{ .Values.tornjak.config.backend.dataStore.driver }}"
+          # TODO is this a good location?
+          filename = "{{ .Values.tornjak.config.backend.dataStore.file }}"
+        }
+      }
+      {{- end }}
+
+      {{- if .Values.tornjak.config.enableUserMgmt }}
+      UserManagement "KeycloakAuth" {
```
kfox1111, line 27: Nice. :)

kfox1111: Are there non KeycloakAuth UserManagement plugins? If so, should the config take that into account?

charts/spire/charts/spire-server/templates/statefulset.yaml
```diff
@@ -174,6 +213,14 @@ spec:
           configMap:
             name: {{ include "spire-controller-manager.fullname" . }}
         {{- end }}
+        {{- if eq (.Values.tornjak.enabled | toString) "true" }}
+        {{- if .Values.tornjak.config }}
+        - name: tornjak-config
```
kfox1111, line 218: A lot of the object names are not templated out. They should be throughout the chart.

charts/spire/charts/spire-server/templates/tornjak-ingress.tpl
```diff
@@ -0,0 +1,54 @@
+{{- if eq (.Values.tornjak.enabled | toString) "true" }}
```
kfox1111, line 1: I'm thinking maybe we should skip the ingress stuff entirely for this PR? We can do a full blown tls supported one there.

charts/spire/charts/spire-server/values.yaml
```diff
@@ -197,3 +197,32 @@ nodeAttestor:
...
+# tornjak - Tornjak specific configuration
+tornjak:
+  enabled: true
+  # image - Tornjak image (frontend + backend) if not separated above
+  image: # ghcr.io/spiffe/tornjak
+    registry: ghcr.io
+    repository: spiffe/tornjak
+    pullPolicy: IfNotPresent
+    # Overrides the image tag whose default is the chart appVersion.
+    # TODO we should use a specific Tornjak version instead of 'latest'
```
kfox1111, line 210: +1. What would it take to fix it so that latest isnt needed?

kfox1111 (changes requested):
This looks like a good start. Comments inline.

charts/spire/charts/spire-server/templates/tornjak-config.tpl
mrsabath: My bad, I missed that one....

charts/spire/charts/spire-server/templates/tornjak-config.tpl
mrsabath: This should work with any Authentication that supports OAuth 2.0, not just Keycloak, but I want to bring this as a followup PRs, after we introduce intitial Tornjak components.

charts/spire/charts/spire-server/templates/NOTES.txt
```diff
@@ -4,3 +4,18 @@ Installed {{ .Chart.Name }}…
 
   kubectl exec -n {{ .Release.Namespace }} {{ include "spire-server.fullname" . }}-0 -c spire-server -- \
     spire-server entry show
+
```
kfox1111, line 7: Needs a conditional around it if only Tornjak is enabled?

charts/spire/charts/spire-server/templates/tornjak-ingress.tpl
```diff
@@ -0,0 +1,54 @@
...
+
+{{- if .Values.tornjak }}
+{{- if .Values.tornjak.config }}
+
+{{- if .Values.tornjak.config.frontend }}
+{{- if .Values.tornjak.config.frontend.ingress }}
```
kfox1111, line 7: Should probably stub out tornjak.config.frontend config in values, so you don't need to check for all the things here.

charts/spire/charts/spire-server/templates/statefulset.yaml
```diff
@@ -137,6 +137,45 @@ spec:
               mountPath: /tmp
               readOnly: false
         {{- end }}
+
```
kfox1111, line 140: up at the very top of the file, should have:
a configmap checksum and annotation so that it updates on value chage.

kfox1111:
A couple of simple tests (frontend, and backend port) added here would be good:
charts/spire/charts/spire-server/templates/tests/test-connection.yaml

Can pattern it off of the wget tests that are here:
charts/spire/charts/spiffe-oidc-discovery-provider/templates/tests/test-connection.yaml

charts/spire/charts/spire-server/templates/tornjak-ingress.tpl
mrsabath: Yes, to simplify things even more, I will remove the ingress for now.

charts/spire/charts/spire-server/templates/NOTES.txt
```diff
@@ -4,3 +4,15 @@ Installed {{ .Chart.Name }}…
...
+{{- if eq (.Values.tornjak.enabled | toString) "true" }}
+
```
kfox1111, line 9: Hmm... Without a networkpolicy or some kind of user auth, anyone within the cluster can poke at things? Should we put a hard to miss  warning here?

charts/spire/charts/spire-server/templates/service.yaml
```diff
@@ -26,3 +26,36 @@ spec:
...
+
+{{- if eq (.Values.tornjak.enabled | toString) "true" }}
+---
+apiVersion: v1
+kind: Service
+metadata:
+  namespace: {{ include "spire-server.namespace" . }}
+  name: {{ include "spire-tornjak.frontend" . }}
+spec:
+  type: LoadBalancer
```
kfox1111, line 38: ClusterIP?

charts/spire/charts/spire-server/templates/service.yaml
```diff
@@ -26,3 +26,36 @@ spec:
...
+      targetPort: 3000
+      protocol: TCP
+---
+apiVersion: v1
+kind: Service
+metadata:
+  namespace: {{ include "spire-server.namespace" . }}
+  name: {{ include "spire-tornjak.backend" . }}
+spec:
+  type: LoadBalancer
```
kfox1111, line 53: ClusterIP?

charts/spire/charts/spire-server/templates/NOTES.txt
mrsabath: Since this is just for a demo purpose running locally on *kind* or *minikube*, should we really be concern? 
In production we would be using Ingress and User Management that controls the Tornjak API access.

kfox1111: The feature is at demo level, but people might start utilizing it without realizing that? We want to make sure they know that is the case.

mrsabath: OK, I will put some Warning message on the screen via NOTES

charts/spire/charts/spire-server/values.yaml
mrsabath: Will fix it on the next update. It requires a minor change on Tornjak to be consistent between various Tornjak deployment types.

charts/spire/charts/spire-server/templates/NOTES.txt
```diff
@@ -4,3 +4,21 @@ Installed {{ .Chart.Name }}…
...
+### WARNING ### 
+For demo purpose we can enable Kubernetes Port Forwarding 
+to get access to Tornjak APIs (Backend) and Tornjak UI (Frontend).
```
kfox1111, line 12:
```suggestion
Tornjak is configured without authentication. It is only intended for
testing for now. Please do not use it in production.
```

charts/spire/charts/spire-server/values.yaml
kfox1111: There seems to be work on getting releases here:
https://github.com/spiffe/tornjak/pull/142

Since there is a big warning about the auth feature anyway, I don't think this is a showstopper for merging. We can continue to work on it to get it out of that testing feature mode in follow on pr's.

kfox1111 (approved):

Issue #139, opened 2023-03-18
Image mirroring


kfox1111:
Before 1.0, we need to ensure all images can be overridden to point to local mirrors.

faisal-memon:
Would a global registry flag help here?

kfox1111:
+1 to a global registry flag. was planning on making a pr for that at some point.

marcofranssen:
I think we support this since merging the test images PR. @kfox1111 can you confirm and close the issue?

kfox1111:
Yeah. https://github.com/spiffe/helm-charts/pull/186 was the last of it.

But we should be careful not to reintroduce it when we break out the charts.

Closing.

PR #155, opened 2023-03-22
Ensure images are overridable

kfox1111:
This test looks for any statically defined images. On finding one it lists out the filename and a request to fix it.
It also fixes any known instance of the failure

fixes: https://github.com/spiffe/helm-charts/issues/139


.github/workflows/helm-chart-ci.yaml
```diff
@@ -33,6 +33,28 @@ jobs:
...
         run: ./helm-docs.sh
 
+  ensure-image-overrides:
+    runs-on: ubuntu-22.04
+
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v3.4.0
+        with:
+          fetch-depth: 0
```
marcofranssen, line 43: Only required for ct.

In this job you can remove the with block. As we don't need the other commits here.

.github/workflows/helm-chart-ci.yaml
```diff
@@ -33,6 +33,28 @@ jobs:
...
+      - name: Check objects for images without overrides
+        run: |
+          set +e
+          set -o pipefail
+          # Look for image: defintions that are not templated. If we find none, exit is not 0 and we invert the error code to get the
+          # test to pass
+          grep -r image: charts/spire  | grep templates/ | grep -v 'image: {{ template "'
```
marcofranssen, line 50: Should we quote the greps?

.github/workflows/helm-chart-ci.yaml
```diff
@@ -33,6 +33,28 @@ jobs:
...
+          res=$?
+          if [ $res -eq 0 ]; then
+            echo The previously listed templates were found to be using statically defined images and not overridable ones. Please fix.
```
marcofranssen, line 53: Maybe write to the GITHUB_STEP_SUMMARY as markdown so it is nicely shown in the job summary.

.github/workflows/helm-chart-ci.yaml
```diff
@@ -33,6 +33,28 @@ jobs:
...
+            exit 1
+          fi
+          exit 0
```
marcofranssen, line 56: Not required. If it reached here this already happens implicit.

marcofranssen (commented):
Few small suggestions. Test in pipeline looks like an interesting idea 👍 

Can't check it from my phone though.

.github/workflows/helm-chart-ci.yaml
```diff
@@ -33,6 +33,26 @@ jobs:
...
+          set -o pipefail
+          # Look for image: defintions that are not templated. If we find none, exit is not 0 and we invert the error code to get the
+          # test to pass
+          grep -r "image:" "charts/spire"  | grep "templates/" | grep -v 'image: {{ template "' > /tmp/findings
+          res=$?
+          if [ $res -eq 0 ]; then
+            (echo These templates were found to be using statically defined images and not overridable ones. Please fix.
+            cat /tmp/findings) >> "$GITHUB_STEP_SUMMARY"
+            exit 1
+          fi
```
marcofranssen, line 54:
```suggestion
          if [ "$res" -gt 0 ]; then
            (echo These templates were found to be using statically defined images and not overridable ones. Please fix.
            cat /tmp/findings) >> "$GITHUB_STEP_SUMMARY"
            exit $res
          fi
```

I think the if statement has a mistake.

kfox1111: The test I think is correct, as described in the comment. The negated grep throws off the look of it a bit.

The rest of the suggestion looks good though.

kfox1111: oh. and res will be 0 in that case, so the exit 1 is correct too I think to get the github job to fail.

kfox1111: I pulled in the rest.

.github/workflows/helm-chart-ci.yaml
```diff
@@ -33,6 +33,31 @@ jobs:
...
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v3.4.0
+      - name: Check objects for images without overrides
+        run: |
+          set +e
+          set -o pipefail
+          # Look for image: defintions that are not templated. If we find none, exit is not 0 and we invert the error code to get the
+          # test to pass
+          grep -r "image:" "charts/spire"  | grep "templates/" | grep -v 'image: {{ template "' > /tmp/findings
```
faisal-memon, line 48: Do you need the quotes around `charts/spire`?

.github/workflows/helm-chart-ci.yaml
faisal-memon: I think you can simplify the grep a little bit:

```
grep -r "image:" charts/spire  --include "*/templates/*" | grep -v 'image: {{ template "'
```

.github/workflows/helm-chart-ci.yaml
marcofranssen:
```suggestion
          grep -r "image:" charts/spire | grep "templates/" | grep -v 'image: {{ template "' > /tmp/findings
```

charts/spire/charts/spiffe-oidc-discovery-provider/values.yaml
```diff
@@ -155,3 +155,10 @@ ingress:
   #  - secretName: chart-example-tls
   #    hosts:
   #      - oidc-discovery.example.org
+
+tests:
+  busybox:
+    image:
+      registry: docker.io
+      repository: library/busybox
+      version: latest
```
marcofranssen, line 164: Chainguard images are known to be secured and smaller then others.

I like the fact they are close to 0 or 0 CVEs, they include signatures, provenance, and sboms.

https://github.com/chainguard-images/images/tree/main/images/busybox

```suggestion
      registry: cgr.dev
      repository: chainguard/busybox
      version: glibc-1.36
```

charts/spire/charts/spiffe-oidc-discovery-provider/README.md
```diff
@@ -68,6 +68,9 @@ A Helm chart to install the SPIFFE OIDC discovery provider.
 | telemetry.prometheus.nginxExporter.image.version | string | `"0.11.0"` |  |
 | telemetry.prometheus.nginxExporter.resources | object | `{}` |  |
 | telemetry.prometheus.port | int | `9988` |  |
+| tests.busybox.image.registry | string | `"docker.io"` |  |
+| tests.busybox.image.repository | string | `"library/busybox"` |  |
+| tests.busybox.image.version | string | `"latest"` |  |
```
marcofranssen, line 73:
```suggestion
| tests.busybox.image.registry | string | `"cgr.dev"` |  |
| tests.busybox.image.repository | string | `"chainguard/busybox"` |  |
| tests.busybox.image.version | string | `"glibc-1.36"` |  |
```

marcofranssen (changes requested):
Small suggestion to use the Chainguard images.

In addition we also don't hit the Docker Hub rate limits that fast

charts/spire/charts/spiffe-oidc-discovery-provider/values.yaml
faisal-memon: +1 for chainguard. can we use `latest`  instead of a specific version?  So we dont have to keep updating it.

marcofranssen: We could, although it is a bad practice to use latest tags in general.

faisal-memon:
Instead of changing all the tests and adding a `tests` section to the values.yaml, how about just excluding the tests directory and allowing those images to be hard coded?

charts/spire/charts/spiffe-oidc-discovery-provider/values.yaml
faisal-memon: For an appversion i wouldnt use latest. For a busybox container to use in a test i think its ok.

.github/workflows/helm-chart-ci.yaml
kfox1111: No, but there isn't really a need to put explicit quotes around the image: either. so I was being consistent.

charts/spire/charts/spiffe-oidc-discovery-provider/values.yaml
kfox1111: I'm not against changing things to use chainguard, but there are other busyboxes in the chart other then the ones I had to modify here. Lets have a different pr do the chainguard stuff.

kfox1111:
> Instead of changing all the tests and adding a `tests` section to the values.yaml, how about just excluding the tests/ directory and allowing those images to be hard coded?
> 
> ```
> grep -r "image:" charts/spire  --include "*/templates/*" --exclude "*/tests/*" | grep -v 'image: {{ template "'
> ```

No. Tests need to work on isolated clusters too. Hardcoding images is an anti best practice.

.github/workflows/helm-chart-ci.yaml
kfox1111: Tried it. It doesn't work...
For testing, you can change the last grep to something else and it should show some 'image: xxxx' entries. it fails to do so with the recommendation. Not sure why off hand.

faisal-memon:
> > Instead of changing all the tests and adding a `tests` section to the values.yaml, how about just excluding the tests/ directory and allowing those images to be hard coded?
> > ```
> > grep -r "image:" charts/spire  --include "*/templates/*" --exclude "*/tests/*" | grep -v 'image: {{ template "'
> > ```
> 
> No. Tests need to work on isolated clusters too. Hardcoding images is an anti best practice.

Why is that important? I don't like the extra config supporting that introduces.

kfox1111:
Isolated clusters are not a second class use case IMO. In fact, when it comes to a security sensitive service such as spire, it could even be considered to be made a recommendation.

Same with it as a production recommendation. For bigger organizations, even with internet connectivity, having images mirrored also helps:
1. ensure you don't run up against the docker hub pull limits (saves the docker org some money too)
2. ensure you can make changes when having connectivity issues

faisal-memon:
> Isolated clusters are not a second class use case IMO. In fact, when it comes to a security sensitive service such as spire, it could even be considered to be made a recommendation.
> 
> Same with it as a production recommendation. For bigger organizations, even with internet connectivity, having images mirrored also helps:
> 
> 1. ensure you don't run up against the docker hub pull limits (saves the docker org some money too)
> 2. ensure you can make changes when having connectivity issues

Switching to chainguard will address the pull limit issue.  @marcofranssen what do you think?

.github/workflows/helm-chart-ci.yaml
```diff
@@ -33,6 +33,31 @@ jobs:
       - name: Verify Docs updated
         run: ./helm-docs.sh
 
+  ensure-image-overrides:
+    runs-on: ubuntu-22.04
+
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v3.5.0
```
marcofranssen, line 41:
```suggestion
```

Just rename the job at line 26 to `checks` or something like that.

This saves a clone for this very lightweight operation.

marcofranssen:
> > Isolated clusters are not a second class use case IMO. In fact, when it comes to a security sensitive service such as spire, it could even be considered to be made a recommendation.
> > Same with it as a production recommendation. For bigger organizations, even with internet connectivity, having images mirrored also helps:
> > 
> > 1. ensure you don't run up against the docker hub pull limits (saves the docker org some money too)
> > 2. ensure you can make changes when having connectivity issues
> 
> Switching to chainguard will address the pull limit issue. @marcofranssen what do you think?

I'm all for as few values as we can to support the majority of usecases. So having some of these things hardcoded for tests feels ok for me.

Are people really going to run `helm test` on their locked down prod clusters? Or will they do that in their Dev and QA clusters which will most probably be able to pull images from outside. Switching to the chainguard ones at least resolves the Pull limits.

kfox1111:
> 
> Switching to chainguard will address the pull limit issue. @marcofranssen what do you think?

Will it? The docker corporation had to enable them due to the cost of ruining the service.  The Kubernetes org just moved their repos to another one to help deal with the cost. Are we going to have the same issues with chainguard at some point once its more commonly used? Wouldn't mirroring ability help prevent that being a problem?

.github/workflows/helm-chart-ci.yaml
kfox1111: done

kfox1111:
We may even want to mirror the busybox test image within github for less cost and more reliable testing.

faisal-memon (changes requested):
This test looks great overall. Changes needed to get this merged:
1. Don't need to configurability of tests image at this time. Remove `tests` section from values.yaml files
2. Use [updated grep](https://github.com/spiffe/helm-charts/pull/155#issuecomment-1484692359) statement to remove tests/ directory from image template check

Shouldn't need any changes to existing tests with above changes. @marcofranssen Has a suggestion on how to override tests images using kustomize.

Alternatively can create an issue to allow tests to run offline. If we can do without adding additonal things to values.yaml, or many other people need we can reconsider.

kfox1111:
This pr's very broken. not sure how. Going to make a new one.

Issue #141, opened 2023-03-20
Values in root chart using yaml variables are not overwritable


marcofranssen:
Values like `clusterName` and `clusterDomain` and `trustBundle` are not overwriteable.

https://github.com/spiffe/helm-charts/blob/main/charts/spire/values.yaml#L14-L17

When defining different values for these, they are still using the values as defined in the main chart.

e.g.:

```yaml
spire-server:
  clusterName: my-cluster
  clusterDomain: my-cluster-domain.com
  trust-bundle: my-trust-domain
```

The solution can probably be found in https://helm.sh/docs/chart_template_guide/subcharts_and_globals/

kfox1111:
yeah. having it read from .Values.global.spire.clusterName/Domain/trustBundle too would be nice.

PR #156, opened 2023-03-23
Enable global config for clusterName, trustDomain, and bundleConfigMap

kfox1111:
This patch updates the charts so that the clusterName, trustDomain, and bundleConfigMap settable via globals or individually.

fixes: https://github.com/spiffe/helm-charts/issues/141


marcofranssen (commented):
Do we have to redefine the global in all subcharts? Or can we also just use the dig approach? Global only daarna to make sense in the error chart. At least the API feels weird to me.

kfox1111:
> Do we have to redefine the global in all subcharts? Or can we also just use the dig approach? Global only daarna to make sense in the error chart. At least the API feels weird to me.

We could do it either way I think. I'm not tied to which way we go. It seemed like the macros might be a little more readable this way?  we could leave globals out of the subcharts in this pr, but when we go to split the charts out to standalone, that bit would need to be added at that point.

charts/spire/charts/spire-agent/values.yaml
```diff
@@ -2,6 +2,10 @@
 # This is a YAML-formatted file.
 # Declare variables to be passed into your templates.
 
+global:
+  spire:
+    clusterName: example-cluster
```
faisal-memon, line 7: Not sure about having a `global` section in a sub chart.

charts/spire/values.yaml
```diff
@@ -1,8 +1,11 @@
-# You can enable features that affect all services here.
-# global:
-#   telemetry:
-#     prometheus:
-#       enabled: true
+# You can enable config/features that affect all services here.
+global:
```
faisal-memon, line 2: Do we have to have the extra `global` level or can we just assume if its not under `spire-server`, `spire-agent`, etc. that its global?

marcofranssen: I think global in the root chart is fine, not sure on the sub-charts though.

- https://helm.sh/docs/chart_template_guide/subcharts_and_globals/#global-chart-values
- https://helm.sh/docs/chart_template_guide/subcharts_and_globals/#helm

kfox1111: Removed.

kfox1111: > Do we have to have the extra `global` level or can we just assume if its not under `spire-server`, `spire-agent`, etc. that its global?

global is a special value section that helm copies verbatim  into subcharts. no other values can do it. It has to be called global to work.

You can read more about it at:
https://helm.sh/docs/chart_template_guide/subcharts_and_globals/#global-chart-values

marcofranssen:
@kfox1111 `partially fixes` will close the issue because `fixes` is in front of the issue number. Could you change that so the issue remains open to fix the remainder before closing it.

charts/spire/charts/spire-server/templates/_helpers.tpl
```diff
@@ -111,6 +111,14 @@ Create the name of the service account to use
 {{- end }}
 {{- end }}
 
+{{- define "spire-server.cluster-name" }}
+{{- if ne (len .Values.clusterName) 0 }}
+{{- .Values.clusterName }}
+{{- else }}
+{{- .Values.global.spire.clusterName }}
+{{- end }}
+{{- end }}
```
marcofranssen, line 120: Can't we do this with the `dig` approach.

So if global is set we take `global` else clustername.

This allows us to still have the defaults when we later on make spire-server and spire-agent a root level chart. That way I think it might work as well without having to change defaults for root level charts.

charts/spire/values.yaml
marcofranssen: Thank Kevin, re-reading those globals docs your original with globals in subcharts was probably the better choice. Also considering the fact if we move the subcharts later out to act as a root chart I think using the global config we can maintain a more stable API. Also see my previous suggestion, maybe we should discuss this one in a call on slack to align on the options, pros, and cons. Sorry for this back and forth, also just learning about this.

faisal-memon: globals section now makes sense to me too.

charts/spire/charts/spire-server/values.yaml
```diff
@@ -82,7 +82,7 @@ dataStorage:
 logLevel: info
 jwtIssuer: oidc-discovery.example.org
 
-clusterName: example-cluster
+clusterName: ""
```
marcofranssen, line 85: I think we should revert this part, so when the chart moves out as subchart all still works the same.

charts/spire/charts/spire-server/README.md
```diff
@@ -22,7 +22,7 @@ A Helm chart to install the SPIRE server.
 | autoscaling.maxReplicas | int | `100` |  |
 | autoscaling.minReplicas | int | `1` |  |
 | autoscaling.targetCPUUtilizationPercentage | int | `80` |  |
-| bundleConfigMap | string | `"spire-server"` |  |
+| bundleConfigMap | string | `"spire-bundle"` |  |
```
kfox1111, line 25: The default was wrong here but overridden in the main values. Once removed from the main values, showed up as broken here.

faisal-memon: good catch

charts/spire/values.yaml
```diff
@@ -1,8 +1,16 @@
...
+  spire:
+    # -- Override all instances of clusterName
```
faisal-memon, line 4: Instead of `Override...` suggest just to use the description for the actual configurable. You can put a comment at the top of the section noting that the settings in this section apply globally and override stuff in subcharts.

charts/spire/README.md
```diff
@@ -105,20 +105,16 @@ Kubernetes: `>=1.21.0-0`
 | Key | Type | Default | Description |
 |-----|------|---------|-------------|
 | fullnameOverride | string | `""` |  |
+| global.spire.bundleConfigMap | string | `""` | Override all instances of bundleConfigMap |
+| global.spire.clusterName | string | `""` | Override all instances of clusterName |
+| global.spire.trustDomain | string | `""` | Override all instances of trustDomain |
```
faisal-memon, line 110: What do you think of populating default values here instead of the subchart level?

marcofranssen: I would expect this to be filled as well.

kfox1111: I've done it for two of them. clusterName, and trustDomain. They always need to be the same in all charts.

bundleConfigMap needs to be separately configurable in nestedSpire. With globals recoded the way it is, they take preference to local settings, so you can't set it globally and override it in a single chart.

If we had nested spire pr merged, we'd catch such issues in testing.

charts/spire/values.yaml
```diff
@@ -1,8 +1,16 @@
...
-#       enabled: true
+# You can enable config/features that affect all services here.
+global:
+  spire:
+    # -- Set the name of the Kubernetes cluster
+    clusterName: "example-cluster"
+    # -- Set the trust domain to use for the spiffe identifiers
+    trustDomain: "example.org"
+    # -- Override all instances of bundleConfigMap
+    bundleConfigMap: ""
```
marcofranssen, line 9: Why don't we set the bundle here?

kfox1111: https://github.com/spiffe/helm-charts/pull/156#discussion_r1155107271

marcofranssen (approved):

PR #161, opened 2023-03-24
Improve Spire Chart documentation

marcofranssen:


charts/spire/README.md
```diff
@@ -8,9 +8,20 @@ A Helm chart for deploying the complete Spire stack including: spire-server, spi
 
 **Homepage:** <https://github.com/spiffe/helm-charts/tree/main/charts/spire>
 
-> **Warning**: Please note this chart requires Projected Service Account Tokens which has to be enabled on your k8s api server.
```
kfox1111, line 11: I can't comment on the lines above so I'll do it here. can we add the image shield relating to spire dev state from the root readme at the top here?

faisal-memon: +1

marcofranssen: Done

Issue #162, opened 2023-03-24
Test for versions matching


kfox1111:
We should have a quick test to ensure the spire versions are all set to the same version.

PR #163, opened 2023-03-25
Add a test to ensure the chart versions match

kfox1111:
We had a case where the root chart version didn't match the subchart versions. Add a test to ensure it doesn't happen again.

fixes: https://github.com/spiffe/helm-charts/issues/162


.github/workflows/helm-chart-ci.yaml
```diff
@@ -33,6 +33,28 @@ jobs:
...
+            VER=$(yq .appVersion charts/spire/charts/$FILE/Chart.yaml)
+            if [ "$VER" != "$BASEVER" ]; then
+              {
+                echo "## Version mismatch"
+                echo
+                echo "There is a mismatch between the chart versions (excluding the csi-driver). Please fix."
+              } >> "$GITHUB_STEP_SUMMARY"
+              exit 1
+            fi
+          done
```
marcofranssen, line 56:
```suggestion
      - name: Verify Spire appVersion
        run: |
          set +e
          BASEVER=$(yq e .appVersion charts/spire/Chart.yaml)
          for FILE in spiffe-oidc-discovery-provider spire-agent spire-server; do
            VER=$(yq .appVersion charts/spire/charts/$FILE/Chart.yaml)
            if [ "$VER" != "$BASEVER" ]; then
              {
                echo "## Version mismatch"
                echo
                echo "There is a mismatch between the chart versions (excluding the csi-driver and spire-controller-manager). Please fix."
              } >> "$GITHUB_STEP_SUMMARY"
              exit 1
            fi
          done
```

And rename the job name at line 26 to `checks` or something like that. Saves some build time.

.github/workflows/helm-chart-ci.yaml
```diff
@@ -33,6 +33,23 @@ jobs:
...
+        run: |
+          set +e
+          BASEVER=$(yq e .appVersion Chart.yaml)
+          for FILE in spiffe-oidc-discovery-provider spire-agent spire-server; do
+            VER=$(yq .appVersion charts/$FILE/Chart.yaml)
+            if [ "$VER" != "$BASEVER" ]; then
+              {
+                echo "## Version mismatch"
+                echo
+                echo "There is a mismatch between the chart versions (excluding the csi-driver and spire-controller-manager). Please fix."
```
faisal-memon, line 46: Can you print $VER and $BASEVER and maybe $FILE so we know where the error was.

marcofranssen (approved):
LGTM 🚀

Issue #158, opened 2023-03-24
cluster.local override


kfox1111:
Not all clusters use cluster.local for the root dns for service discovery. It should be made overridable.

PR #164, opened 2023-03-25
Make service dns domain configurable

kfox1111:
Some clusters do not use cluster.local for the root of their service dns domain. This patch makes the chart usable on those clusters.

fixes: https://github.com/spiffe/helm-charts/issues/158


charts/spire/charts/spiffe-oidc-discovery-provider/values.yaml
```diff
@@ -113,6 +113,7 @@ tolerations: []
 affinity: {}
 
 trustDomain: example.org
+serviceDnsDomain: cluster.local
```
marcofranssen, line 116: DnsDomain sounds a bit like duplicate naming.

Domain Name Service Domain 🤔 

Maybe just `clusterDomain`?

marcofranssen: Fair point.

Maybe add a documentation comment with that explanation so it ends up in the docs.

faisal-memon:
In controller manager we auto detect this by querying `kubernetes.default.svc` and extracting the domain from the reply. Can we do something similar here? https://github.com/spiffe/spire-controller-manager/blob/main/main.go#L336-L348

charts/spire/charts/spiffe-oidc-discovery-provider/values.yaml
kfox1111: Ok, I can get behind that. I didn't see anything in the helm chart but the spire-controller-manager's config having an option for it makes sense. I'll fix it up.

kfox1111:
> In controller manager we auto detect this by querying `kubernetes.default.svc` and extracting the domain from the reply. Can we do something similar here? https://github.com/spiffe/spire-controller-manager/blob/main/main.go#L336-L348

Interesting idea....

Yes and no.... the oidc configmap has it hardcoded into it by the look of it. So, the only way I can think of to fix that would be to:
1. add an emptydir
2. add an init container:
    1. Do the lookup
    2. cut it down to the subdomain
    3. template out the config from configmap a second time, injecting the discovered domain into it
    4. write the config to the emptydir
3. Modify the main container to read from the emptydir copy of the config instead of the configmap

We may or may not need a custom image to do the lookup too. not sure busybox has enough tooling in it to do step 2.

Seems possible, but is an awful lot of work for something most people don't set on their clusters?

charts/spire/charts/spiffe-oidc-discovery-provider/values.yaml
faisal-memon: We havent added it to the helm chart yet. Not sure its been released yet, it was just merged a couple weeks ago.

faisal-memon:
> > In controller manager we auto detect this by querying `kubernetes.default.svc` and extracting the domain from the reply. Can we do something similar here? https://github.com/spiffe/spire-controller-manager/blob/main/main.go#L336-L348
> 
> Interesting idea....
> 
> Yes and no.... the oidc configmap has it hardcoded into it by the look of it. So, the only way I can think of to fix that would be to:
> 
> 1. add an emptydir
> 2. add an init container:
>    
>    1. Do the lookup
>    2. cut it down to the subdomain
>    3. template out the config from configmap a second time, injecting the discovered domain into it
>    4. write the config to the emptydir
> 3. Modify the main container to read from the emptydir copy of the config instead of the configmap
> 
> We may or may not need a custom image to do the lookup too. not sure busybox has enough tooling in it to do step 2.
> 
> Seems possible, but is an awful lot of work for something most people don't set on their clusters?

Does seem like a lot of work. I was thinking a pre install hook. We can always add an auto detect feature later if we think of a good way to do it.

kfox1111:
> 
> Does seem like a lot of work. I was thinking a pre install hook. We can always add an auto detect feature later if we think of a good way to do it.

In general, I dislike hooks. and prefer if they can be avoided. I'd prefer the initcontainer thing over a hook.
I'm also slightly skeptical that the cluster domain can always be reliably detected. So a manual override may still be needed anyway?

Yeah, maybe we merge this and then consider a separate pr?

kfox1111:
Oh. I think I may understand the question. Did you mean:
Should we be setting clusterDomain: cluster.local in charts/spire/values.yaml ?

faisal-memon:
> Oh. I think I may understand the question. Did you mean: Should we be setting clusterDomain: cluster.local in charts/spire/values.yaml ?

Yes, thats what i meant

kfox1111:
Yeah. I can add that.

marcofranssen (changes requested):
.

charts/spire/charts/spiffe-oidc-discovery-provider/values.yaml
```diff
@@ -113,6 +113,8 @@ tolerations: []
 affinity: {}
 
 trustDomain: example.org
+# -- Update this if you used the kubeadm init --service-dns-domain flag
```
marcofranssen, line 116:
```suggestion
# -- This is the value of your clusters `kubeadm init --service-dns-domain` flag
```

charts/spire/values.yaml
```diff
@@ -10,6 +10,8 @@ fullnameOverride: ""
 # subcharts
 spire-server:
   enabled: true
+  # -- Update this if you used the kubeadm init --service-dns-domain flag
```
marcofranssen, line 13:
```suggestion
# -- This is the value of your clusters `kubeadm init --service-dns-domain` flag
```

charts/spire/values.yaml
```diff
@@ -21,6 +23,8 @@ spire-server:
 
 spire-agent:
   enabled: true
+  # -- Update this if you used the kubeadm init --service-dns-domain flag
```
marcofranssen, line 26:
```suggestion
# -- This is the value of your clusters `kubeadm init --service-dns-domain` flag
```

charts/spire/charts/spire-server/values.yaml
```diff
@@ -86,6 +86,8 @@ clusterName: example-cluster
 trustDomain: example.org
 
 bundleConfigMap: spire-server
+# -- Update this if you used the kubeadm init --service-dns-domain flag
```
marcofranssen, line 89:
```suggestion
# -- This is the value of your clusters `kubeadm init --service-dns-domain` flag
```

marcofranssen:
If we can discuss and settle on #156 approach we can apply same pattern here on these changes.

charts/spire/values.yaml
```diff
@@ -7,6 +7,8 @@ global:
     trustDomain: ""
     # -- Override all instances of bundleConfigMap
     bundleConfigMap: ""
+    # -- This is the value of your clusters `kubeadm init --service-dns-domain` flag
+    clusterDomain: ""
```
marcofranssen, line 11: This is a non spire var.

Lift one level up.

    global.clusterDomain

marcofranssen (changes requested):

charts/spire/values.yaml
kfox1111: And clusterDomain, while makes sense in the context of spire, could mean different things in the context of a chart depending on the spire chart. Its not very specific a value.

charts/spire/values.yaml
```diff
@@ -1,6 +1,8 @@
 # You can enable config/features that affect all services here.
 global:
   spire:
+    # -- Set to the value of your clusters `kubeadm init --service-dns-domain` flag
+    clusterDomain: "cluster.local"
```
marcofranssen, line 5:
```suggestion
global:
  # -- Set to the value of your clusters `kubeadm init --service-dns-domain` flag
  clusterDomain: "cluster.local"

  spire:
```

This isn't a spire setting. Think its better of at root level.

charts/spire/values.yaml
```diff
@@ -1,5 +1,8 @@
 # You can enable config/features that affect all services here.
 global:
+  k8s:
+    # -- Set to the value of your clusters `kubeadm init --service-dns-domain` flag
```
marcofranssen, line 4:
```suggestion
# -- This is the value of your clusters `kubeadm init --service-dns-domain` flag
```

To align with the other descriptions, also requires doc update.

charts/spire/values.yaml
```diff
@@ -1,5 +1,8 @@
 # You can enable config/features that affect all services here.
 global:
+  k8s:
+# -- This is the value of your clusters `kubeadm init --service-dns-domain` flag
+    clusterDomain: "cluster.local"
```
marcofranssen, line 5:
```suggestion
    # -- This is the value of your clusters `kubeadm init --service-dns-domain` flag
    clusterDomain: "cluster.local"
```

marcofranssen (approved):
LGTM 🚀

Issue #27, opened 2023-02-20
Prometheus support


kfox1111:
Prometheus is one of the most common monitoring stacks for Kubernetes. The chart should be deployable with support for Prometheus to be able to fetch metrics.

kfox1111:
What about Tornjak?

PR #165, opened 2023-03-26
Add podmonitors

kfox1111:
This patch adds support for podmonitors. This enables the prometheus-operator to discover all the pods and automatically configure prometheus to gather metrics from them.

partially implements: https://github.com/spiffe/helm-charts/issues/27


charts/spire/charts/spire-server/templates/podmonitor.yaml
```diff
@@ -0,0 +1,27 @@
...
+    {{- end }}
+    {{- with .Values.telemetry.prometheus.podMonitor.labels }}
+    {{- toYaml . | nindent 4 }}
+    {{- end }}
+spec:
+  selector:
+    matchLabels:
+      {{- include "spire-server.selectorLabels" . | nindent 6 }}
+  podMetricsEndpoints:
+  - port: prom
```
marcofranssen, line 22:
```suggestion
    - port: prom
```

charts/spire/charts/spire-agent/templates/podmonitor.yaml
```diff
@@ -0,0 +1,27 @@
...
+    {{- end }}
+    {{- with .Values.telemetry.prometheus.podMonitor.labels }}
+    {{- toYaml . | nindent 4 }}
+    {{- end }}
+spec:
+  selector:
+    matchLabels:
+      {{- include "spiffe-oidc-discovery-provider.selectorLabels" . | nindent 6 }}
+  podMetricsEndpoints:
+  - port: prom
```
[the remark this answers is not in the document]
kfox1111: Done.

charts/spire/charts/spiffe-oidc-discovery-provider/values.yaml
```diff
@@ -118,6 +118,11 @@ telemetry:
   prometheus:
     enabled: false
     port: 9988
+    podMonitor:
+      enabled: false
+      # - Override where to install the podMonitor
```
faisal-memon, line 123:
```suggestion
      # -- Override where to install the podMonitor, if not set will use the same namespace as oidc
```

faisal-memon: Similar suggestion for the others.

marcofranssen (changes requested):
Few nitpicky suggestions on documentation

.github/tests/prometheus/pre-install.sh
```diff
@@ -0,0 +1,3 @@
+#!/usr/bin/env bash
+
+helm install kube-prometheus-stack kube-prometheus-stack --version 45.7.1 --repo https://prometheus-community.github.io/helm-charts -n "$scenario" --wait
```
faisal-memon, line 3: Worried about this version getting out of date and having to keep it up to date. Is there a latest tag we can use instead?

kfox1111: I usually prefer to pin the versions in tests to ensure they don't break at random times and we spend a bunch of our time, or contributor time, trying to "fix a broken pr" when its the test that spontaneously breaks because upstream changed something. It lets us catch the breakage when we bump up the version of the chart in its own pr. The podMonitor api hasn't changed for ages so we don't need to keep on top of the bleeding edge of kube-prometheus-stack as well.

kfox1111: Also, we already have versions pinned for things in the other tests...

That gives me an idea though... actions can run periodically. I could put the version numbers in its own file and add a job to check for updates and either let us know, or maybe even a pr, with updated versions.

kfox1111: Here's a start of that idea: https://github.com/spiffe/helm-charts/pull/184

.github/tests/prometheus/values.yaml
```diff
@@ -2,3 +2,12 @@ global:
   telemetry:
     prometheus:
       enabled: true
+      podMonitor:
+        enabled: true
+        labels:
+          release: kube-prometheus-stack
```
marcofranssen, line 8: What is the usecase of this label?

Isn't helm already adding a `helm.sh/release` label?

Maybe to avoid confusion like mine add another dummy label if there is no specific reason for this label.

charts/spire/charts/spiffe-oidc-discovery-provider/templates/podmonitor.yaml
```diff
@@ -0,0 +1,29 @@
+{{- if .Values.insecureScheme.enabled }}
```
marcofranssen, line 1: Why only if there is an insecure Scheme? or is this a copy paste thing you missed.

kfox1111: That prometheus export only exists if the insecure Scheme flag is set. its the nginx monitor.

.github/tests/prometheus/values.yaml
kfox1111: kube-prometheus-stack only watches for podmonitors that have labels that match its own labels. So you can have multiple stacks with different labels on the same cluster. It won't be collected without the label.

charts/spire/values.yaml
```diff
@@ -11,6 +11,11 @@ global:
 #  telemetry:
 #    prometheus:
 #      enabled: true
+#      podMonitor:
+#        enabled: true
+#        # Allows to install the PodMonitor in another namespace then the spire components are installed into.
```
faisal-memon, line 16: Does need the `--` so it shows up in the readme?

marcofranssen, line 16:
```suggestion
#        # -- Allows to install the PodMonitor in another namespace then the spire components are installed into.
```

marcofranssen (approved):
LGTM 🚀

Issue #146, opened 2023-03-21
Values Profiles


kfox1111:
Having a set of prebuilt values files (see examples) for use with various production deployments may prove too burdensome for the end users to be able to juggle.

I'm proposing a possible change. First we come up with an api like:

```yaml
global:
  spire:
    # - Select a list of predefined sets of values to automatically default to.
    tags:
    - production
    - expose-oidc-ingress-nginx
    - expose-spire-server-ingress-nginx
```

What settings are for what tags are not really important at the moment.

Each chart would look for a set of tags that it supports and use the merge function to add some extra defaults into values, preferring what the user specified over its own values.

PR #166, opened 2023-03-26
Values profiles

kfox1111:
This patch enables sets of values files to be included in the chart and selected by the user at release time.

Currently a partial prototype to get feedback on the general idea.

partially implements: https://github.com/spiffe/helm-charts/issues/146


faisal-memon:
The idea seems interesting, though falls in the "nice to have" bucket, rather than "need to have". With the amount of "need to have" work still remaining I wont have time to review this. I suggest to close this for now and revisit if there is interest.

kfox1111:
Closing for now.

edwbuck:
I'm just adding this comment to document why this was closed.

This branch introduces a bunch of pretty good ideas, they are all mixed together such that the odds of getting all of the ideas submitted as one are slim.

There's support for "out of the box" ingress for specific scenarios, including federation, etc.  There's a reorganization of values that might be a good idea, and there's a few small cleanups.  Even though some of these items have merit, the combination of all of them breaks the PR process, where items are supposed to be limited in scope, to the point, and free of additional side-bar issues, to make the PR easy to perform a high quality review and easy to merge.

Kevin may opt to add back one issue (at a time) in future PRs, or he might choose not to do so, it's his prerogative.  I think some of the ideas have merit, but in combination it's just too much for a single PR.

PR #169, opened 2023-03-27
Fix formatting issues introduced with #152

marcofranssen:


kfox1111:
+1 for the fix.

Is there a way to get an enforcing test for formatting? That would be better then relying on an editor config file that not all editors read.

kfox1111:
+1 for the fix.

Is there a way to get an enforcing test for formatting? That would be better then relying on an editor config file that not all editors read.

PR #179, opened 2023-03-28
Split Tornjak Frontend into separate subchart

mrsabath:
This PR creates a separate sub-chart for Tornjak UI (backend)

PR #180, opened 2023-03-30
Revert adding tornjak to be releasable

marcofranssen:
I have created a new branch that has tornjak.

 See https://github.com/spiffe/helm-charts/tree/tornjak

 I also retargeted #179 to merge into this new tornjak branch

 This allows us to iterate with additional PRs to get tornjak into a releaseable state.

 So we can later on make a tornjak specific release without blocking the release of non tornjak changes.

 Would be great if we can target all tornjak improvements onto this new branch so we can iterete over them until reach a stable state (API wise).


faisal-memon (approved):
Thanks for taking care of this. I think it will be good to divert tornjak dev to its own branch and merge when ready.

PR #182, opened 2023-03-31
Test for configurable images

kfox1111:
This adds a test to ensure we don't add any static images to the templates. Tests are excluded.

### DIFF
--- a/.github/workflows/helm-chart-ci.yaml
+++ b/.github/workflows/helm-chart-ci.yaml
@@ -53,7 +53,7 @@ jobs:
         run: |
           set +e
           set -o pipefail
-# Look for image: definitions that are not templated. If we find none, exit is not 0 and we invert the error code to get the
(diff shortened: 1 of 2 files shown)


.github/workflows/helm-chart-ci.yaml
```diff
@@ -49,6 +49,25 @@ jobs:
             fi
           done
 
+      - name: Check objects for images without overrides
+        run: |
+          set +e
+          set -o pipefail
+# Look for image: definitions that are not templated. If we find none, exit is not 0 and we invert the error code to get the
+          # test to pass. Ignore tests for now...
```
marcofranssen, line 57:
```suggestion
          # Look for image: definitions that are not templated. If we find none, exit is not 0 and we invert the error code to get the
          # test to pass. Ignore tests for now...
```

marcofranssen (approved):
LGTM :rocket:

PR #183, opened 2023-03-31
Fix production example test

kfox1111:
The install script doesn't wait for the chart to come up and run the tests. This patch fixes it.

### DIFF
--- a/.github/tests/production-example/install.sh
+++ b/.github/tests/production-example/install.sh
@@ -1,6 +1,6 @@
 #!/usr/bin/env bash
 
-set -x
+set -xe
 
 SCRIPT=$(readlink -f "$0")
 SCRIPTPATH=$(dirname "$SCRIPT")
@@ -8,4 +8,6 @@ SCRIPTPATH=$(dirname "$SCRIPT")
 helm install \
   --namespace spire-server \
   --values "${SCRIPTPATH}/../../../examples/production/values.yaml" \
-  spire charts/spire
+  spire charts/spire --wait
+
+helm test spire --namespace spire-server
--- a/charts/spire/Chart.yaml
+++ b/charts/spire/Chart.yaml
@@ -3,7 +3,7 @@ name: spire
 description: >
   A Helm chart for deploying the complete Spire stack including: spire-server, spire-agent, spiffe-csi-driver, spiffe-oidc-discovery-provider and spire-controller-manager.
 type: application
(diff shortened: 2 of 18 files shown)


charts/spire/Chart.yaml
```diff
@@ -3,7 +3,7 @@ name: spire
 description: >
   A Helm chart for deploying the complete Spire stack including: spire-server, spire-agent, spiffe-csi-driver, spiffe-oidc-discovery-provider and spire-controller-manager.
 type: application
-version: 0.5.1
+version: 0.5.2
```
faisal-memon, line 6: Is the version bump intentional?

.github/tests/production-example/install.sh
```diff
@@ -1,11 +1,13 @@
...
 SCRIPT=$(readlink -f "$0")
 SCRIPTPATH=$(dirname "$SCRIPT")
 
 helm install \
   --namespace spire-server \
   --values "${SCRIPTPATH}/../../../examples/production/values.yaml" \
-  spire charts/spire
+  spire charts/spire --wait
+
+helm test spire --namespace spire-server
```
faisal-memon, line 13: I understand adding the `--wait`. Was this `helm test` line supposed to be there but omitted before?

charts/spire/Chart.yaml
kfox1111: So that we can see the test actually works. Cant do that without a dummy commit to the chart. Once verified, we will remove it before merge.

.github/tests/production-example/install.sh
kfox1111: helm install doesn't do helm test. both are needed and before it was skipping all the tests.

charts/spire/Chart.yaml
kfox1111: can someone take note that the tests are passing, and if there are no more changes other then removing the version bump, (ready for approval), I'll remove it and we can get it merged.

.github/tests/production-example/install.sh
marcofranssen: just for reference and completeness, most tests use the chart-testing tool to install and test the chart. `ct install` does actually run the tests, but for this test we do a specific helm-intall and therefore we have to run it here as well.

marcofranssen (approved):
LGTM 🚀

PR #196, opened 2023-04-06
Allow post-install.sh to fail the test run too.

kfox1111:
This patch enables the post-install.sh script to return non 0 exit codes that get passed back as the test result as well. This allows further tests to be processed in post-install.sh as needed.


.github/workflows/helm-chart-ci.yaml
```diff
@@ -190,6 +190,8 @@ jobs:
         run: |
           post-install() {
             [ -x "${TEST_DIR}/post-install.sh" ] && "${TEST_DIR}/post-install.sh" $1
+            res=$?
+            [ $res != 0 ] && exit $res
```
faisal-memon, line 194: why not just `exit $res`?

kfox1111: $1 is the exit code from the tests. If it succeeds, its 0 and if it fails its != 0.

This change enables post-install.sh to return 0 for don't override what the tests returned, or, !=0 == fail.

0 from post-install.sh still needs to fall through and use $1 instead of its own value, $res.

faisal-memon: sorry for dropping the ball on this one. My shell scripting knowledge is not so good. I always thought `$1` was an arg not a return value?

kfox1111: It can be two things. :)

Your right, its an arg. but the arg is defined here (just off screen):
https://github.com/spiffe/helm-charts/pull/196/files#diff-7089a3541750166ffb3b1e90b6dc6ab674351946b677a390bb4b7881120c4a2eL197
to be the exit code that the main script exited with. ($?). Traps are a little weird....

kfox1111:
I'm going to close this for now. It was somewhat preemptive as I thought I would need it soon for other tests. but I'll just resurrect it when we get there. We can't seem to merge small patches very fast anymore, so its just getting in the way.

PR #113, opened 2023-03-12
Change spire-server configmap to yaml, rendered as json

marcofranssen:
Somehow this is resulting in the following log on spire-server.

```shell
trust_domain must be configured
```

This is how my rendered config looks like

```json
{
  "health_checks": {
    "bind_address": "0.0.0.0",
    "bind_port": "8080",
    "listener_enabled": true,
    "live_path": "/live",
    "ready_path": "/ready"
  },
  "plugins": [
    {
      "DataStore": {
        "sql": {
          "plugin_data": {
            "connection_string": "/run/spire/data/datastore.sqlite3",
            "database_type": "sqlite3"
          }
        }
      }
    },
    {
      "NodeAttestor": {
        "k8s_psat": {
          "plugin_data": {
            "clusters": {
              "aft-249f0a2-cluster": {
                "service_account_allow_list": [
                  "spire-system:spire-agent"
                ]
              }
            }
          }
        }
      }
    },
    {
      "KeyManager": {
        "disk": {
          "plugin_data": {
            "keys_path": "/run/spire/data/keys.json"
          }
        }
      }
    },
    {
      "Notifier": {
        "k8sbundle": {
          "plugin_data": {
            "config_map": "spire-bundle",
            "namespace": "spire-system"
          }
        }
      }
    }
  ],
  "server": {
    "bind_address": "0.0.0.0",
    "bind_port": "8081",
    "ca_key_type": "rsa-2048",
    "ca_subject": [
      {
        "common_name": "spire.eu-west-1-marco-test.mydomain.com",
        "country": [
          "NL"
        ],
        "organization": [
          "Marco Franssen"
        ]
      }
    ],
    "ca_ttl": "24h",
    "data_dir": "/run/spire/data",
    "default_jwt_svid_ttl": "1h",
    "default_x509_svid_ttl": "4h",
    "jwt_issuer": "oidc.spire.eu-west-1-marco-test.mydomain.com",
    "log_level": "info",
    "trust_domain": "eu-west-1-marco-test.mydomain.com"
  }
}
```

PR #198, opened 2023-04-07
Add customPlugins and unsupportedBuiltInPlugins sections to spire-server

kfox1111:
This patch enables end users to configure external plugins in the spire-server config. Unsupported internal plugins are not able to be set.


examples/spire-plugins/node-agent-tpm/values.yaml
```diff
@@ -0,0 +1,25 @@
...
+    repository: kfox1111/misc
+    version: tpm-server-test
+  plugins:
+    NodeAttestor:
+      tpm:
+        plugin_cmd: /bin/tpm_attestor_server
+        plugin_checksum: 97442358ae946e3fb8f2464432b8c23efdc0b5d44ec1eea27babe59ef646cc2f
+        plugin_data:
+        - ca_path: /run/spire/data/certs
+          hash_path: /run/spire/data/hashes
```
marcofranssen, line 25:
```suggestion
  initContainers:
    - name: tpm-init
      image: busybox
      command:
        - sh
        - -c
        - |
          mkdir -p /run/spire/data/certs
          mkdir -p /run/spire/data/hashes
    volumeMounts:
      - mountPath: /run/spire/data
        name: spire-data
  image:
    registry: docker.io
    repository: kfox1111/misc
    version: tpm-server-test
  plugins:
    NodeAttestor:
      tpm:
        plugin_cmd: /bin/tpm_attestor_server
        plugin_checksum: 97442358ae946e3fb8f2464432b8c23efdc0b5d44ec1eea27babe59ef646cc2f
        plugin_data:
          - ca_path: /run/spire/data/certs
            hash_path: /run/spire/data/hashes
```

Too follow the indent style in the remainder of the project.

marcofranssen (changes requested):
I prefer to have a couple settings in here like we already have.

```yaml
spire-server:
  nodeAttestor:
    xYZ:
      enabled: true
```

and same for other plugins.

```yaml
spire-server:
  upstreamAuthority:
    xYZ:
      enabled: true
```

We could then loop the keys in those objects and people can still add their own plugins within there.

This probably adds some duplication in templates, but it also makes it more straight forward and less magic to merge things in the templates, so easier to understand and digest.

Futhermore the API will me better documented.

kfox1111:
> I prefer to have a couple settings in here like we already have.
> 
> ```yaml
> spire-server:
>   nodeAttestor:
>     xYZ:
>       enabled: true
> ```
> 
> and same for other plugins.
> 
> ```yaml
> spire-server:
>   upstreamAuthority:
>     xYZ:
>       enabled: true
> ```
> 
> We could then loop the keys in those objects and people can still add their own plugins within there.
> 
> This probably adds some duplication in templates, but it also makes it more straight forward and less magic to merge things in the templates, so easier to understand and digest.
> 
> Futhermore the API will me better documented.

There are... dragons there... But seems like there may be them here either way, maybe just less so with the plugin section?...

The issue we're going to hit, is in toYaml'ing the sections. If we're taking values, such as xYZ.enabled and jsoning it over to spire, its going to get confused by settings that are for helm, not for the actual server. We could remove them from the doc before toPrettyJson, but that can get messy too. With plugin being a different section, it allowed for the k8s config and the raw yaml things to be separate. Its all still able to be validated, as its proper yaml in helm values, so its not so bad.

I'll do a quick inventory though and see how bad removing the k8s stuff from the json at runtime would be....

kfox1111:
As expected, its much harder to do. Working prototype here:
https://github.com/spiffe/helm-charts/pull/227

Did it as an alternate pr so we can compare the two for maintainability.

kfox1111:
> @kfox1111 figure out which one of the two PRs (this one or #227) you want to move forward with on this effort, merge your best ideas into one submission, and then please close the other PR.
> 
> We can't push two cars at once. Some of the lack of velocity is that the different PRs split efforts, leading to indecision, which then becomes inaction. Nobody wants to review the PR that won't possibly merge, and we just don't know which one of the two that is.

I have no say in the matter as a contributor. only the reviewers do... If two reviewers step up and pick one they like in preference to the other, lets go that direction.

kfox1111:
> I like the simplicity of this PR over all the complexity of the alternative implementation, However this one does require changing the existing values to nest each plugin in a plugins section.
> 
> What I'm missing on this one is how we define the defaults like we have today for the upstreamAuthorities and nodeAttestors etc. Basically we are going to have a weakly typed API.

As far as this PR, we don't I think. The plugin section is a relief valve for when we don't support a particular plugin explicitly. The user still has an option to configure their plugins via passthrough.

> What would be against defining a value for each of the pluginTypes and just spec them out as opposed to doing all this magic?

Spire doesn't control all plugins. There are built in ones, of which there are already a lot, plus end users can create their own such as https://github.com/boxboat/spire-tpm-plugin/ that need supporting, and we can't possibly support pr's for every combination of plugin that users want to create/use.

> IMHO both PRs at the moment have severe drawbacks.
> 
> * This one making the chart API more weakly typed and less documented.
> * The other one having a complicated macro logic which is hard to grasp and can cause problems in the future from a maintenance point of view, but also increase the cognitive load for future contributors.

Both are valid concerns. But I can't think of a way to properly support plugins without making one of these tradeoffs. Can you?

> IMHO we just spec out a couple of values for the different plugins. When more has to be supported that means a new PR has to be raised, but at least it is very transparant and easy to understand for contributors and maintainers.

-1. The users would likely never be able to use all the plugins available to them then and we'd be chasing the problem forever. I don't think that is a viable solution.

faisal-memon:
@marcofranssen This PR similar to one I opened before for custom plugins support: https://github.com/spiffe/helm-charts/pull/70/files. 

I think we should still have all the standard plugins natively defined. Then this section would be for just custom plugins, or ones we havent natively implemented yet.

charts/spire/charts/spire-server/values.yaml
```diff
@@ -389,6 +389,12 @@ tornjak:
     #   cpu: 100m
     #   memory: 128Mi
 
+plugins:
+  KeyManager: {}
+  NodeAttestor: {}
+  UpstreamAuthority: {}
+  Notifier: {}
```
marcofranssen, line 396: This doesn't align with the naming conventions.

https://helm.sh/docs/chart_best_practices/values/#naming-conventions

kfox1111: Do we want simple and smaller/manageable code, or do we want pretty but with extra complexity? We already established that we can't have both. Just having these here at all mandated more complexity in the templating to remove the blank ones because spire can't tolerate blank ones configured.

charts/spire/charts/spire-server/templates/configmap.yaml
```diff
@@ -32,71 +44,71 @@ server:
 
 plugins:
   DataStore:
-    - sql:
-        plugin_data:
-          {{ include "spire-server.datastore-config" . | nindent 10 }}
+    sql:
```
edwbuck, line 47: As this has to do with custom plugins, I think this change is "extra".  

Also, the datastore-config is not going to be variable, I'd be happier when we get around to this to calling out the fields individually.

edwbuck (dismissed):
Please address the questions I've asked.

charts/spire/charts/spire-server/templates/configmap.yaml
kfox1111: Its so that it can do dictionary merging in helm. Array type values are overwritten. The macro converts it back to an array after merge.

charts/spire/charts/spire-server/values.yaml
kfox1111: @marcofranssen can you please confirm if you'd like the added complexity to do the lowercase first character conversion?

kfox1111:
@edwbuck I think I incorporated all the feedback you gave via slack. Thanks again for the review.

charts/spire/charts/spire-server/values.yaml
faisal-memon: Is the problem here that the first letter is upper case instead of lower case? Or something else?

marcofranssen: yes conventions for values are camelCase. We have one issue here though. Those things like KeyManager and NodeAttestor should not be in the values file over here as they are part of the builtin plugins for which we already have an implementation.

charts/spire/charts/spire-server/templates/configmap.yaml
```diff
@@ -32,71 +50,71 @@ server:
...
-    - cert-manager:
-        plugin_data:
-          issuer_name: {{ default (include "spire-server.fullname" $root) .issuer_name }}-ca
-          issuer_kind: {{ .issuer_kind | quote }}
-          issuer_group: {{ .issuer_group | quote }}
-          namespace: {{ default $root.Release.Namespace .namespace | quote }}
-          {{- if ne .kube_config_file "" }}
-          kube_config_file: {{ .kube_config_file | quote }}
-          {{- end }}
+    cert-manager:
```
marcofranssen, line 97: Previously this broke the config as the implementation expected an array? 🤔

charts/spire/charts/spire-server/templates/configmap.yaml
```diff
@@ -1,3 +1,21 @@
+{{- range $type, $tvals := .Values.customPlugins }}
+{{-   if not (has $type (list "KeyManager" "NodeAttestor" "UpstreamAuthority" "Notifier")) }}
+{{-     fail (printf "Unknown plugin type specified: %s" $type) }}
+{{-   end }}
```
marcofranssen, line 4:
```suggestion
{{-   if (has $type (list "KeyManager" "NodeAttestor" "UpstreamAuthority" "Notifier")) }}
{{-     fail (printf "Please use the existing configurables for this plugin type: %s" $type) }}
{{-   end }}
```

I think this is wrong. For UpstreamAuthority and NodeAttestors and such we have already other configurables that we use in the configMap.

charts/spire/charts/spire-server/values.yaml
kfox1111: I'll change it to camelcase. Its going to bloat up the macro, which is not worth it IMO, but I care more about getting this thing finally implemented then some extra boiler plate code.

charts/spire/charts/spire-server/templates/configmap.yaml
kfox1111: And again here: https://github.com/spiffe/helm-charts/pull/198/files#r1276506204

charts/spire/charts/spire-server/templates/configmap.yaml
kfox1111: No, this suggestion is incorrect. It makes it impossible to use any 3rd party plugin. See the included example for something that needs to work.

charts/spire/charts/spire-server/values.yaml
kfox1111: I found a way to do it using some of the string manipulation functions that doesn't require a lot of mapping. Updated.

marcofranssen (approved):
LGTM 🚀

marcofranssen:
> @edwbuck I think I incorporated all the feedback you gave via slack. Thanks again for the review.

Would be good to have the feedback in the PR. I see there is currently still a block by @edwbuck even though there is no change requested in this PR.

edwbuck:
> > @edwbuck I think I incorporated all the feedback you gave via slack. Thanks again for the review.
> 
> Would be good to have the feedback in the PR. I see there is currently still a block by @edwbuck even though there is no change requested in this PR.

My concern was that the configuration section for the item was open ended.  Basically it took a YAML object.

There are a number of subtle issues here.  First, the config file output is not YAML, it is HCL.  The syntax won't mesh up under many scenarios.  Also, since we are simply passing a whole object, there isn't checking that the object is an array or an object, or something even more simple (a value).

That latter part can be fixed, but to have valid HCL output, you might need to force this into a multi-line string.  Finally, the strings depend on the kind of the custom plugin / non-custom plugin, where some non-custom plugins are obviously without configuration, while others contain standard configurations (memory keyStore vs disk keyStore) that differ from partically required fields as in a custom keyStore.

kfox1111:
> > > @edwbuck I think I incorporated all the feedback you gave via slack. Thanks again for the review.
> > 
> > 
> > Would be good to have the feedback in the PR. I see there is currently still a block by @edwbuck even though there is no change requested in this PR.
> 
> My concern was that the configuration section for the item was open ended. Basically it took a YAML object.
> 
> There are a number of subtle issues here. First, the config file output is not YAML, it is HCL. The syntax won't mesh up under many scenarios. Also, since we are simply passing a whole object, there isn't checking that the object is an array or an object, or something even more simple (a value).
> 
> That latter part can be fixed, but to have valid HCL output, you might need to force this into a multi-line string. Finally, the strings depend on the kind of the custom plugin / non-custom plugin, where some non-custom plugins are obviously without configuration, while others contain standard configurations (memory keyStore vs disk keyStore) that differ from partially required fields as in a custom keyStore.

We need some relief valve for users able to do unsupported stuff until we can support them. They have to sign up for it being unsupported though, which I think is a good tradeoff.

The config file output isn't YAML. We build up the config in YAML since thats easy to do in helm, then convert it to JSON (also easy to do in helm) and spire has the ability to take its config in either HCL or JSON. so we have a direct path via JSON. This avoids all the ugliness of generating HCL. This was done a while ago in https://github.com/spiffe/helm-charts/pull/113

faisal-memon (approved):

Issue #199, opened 2023-04-07
Switch value to persistence to follow helm convention


kfox1111:

faisal-memon:
This is now completed.

PR #200, opened 2023-04-07
Switch to persistence

kfox1111:
This patch switches the value to persistence to follow the helm convention.

fixes: https://github.com/spiffe/helm-charts/issues/199


mrsabath (approved):
Since this PR only changes the name of the variable, LGTM.

marcofranssen:
Probably should do the same on the tornjak PR then. To keep things consistent.

PR #201, opened 2023-04-08
Add global for image registry

kfox1111:
This patch enables a user to override all image registry settings all at once to point to their own registry to enable easy custom mirrors.

partially implements: https://github.com/spiffe/helm-charts/issues/139


charts/spire/values.yaml
```diff
@@ -11,6 +11,10 @@ global:
     # -- Override all instances of bundleConfigMap
     bundleConfigMap: ""
 
+    image:
+      # -- Override all images to point at your own registry
```
marcofranssen, line 15:
```suggestion
      # -- Override all Spire image registries at once
```

marcofranssen (commented):
Doesn't necessarily have to be your own registry, would make the doc a bit more generic.

faisal-memon:
This looks good to me. @marcofranssen what do you think?

marcofranssen (approved):
LGTM 🚀

Issue #197, opened 2023-04-07
Helm hooks failed leave resources behind, causing failures on subsequent installs


marcofranssen:
On one of our clusters we have seen issues upgrading Spire to a new release of this Helm chart.

```shell
│ Error: pre-upgrade hooks failed: warning: Hook pre-upgrade spire/charts/spire-server/templates/pre-upgrade-hook.yaml failed: serviceaccounts "spire-server-pre-upgrade" already exists
```

We have been unable to determine the root cause. As a workaround we have added following workaround.

## Workaround

As a workaround prior to running `helm upgrade` we run the following in our pipelines to cleanup leftover resources from the hooks that might have failed in a previous run.

```shell
  # safetynet for failed hooks that left resources behind, root cause unknown
  spire_hooks=(spire-server-pre-upgrade spire-server-post-upgrade spire-server-post-install)
  kubectl -n spire-system delete serviceaccount "${spire_hooks[@]}" || true
  kubectl -n spire-system delete clusterrole "${spire_hooks[@]}" || true
  kubectl -n spire-system delete clusterrolebinding "${spire_hooks[@]}" || true
  kubectl -n spire-system delete job "${spire_hooks[@]}" || true
```

kfox1111:
Hence, my dislike of helm hooks. They are not always reliable. That we hit one case of breakage in this, less then 3 days after merging when we have so few users makes me again, really question this approach, along with the inability to chose an alternate as an end user.

faisal-memon:
I think adding [before-hook-creation](https://helm.sh/docs/topics/charts_hooks/#:~:text=Description-,before%2Dhook%2Dcreation,Delete%20the%20previous%20resource%20before%20a%20new%20hook%20is%20launched%20(default),-hook%2Dsucceeded) to the delete policy will resolve this problem

kfox1111:
Yeah, that sounds like it would help.

PR #202, opened 2023-04-09
Add  before-hook-creation delete policy

faisal-memon:
Adds [before-hook-creation](https://helm.sh/docs/topics/charts_hooks/#:~:text=Description-,before%2Dhook%2Dcreation,Delete%20the%20previous%20resource%20before%20a%20new%20hook%20is%20launched%20(default),-hook%2Dsucceeded) to the delete policy to ensure all hook objects are delete prior to install/upgrade

Should fix https://github.com/spiffe/helm-charts/issues/197


marcofranssen (approved):
LGTM 🚀

PR #203, opened 2023-04-09
Enable job hooks to be disabled

kfox1111:
This eliminates a bunch of code for folks that don't need the hooks:
```console
$ helm template charts/spire/ | wc -l
1150
$ helm template charts/spire/ --set spire-server.controllerManager.upgradeHooks.enabled=false | wc -l
868
```

It also runs faster in some environments.


marcofranssen:
There is a --no-hook option no need for more configs.

faisal-memon:
I agree with @marcofranssen. We also want to make sure everyone follows the same install/upgrade path.  As an alternative you can fix the [bug you mentioned](https://github.com/spiffe/helm-charts/pull/128#issuecomment-1500472884) with the --no-hooks flag.

kfox1111:
As stated several times before, --no-hook doesn't work with whats been released when you want to use the Fail policy.

kfox1111:
Closing in favor of https://github.com/spiffe/helm-charts/pull/206. Its a much simpler and still can cover the use case.

PR #206, opened 2023-04-11
Make it possible to force the webhook mode

kfox1111:
When disabling webhooks it ignores the webhook mode setting. This patch makes it work when doing that.


kfox1111:
Alternate approach to https://github.com/spiffe/helm-charts/pull/203

faisal-memon:
I dont think this is a good idea. Disabling the job hook will lead to an upgrade failure.

kfox1111:
I'll close this for now...

My needs and those of others are not being addressed though without some change.

I will try and come up with another alternate, that hopefully is more tolerable. Please keep an open mind to alternative approaches, along with making constructive suggestions. We need to come up with a solution that works for everyone. We're not there.

Issue #177, opened 2023-03-28
Charts don't support configuring to not run as root


keeganwitt:
Both the agent and server YAML files have no templating to support running as a non-root user.

kfox1111:
The server can run restricted. We're testing against it with each PR:
https://github.com/spiffe/helm-charts/blob/main/examples/production/values.yaml#L31-L52

The agent needs to be root so it can probe the unix socket for the user at the other end of the socket. It can't run as nonroot and function properly.

kfox1111:
Please reopen if there is something that you need that still isn't handled.

keeganwitt:
Sorry, I should have been clearer about how I'd been doing this. It's not perfect, but it does reduce the risk surface a little. Doing this isn't possible with the chart.

```yaml
apiVersion: apps/v1
kind: DaemonSet
metadata:
  name: agent
  namespace: {{ $.Release.Namespace }}
  labels:
    app: agent
spec:
  selector:
    matchLabels:
      app: agent
  template:
    metadata:
      annotations:
      labels:
        app: agent
    spec:
      securityContext:
        runAsUser: 10000
        runAsGroup: 10000
        fsGroup: 10000
      initContainers:
        - name: chmod-socket-directory
          image: busybox
          command: [ "sh", "-c", "chown -R 10000:10000 /run/spire/socket" ]
          securityContext:
            runAsUser: 0
            runAsGroup: 0
          volumeMounts:
            - name: agent-socket
              mountPath: /run/spire/socket
```

kfox1111:
And it functions correctly? Its able to dig up all the needed information from the cgroups to attest workloads?

keeganwitt:
This is pulled from my live production cluster, yes.

kfox1111:
I'm trying to confirm with the developers if there will be any issues. So far, none have been identified. The last thing to verify is if it will continue to work with cgroupsv2 which is going to be supported soon. If it all works, I think its a great idea. :)

keeganwitt:
You're right, the server's statefulset.yaml does have a `securityContext` in it. I missed that before.

kfox1111:
Yeah. At this point, it may be something we add to examples/production/values.yaml so folks can get it setup easily.

kfox1111:
Chatted with a couple of developers. We can't find any reason why it shouldn't work.

kfox1111:
https://github.com/spiffe/spire/blob/main/doc/docker_images.md   has some more details too.

PR #209, opened 2023-04-12
Allow for SPIRE Agent to run as non root user

kfox1111:
This patch implements a locked down spire-agent along with updating the production example to use it.

fixes: https://github.com/spiffe/helm-charts/issues/177


charts/spire/charts/spire-agent/templates/daemonset.yaml
```diff
@@ -43,6 +43,22 @@ spec:
           args: ["-t", "30", "-h", "{{ include "spire-agent.server-address" . | trim }}", "-p", {{ .Values.server.port | quote }}]
           resources:
             {{- toYaml .Values.waitForIt.resources | nindent 12 }}
+        {{- if gt (dig "fsGroup" 0.0 .Values.podSecurityContext) 0.0 }}
```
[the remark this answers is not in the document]
kfox1111: Unfortunately, helm is casting integers to floats. So when I added
```yaml
podSecurityContext:
  fsGroup: 0
```

it was failing the gt test because float != int with
gt .Values.podSecurityPolicy.fsGroup 0.
Casting all the compared 0's in the helm template to 0.0 fixed the test.

Would a comment like:
{{/* casting 0's to floats via 0.0 to fix gt comparison */}}
The line above help future readers of the code?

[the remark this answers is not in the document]
kfox1111: I found a different way to fix it. can cast it back to an int. Will change it to do that.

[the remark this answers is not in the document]
kfox1111: Done

faisal-memon:
whats the bash container for?

kfox1111:
> whats the bash container for?

runs as root:
https://github.com/spiffe/helm-charts/pull/209/files#diff-e860a3dbff565712efc8045d2a42937c0f58f6994939846799346e253a182e75R59-R60

Then chowns the socket dir so the nonroot user the agent runs as can get to it:
https://github.com/spiffe/helm-charts/pull/209/files#diff-e860a3dbff565712efc8045d2a42937c0f58f6994939846799346e253a182e75R52

This limits the amount of root needed to a second or less at the startup time of the agent.

charts/spire/charts/spire-agent/templates/daemonset.yaml
```diff
@@ -43,6 +43,22 @@ spec:
...
+          image: {{ template "spire-lib.image" (dict "image" .Values.bash.image "global" .Values.global) }}
+          imagePullPolicy: {{ .Values.bash.image.pullPolicy }}
+          command: ["bash", "-c"]
+          args:
+            - "chown -R {{ .Values.podSecurityContext.runAsUser }}:{{ .Values.podSecurityContext.fsGroup }} {{ include "spire-agent.socket-path" . | dir }}"
+          resources:
+            {{- toYaml .Values.fsGroupFix.resources | nindent 12 }}
+          volumeMounts:
+          - name: spire-agent-socket-dir
+            mountPath: {{ include "spire-agent.socket-path" . | dir }}
```
marcofranssen, line 57:
```suggestion
            - name: spire-agent-socket-dir
              mountPath: {{ include "spire-agent.socket-path" . | dir }}
```

marcofranssen (changes requested):
Can't we do this without this without the init container?

This was a pretty recent change on the spire containers, see this commit https://github.com/spiffe/spire/commit/0b91ce28224d5846bc25ce4220ec8b1f550e5a05

I think we should be able to do without the bash container now.
We just need to ensure we mount the volumes on these default paths in the container as well keep the sockets in those default locations and ensure the volumemounts use the same UID and GID when mounting the volumes.

https://github.com/spiffe/spire/blob/main/Dockerfile#L28-L54

charts/spire/charts/spire-agent/values.yaml
```diff
@@ -94,6 +94,22 @@ waitForIt:
     tag: latest-20230517
   resources: {}
 
+bash:
```
faisal-memon, line 97: Maybe a small comment here that this will be used to patch up the container to run as non-root?

charts/spire/charts/spire-agent/values.yaml
```diff
@@ -94,6 +94,22 @@ waitForIt:
...
+    repository: chainguard/bash
+    # -- The image pull policy
+    pullPolicy: Always
+    # -- This value is deprecated in favor of tag. (Will be removed in a future release)
+    version: ""
+    # -- Overrides the image tag
+    tag: latest-20230517
+
+fsGroupFix:
+  resources: {}
```
faisal-memon, line 111: What is this used for?

charts/spire/charts/spire-agent/values.yaml
kfox1111: I was a bit on the fence if it should just be bash, or image stuck under fsGroupFix. Would the latter be more understandable? If we ever needed another image though, there may be some duplication.

charts/spire/charts/spire-agent/values.yaml
kfox1111: For users to specify the resources needed by the container. A best practice on production clusters. https://kubernetes.io/docs/concepts/configuration/manage-resources-containers/

charts/spire/charts/spire-agent/values.yaml
faisal-memon: I think so yea.

kfox1111: Updated

charts/spire/charts/spire-agent/values.yaml
faisal-memon: Can you add a comment for this?

charts/spire/charts/spire-agent/values.yaml
```diff
@@ -106,6 +106,20 @@ waitForIt:
     tag: latest-20230517
   resources: {}
 
+fsGroupFix:
```
faisal-memon, line 109: Can you add a comment above here describing what this section is for?

charts/spire/charts/spire-agent/values.yaml
kfox1111: We don't currently document it anywhere else, but gotta start somewhere I guess.

kfox1111: done

charts/spire/charts/spire-agent/values.yaml
kfox1111: done

charts/spire/charts/spire-agent/values.yaml
```diff
@@ -106,6 +106,22 @@ waitForIt:
     tag: latest-20230517
   resources: {}
 
+# When running as non root, ensure the socket path is the correct permissions
```
faisal-memon, line 109:
```suggestion
# When running as non root, needed to ensure the socket path has the correct permissions.
# Set runAsUser to non-zero value in podSecurityContext to run as non-root user
```

faisal-memon:
Can we change the PR title to something like "Allow for SPIRE Agent to run as non root user". I think that will look better in release notes and closer to what we are accomplishing here.

charts/spire/charts/spire-agent/templates/daemonset.yaml
```diff
@@ -43,6 +43,22 @@ spec:
...
+            - "chown -R {{ .Values.podSecurityContext.runAsUser }}:{{ .Values.podSecurityContext.fsGroup }} {{ include "spire-agent.socket-path" . | dir }}"
+          resources:
+            {{- toYaml .Values.fsGroupFix.resources | nindent 12 }}
+          volumeMounts:
+            - name: spire-agent-socket-dir
+              mountPath: {{ include "spire-agent.socket-path" . | dir }}
+          securityContext:
+            runAsUser: 0
+            runAsGroup: 0
+        {{- end }}
```
marcofranssen, line 61: Please check if we can do the following:

Remove the configurable for the spire-agent mountPath to use the image default.

https://github.com/spiffe/spire/blob/main/Dockerfile#L52-L54

This image default already has the right permissions. In that case we don't even need the configurable for running as root. Then we should just always run as non-root. Don't see a reason to run as root if we can by default run as non-root.

Probably remove this configurable so people can't break the permissions that should be good by default.

```
# -- The unix socket path to the spire-agent
socketPath: /run/spire/agent-sockets/spire-agent.sock
```

faisal-memon: Seems like we have a few 2 options here regarding the configurable:
1. Leave the configurable in place but default it to `/tmp/spire-agent/public/api.sock` with note that changing it will require running as root
2. Remove configurable altogether. Is it needed?

Regarding running as root, I think we can still support that with the `runAsUser` configurable but having the default be `1000`?

kfox1111:
It is defaulting to root when you don't include the production example.

charts/spire/charts/spire-agent/templates/daemonset.yaml
kfox1111: I would be happy if the configurable is not needed. I have not been able to prove that is it not needed. Someone needs to step up and find an alternate solution if this one is unacceptable.

kfox1111: Wait... I'm getting confused. there are lots of configurables at this point being discussed. Lets recap so we're all on the same page.

The agent socket needs to be configurable, as it needs to be exposed on a hostpath. It isn't possible to unhardcode that part and support nested spire or kubernetes clusters that don't install kubelet in the default location.

It may be possible to hardcode its mountpoint in the container. but that doesn't affect the permissions of that mount. The permissions come from the host/kubelet. Not sure what the benefit to that change is.

This pr changes the permissions to match what is specified by the configuration setting that a user can optionally set in the securityContext, if they have done so. This fixes the bug.

And, this pr contains the recommendation in the example pr to run as nonroot. But the user can override it back to root if needed. The default for the chart, when not using the production recommendations is to run as root.

I'm not seeing any alternative to this pr that will work. If someone wants to submit a different pr, with a workable solution for all of these things, I'd love to see one.

faisal-memon: Got it, does this make sense then:

- Change default of socket path to tmp dir
- Change default of runAsUser to 1000 or some non root value
- If socket patch is changed and runAsUser is non root, run the init container

marcofranssen: Will check in coming week to see if I can get it to work. This week was a busy week.

marcofranssen (approved):
LGTM 🚀

PR #210, opened 2023-04-12
Revert "Switch to persistence"

kfox1111:
Reverts spiffe/helm-charts#200


mrsabath (approved):
The revert looks correct.

PR #211, opened 2023-04-12
Rename dataStorage to persistence in values.yaml

kfox1111:
This patch switches the value to persistence to follow the helm convention.

fixes: https://github.com/spiffe/helm-charts/issues/199


mrsabath (approved):
I understand this is a part of the batch and it had to be reverted earlier. 
lgtm

kfox1111:
Having experimented in the database side of implementation, I'm convinced this is the right thing to do now. As the persisted volume has stuff in it other then just sqlite and will be needed even if an external database is used.

charts/spire/charts/spire-server/values.yaml
```diff
@@ -73,7 +73,8 @@ affinity: {}
 
 topologySpreadConstraints: []
 
-dataStorage:
+# Used by sqlite3 for spire-server, by the disk KeyStore, and/or by Tornjak.
```
marcofranssen, line 76: This is a bit confusing. and/or, if it is a or, it should probably have dedicated PVC so you can pick the desired storage class and capacity for that specific need.

To have it show up in the docs please use the helm-docs syntax.

kfox1111: 'and/or' is kind of a weird English thing because its imprecise and there is no word for the possibility that between 1 and all of something is possible. I'm open to a different wording though. Wording is hard.

Dedicated PVC's for each task would be quite wasteful. I've experienced bugs with pvcs that were allocated < 1gb. The disk KeyStore needs to persistently store kb of data. Tornjak and spire-server sqlite will likely be in the 10's of mb. So its already pretty wasteful at a minimal 1gb and it would be up to 3x wasteful having one per.

I'll update it to be helm-docs format.

kfox1111: Updated

faisal-memon:
I think this can be closed in favor of #225 ?

kfox1111:
@marcofranssen do you concur? If so, thats fine, we'll finish it in the db pr.

kfox1111:
/me sighs... two buttons right next to each other that do different things....

PR #218, opened 2023-04-14
Move reusable macros to root chart

kfox1111:
This patch adds the start of a library to help reduce code duplication. It will be split out to its own library chart at the same time the other charts are split out.


marcofranssen:
In general subcharts can reuse the functions from the root chart. So no need for another chart. This only makes sense if we move to multiple root level charts to have a lib chart. The lib chart would then have to be a dependency of each individual chart.

To keep it simple I would postpone this refactor to a later stage. Or just define the _helpers in the root chart for now.

charts/spire/charts/spire-server/templates/bundle-configmap.yaml
```diff
@@ -2,5 +2,5 @@
 apiVersion: v1
 kind: ConfigMap
 metadata:
-  name: {{ include "spire-server.bundle-configmap" . }}
+  name: {{ include "spire-lib.bundle-configmap" . }}
```
marcofranssen, line 5: This one was a really spire-server specific one I think.

marcofranssen (approved):
LGTM 🚀

PR #225, opened 2023-04-19
External database configuration

kfox1111:
This patch makes all the database settings configurable.

fixes: https://github.com/spiffe/helm-charts/issues/37

### DIFF
--- a/.github/tests/charts.json
+++ b/.github/tests/charts.json
@@ -13,5 +13,15 @@
     "name": "ingress-nginx",
     "repo": "https://kubernetes.github.io/ingress-nginx",
     "version": "4.6.0"
+  },
+  {
+    "name": "mysql",
+    "repo": "https://charts.bitnami.com/bitnami",
+    "version": "9.7.2"
+  },
+  {
+    "name": "postgresql",
+    "repo": "https://charts.bitnami.com/bitnami",
+    "version": "12.2.2"
   }
 ]
--- a/.github/tests/production-external-mysql/install.sh
+++ b/.github/tests/production-external-mysql/install.sh
@@ -1,0 +1,41 @@
+#!/usr/bin/env bash
+
+set -xe
+
+SCRIPT=$(readlink -f "$0")
+SCRIPTPATH=$(dirname "$SCRIPT")
+
+DB=spire
+DBUSER=spire
+DBPW=$(uuidgen)
+DBROOTPW=$(uuidgen)
+
+# Generate random settings to make sure things come up with random settings.
+cat <<EOF > /tmp/$$-db-values.yaml
+auth:
+  database: ${DB}
+  username: ${DBUSER}
+  password: ${DBPW}
+  rootPassword: ${DBROOTPW}
+EOF
+
+cat <<EOF > /tmp/$$-spire-values.yaml
+spire-server:
+  dataStore:
+    sql:
+      plugin_data:
+        database_type: "mysql"
+        connection_string: "${DBUSER}:${DBPW}@tcp(mysql:3306)/${DB}?parseTime=true"
+EOF
+
+helm install mysql mysql --namespace "spire-server" --version "$VERSION_MYSQL" --repo "$HELM_REPO_MYSQL" \
+  --values "${SCRIPTPATH}/mysql-values.yaml" \
+  --values /tmp/$$-db-values.yaml --wait
+
+helm install \
+  --namespace "spire-server" \
+  --values /tmp/$$-spire-values.yaml \
+  --values "${SCRIPTPATH}/../../../examples/production/values.yaml" \
+  spire charts/spire --wait
+
+helm test spire --namespace "spire-server"
--- a/.github/tests/production-external-mysql/install.sh
+++ b/.github/tests/production-external-mysql/install.sh
@@ -23,9 +23,14 @@ cat <<EOF > /tmp/$$-spire-values.yaml
 spire-server:
   dataStore:
     sql:
-      plugin_data:
-        database_type: "mysql"
-        connection_string: "${DBUSER}:${DBPW}@tcp(mysql:3306)/${DB}?parseTime=true"
+      databaseType: mysql
+      databaseName: ${DB}
+      username: ${DBUSER}
+      password: ${DBPW}
+      host: mysql
+      port: 3306
+      options:
+      - parseTime: true
 EOF
 
 helm install mysql mysql --namespace "spire-server" --version "$VERSION_MYSQL" --repo "$HELM_REPO_MYSQL" \
--- a/charts/spire/charts/spire-server/README.md
+++ b/charts/spire/charts/spire-server/README.md
@@ -59,7 +59,7 @@ A Helm chart to install the SPIRE server.
 | dataStore.sql.options | list | `[]` | Only used by "postgres" or "mysql" |
 | dataStore.sql.password | string | `""` |  |
 | dataStore.sql.plugin_data | object | `{}` | Settings from https://github.com/spiffe/spire/blob/main/doc/plugin_server_datastore_sql.md go in this section |
-| dataStore.sql.port | int | `0` | Likely 5432 for postgres and 3306 for mysql |
+| dataStore.sql.port | int | `0` | If 0 (default), it will auto set to 5432 for postgres and 3306 for mysql |
 | dataStore.sql.username | string | `"spire"` |  |
 | defaultJwtSvidTTL | string | `"1h"` |  |
 | defaultX509SvidTTL | string | `"4h"` |  |
--- a/charts/spire/charts/spire-server/README.md
+++ b/charts/spire/charts/spire-server/README.md
@@ -53,13 +53,13 @@ A Helm chart to install the SPIRE server.
 | controllerManager.validatingWebhookConfiguration.upgradeHook.image.registry | string | `"cgr.dev"` |  |
 | controllerManager.validatingWebhookConfiguration.upgradeHook.image.repository | string | `"chainguard/kubectl"` |  |
 | controllerManager.validatingWebhookConfiguration.upgradeHook.image.version | string | `"latest"` |  |
-| dataStore.sql.database | string | `"spire"` |  |
-| dataStore.sql.database_type | string | `"sqlite3"` | Other supported databases are "postgres" and "mysql" |
+| dataStore.sql.databaseName | string | `"spire"` |  |
+| dataStore.sql.databaseType | string | `"sqlite3"` | Other supported databases are "postgres" and "mysql" |
 | dataStore.sql.host | string | `""` |  |
 | dataStore.sql.options | list | `[]` | Only used by "postgres" or "mysql" |
 | dataStore.sql.password | string | `""` |  |
-| dataStore.sql.plugin_data | object | `{}` | Set any setting from https://github.com/spiffe/spire/blob/v1.6.3/doc/plugin_server_datastore_sql.md under this section. |
-| dataStore.sql.port | int | `0` | likely 5432 for postgresql 3306 for mysql |
+| dataStore.sql.plugin_data | object | `{}` | Settings from https://github.com/spiffe/spire/blob/main/doc/plugin_server_datastore_sql.md go in this section |
+| dataStore.sql.port | int | `0` | Likely 5432 for postgres and 3306 for mysql |
 | dataStore.sql.username | string | `"spire"` |  |
 | defaultJwtSvidTTL | string | `"1h"` |  |
 | defaultX509SvidTTL | string | `"4h"` |  |
--- a/charts/spire/charts/spire-server/README.md
+++ b/charts/spire/charts/spire-server/README.md
@@ -53,14 +53,14 @@ A Helm chart to install the SPIRE server.
 | controllerManager.validatingWebhookConfiguration.upgradeHook.image.registry | string | `"cgr.dev"` |  |
 | controllerManager.validatingWebhookConfiguration.upgradeHook.image.repository | string | `"chainguard/kubectl"` |  |
 | controllerManager.validatingWebhookConfiguration.upgradeHook.image.version | string | `"latest"` |  |
-| dataStore.sql.databaseName | string | `"spire"` |  |
+| dataStore.sql.databaseName | string | `"spire"` | Only used by "postgres" or "mysql" |
 | dataStore.sql.databaseType | string | `"sqlite3"` | Other supported databases are "postgres" and "mysql" |
-| dataStore.sql.host | string | `""` |  |
+| dataStore.sql.host | string | `""` | Only used by "postgres" or "mysql" |
 | dataStore.sql.options | list | `[]` | Only used by "postgres" or "mysql" |
-| dataStore.sql.password | string | `""` |  |
+| dataStore.sql.password | string | `""` | Only used by "postgres" or "mysql" |
 | dataStore.sql.plugin_data | object | `{}` | Settings from https://github.com/spiffe/spire/blob/main/doc/plugin_server_datastore_sql.md go in this section |
-| dataStore.sql.port | int | `0` | If 0 (default), it will auto set to 5432 for postgres and 3306 for mysql |
-| dataStore.sql.username | string | `"spire"` |  |
+| dataStore.sql.port | int | `0` | If 0 (default), it will auto set to 5432 for postgres and 3306 for mysql. Only used by those databases. |
+| dataStore.sql.username | string | `"spire"` | Only used by "postgres" or "mysql" |
 | defaultJwtSvidTTL | string | `"1h"` |  |
 | defaultX509SvidTTL | string | `"4h"` |  |
 | extraContainers | list | `[]` |  |
--- a/charts/spire/charts/spire-server/README.md
+++ b/charts/spire/charts/spire-server/README.md
@@ -53,10 +53,14 @@ A Helm chart to install the SPIRE server.
 | controllerManager.validatingWebhookConfiguration.upgradeHook.image.registry | string | `"cgr.dev"` |  |
 | controllerManager.validatingWebhookConfiguration.upgradeHook.image.repository | string | `"chainguard/kubectl"` |  |
 | controllerManager.validatingWebhookConfiguration.upgradeHook.image.version | string | `"latest"` |  |
-| dataStorage.accessMode | string | `"ReadWriteOnce"` |  |
-| dataStorage.enabled | bool | `true` |  |
-| dataStorage.size | string | `"1Gi"` |  |
-| dataStorage.storageClass | string | `nil` |  |
+| dataStore.sql.database | string | `"spire"` |  |
+| dataStore.sql.database_type | string | `"sqlite3"` | Other supported databases are "postgres" and "mysql" |
+| dataStore.sql.host | string | `""` |  |
+| dataStore.sql.options | list | `[]` | Only used by "postgres" or "mysql" |
+| dataStore.sql.password | string | `""` |  |
+| dataStore.sql.plugin_data | object | `{}` | Set any setting from https://github.com/spiffe/spire/blob/v1.6.3/doc/plugin_server_datastore_sql.md under this section. |
+| dataStore.sql.port | int | `0` | likely 5432 for postgresql 3306 for mysql |
+| dataStore.sql.username | string | `"spire"` |  |
 | defaultJwtSvidTTL | string | `"1h"` |  |
 | defaultX509SvidTTL | string | `"4h"` |  |
 | extraContainers | list | `[]` |  |
@@ -80,6 +84,10 @@ A Helm chart to install the SPIRE server.
 | nodeAttestor.k8sPsat.serviceAccountAllowList | list | `[]` |  |
 | nodeSelector | object | `{}` |  |
 | notifier.k8sbundle.namespace | string | `""` | Namespace to push the bundle into, if blank will default to SPIRE Server namespace |
+| persistence.accessMode | string | `"ReadWriteOnce"` |  |
+| persistence.enabled | bool | `true` | Enable persistence used by sqlite3 for spire-server, by the disk KeyStore, and/or by Tornjak. |
+| persistence.size | string | `"1Gi"` |  |
+| persistence.storageClass | string | `nil` |  |
 | podAnnotations | object | `{}` |  |
 | podSecurityContext | object | `{}` |  |
 | replicaCount | int | `1` | SPIRE server currently runs with a sqlite database. Scaling to multiple instances will not work until we use an external database. |
--- a/charts/spire/charts/spire-server/templates/_helpers.tpl
+++ b/charts/spire/charts/spire-server/templates/_helpers.tpl
@@ -105,3 +105,49 @@ Create the name of the service account to use
 {{- end }}
 {{- end }}
 
+{{- define "spire-server.config-mysql-query" }}
+{{- $lst := list }}
+{{- range . }}
+{{- range $key, $value := . }}
+{{- $eValue := toString $value }}
+{{- $entry := printf "%s=%s" (urlquery $key) (urlquery $eValue) }}
+{{- $lst = append $lst $entry }}
+{{- end }}
+{{- end }}
+{{- if gt (len $lst) 0 }}
+{{- printf "?%s" (join "&" $lst) }}
+{{- end }}
+{{- end }}
+
+{{- define "spire-server.config-postgresql-options" }}
+{{- $lst := list }}
+{{- range . }}
+{{- range $key, $value := . }}
+{{- $eValue := toString $value }}
+{{- $entry := printf "%s=%s" $key $eValue }}
+{{- $lst = append $lst $entry }}
+{{- end }}
+{{- end }}
+{{- if gt (len $lst) 0 }}
+{{- printf " %s" (join " " $lst) }}
+{{- end }}
+{{- end }}
+
+{{- define "spire-server.datastore-config" }}
+{{- $config := deepCopy .Values.dataStore.sql.plugin_data }}
+{{- if eq .Values.dataStore.sql.database_type "sqlite3" }}
+  {{- $_ := set $config "database_type" "sqlite3" }}
+  {{- $_ := set $config "connection_string" "/run/spire/data/datastore.sqlite3" }}
+{{- else if eq .Values.dataStore.sql.database_type "mysql" }}
+  {{- $_ := set $config "database_type" "mysql" }}
+  {{- $query := include "spire-server.config-mysql-query" .Values.dataStore.sql.options }}
+  {{- $_ := set $config "connection_string" (printf "%s:${DBPW}@tcp(%s:%d)/%s%s" .Values.dataStore.sql.username .Values.dataStore.sql.host (int .Values.dataStore.sql.port) .Values.dataStore.sql.database $query) }}
+{{- else if eq .Values.dataStore.sql.database_type "postgresql" }}
+  {{- $_ := set $config "database_type" "postgresql" }}
+  {{- $options:= include "spire-server.config-postgresql-options" .Values.dataStore.sql.options }}
+  {{- $_ := set $config "connection_string" (printf "dbname=%s user=%s password=${DBPW} host=%s port=%d%s" .Values.dataStore.sql.database .Values.dataStore.sql.username .Values.dataStore.sql.host (int .Values.dataStore.sql.port) $options) }}
+{{- else }}
+  {{- fail "Unsupported database type" }}
+{{- end }}
+{{- $config | toYaml }}
+{{- end }}
--- a/charts/spire/charts/spire-server/templates/_helpers.tpl
+++ b/charts/spire/charts/spire-server/templates/_helpers.tpl
@@ -135,17 +135,19 @@ Create the name of the service account to use
 
 {{- define "spire-server.datastore-config" }}
 {{- $config := deepCopy .Values.dataStore.sql.plugin_data }}
-{{- if eq .Values.dataStore.sql.database_type "sqlite3" }}
+{{- if eq .Values.dataStore.sql.databaseType "sqlite3" }}
   {{- $_ := set $config "database_type" "sqlite3" }}
   {{- $_ := set $config "connection_string" "/run/spire/data/datastore.sqlite3" }}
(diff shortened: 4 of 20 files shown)


charts/spire/charts/spire-server/values.yaml
```diff
@@ -217,3 +223,11 @@ nodeAttestor:
...
     enabled: true
     serviceAccountAllowList: []
+
+# FIXME Not implemented yet
+postgresql:
+  enabled: false
+
+# FIXME Not implemented yet
+mysql:
+  enabled: false
```
faisal-memon, line 233: I think this should probably go under the persistence of datastore sections

kfox1111: Thanks for the review.

Unfortunately, a nested chart can't contain any nestings in the values file of the chart its nested into.
so can be anything at the root of the chart, such as:
```
postgresql:
  enabled: true
```

but cant be like:
```yaml
dataStore:
  sql:
     postgresql:
       enabled: true
```

Its a limitation of helm itself.

kfox1111:
Split out the internal db support to its own pr: https://github.com/spiffe/helm-charts/pull/233

Should be ready for review.

charts/spire/charts/spire-server/templates/secret.yaml
```diff
@@ -101,10 +100,9 @@ telemetry:
 {{- end }}
 {{- end }}
 apiVersion: v1
-kind: ConfigMap
+kind: Secret
```
marcofranssen, line 103: Why are we switching this to a secret?

Can't we just put the secrets in a secret.

E.g.:

We could store the db creds in a secret and load them into the respective env var, this way we don't need to change the configmap:

See
http://www.postgresqlnow.com/connect-env-variables/

charts/spire/charts/spire-server/values.yaml
```diff
@@ -73,12 +73,22 @@ affinity: {}
...
   storageClass: null
 
+dataStore:
+  sql:
+    # -- Set any setting from https://github.com/spiffe/spire/blob/v1.6.3/doc/plugin_server_datastore_sql.md under this section.
+    plugin_data:
+      # -- Set to either "postgres" or "mysql" if you want to use an external databse
+      database_type: ""
+      # -- Set to the appropriate setting for your chosen external database. Passwords will be protected
+      connection_string: ""
```
marcofranssen, line 90: For documentation purposes and less hidden magic it might be better to default this to sqlite.

```suggestion
dataStore:
  sql:
    # -- Set any setting from https://github.com/spiffe/spire/blob/v1.6.3/doc/plugin_server_datastore_sql.md under this section.
    plugin_data:
      # -- Other supported databases are "postgres" and "mysql"
      database_type: sqlite3
      # -- The connectionstring to connect to your database instance.
      connection_string: /run/spire/data/datastore.sqlite3
```

examples/external-mysql/values.yaml
```diff
@@ -0,0 +1,6 @@
+spire-server:
+  dataStore:
+    sql:
+      plugin_data:
+        database_type: "mysql"
+        connection_string: "spire:CHANGEME@tcp(mysql:3306)/spire?parseTime=true"
```
marcofranssen, line 6:
```suggestion
        host: mysql
        port: 3306
        database: spire
        username: spire
        password: CHANGEME
        options:
          - parseTime: true
```

Too support the confimap combination with secrets we might want to more stronglytype the values file.

examples/external-postgresql/values.yaml
```diff
@@ -0,0 +1,7 @@
+spire-server:
+  dataStore:
+    sql:
+      plugin_data:
+        database_type: "postgres"
+        connection_string: "dbname=YOURDB user=USER password=CHANGEME host=postgresql port=5432 sslmode=disable"
```
marcofranssen, line 6:
```suggestion
        host: postgresql
        port: 5432
        database: spire
        username: spire
        password: CHANGEME
        options: 
          - sslmode: disable
```

Too support the confimap combination with secrets we might want to more strongly type the values file.

marcofranssen (changes requested):
Couple of suggestions to follow patterns used often in other charts I have seen in the past.

charts/spire/charts/spire-server/values.yaml
kfox1111: The logic here was developed with the idea that internal databases as per https://github.com/spiffe/helm-charts/pull/233 could be a thing. The suggested settings somewhat conflict with that.

examples/external-mysql/values.yaml
kfox1111: What happens when they add a new database type that isn't sqllite or a sql server? say, mongo or some other nosql database? Are we backing ourselves into a corner trying to interpret connection_string?

If we're adding config options I don't think they should be under plugin_data, which should match closely what is in spire's config. We probably should move it up a level to under sql: if we do that.

charts/spire/charts/spire-server/templates/secret.yaml
marcofranssen: To some extend it does protect yes, in the end k8s secrets are just obfuscated as base64.

In many cases we pull secrets from external sources injected as env variables, having the entire config as secret complicates this strategy.

examples/external-mysql/values.yaml
marcofranssen: In that case we probably have to add another implementation that builds the connection string based on that database type

charts/spire/charts/spire-server/templates/secret.yaml
kfox1111: Users of k8s often confuse encoding and encryption or assume because one is used it means its designed to be obscured in favor over encryption. This is **not** the case in Kubernetes. base64 was chosen to encode secret data as its often binary data and needs to be put into a yaml/json document which isn't binary safe. But encryption is supported too. In many production systems, secrets are encrypted via:
https://kubernetes.io/docs/tasks/administer-cluster/encrypt-data/

The NSA's recommendation for Kubernetes hardening mentions it explicitly. I believe the DOD and CIS benchmark recommendations are also to turn on encryption for secrets.

My own clusters are encrypted. Secrets are often encrypted while configmaps are often not.

---

As for arbitrary plugin support, that's a debate on governance. I'm on the side of, if the spire folks needed plugin support to support the users, I'm for supporting the users in their needs as well. We'll hash that problem out in the governance meeting.

examples/external-mysql/values.yaml
kfox1111: I guess in the case of internal databases, we can just use enable to flag if we totally ignore the config there. I'll take a stab at it.

.github/tests/charts.json
```diff
@@ -13,5 +13,17 @@
     "name": "ingress-nginx",
     "repo": "https://kubernetes.github.io/ingress-nginx",
     "version": "4.6.0"
+  },
+  {
+
```
marcofranssen, line 18:
```suggestion
```

.github/tests/charts.json
```diff
@@ -13,5 +13,17 @@
...
+    "name": "mysql",
+    "repo": "https://charts.bitnami.com/bitnami",
+    "version": "9.7.2"
+  },
+  {
+
```
marcofranssen, line 24:
```suggestion
```

.github/tests/production-external-mysql/install.sh
```diff
@@ -0,0 +1,41 @@
...
+# Generate random settings to make sure things come up with random settings.
+cat > /tmp/$$-db-values.yaml <<EOF
+auth:
+  database: ${DB}
+  username: ${DBUSER}
+  password: ${DBPW}
+  rootPassword: ${DBROOTPW}
+EOF
+
+cat > /tmp/$$-spire-values.yaml <<EOF
```
marcofranssen, line 22:
```suggestion
cat <<EOF > /tmp/$$-spire-values.yaml
```

examples/external-postgresql/values.yaml
```diff
@@ -0,0 +1,12 @@
...
+    sql:
+      database_type: postgresql
+      host: postgresql
+      port: 5432
+      database: spire
+      username: spire
+      password: CHANGEME
+      options:
+        - sslmode: disable
+
```
marcofranssen, line 12:
```suggestion
```

charts/spire/charts/spire-server/README.md
```diff
@@ -80,6 +84,10 @@ A Helm chart to install the SPIRE server.
 | nodeAttestor.k8sPsat.serviceAccountAllowList | list | `[]` |  |
 | nodeSelector | object | `{}` |  |
 | notifier.k8sbundle.namespace | string | `""` | Namespace to push the bundle into, if blank will default to SPIRE Server namespace |
+| persistence.accessMode | string | `"ReadWriteOnce"` |  |
+| persistence.enabled | bool | `true` |  |
+| persistence.size | string | `"1Gi"` |  |
+| persistence.storageClass | string | `nil` |  |
```
faisal-memon, line 90: What do you think of keep this together with the rest of the datastore config?

charts/spire/charts/spire-server/values.yaml
```diff
@@ -73,12 +73,28 @@ affinity: {}
...
+  sql:
+    # -- Other supported databases are "postgres" and "mysql"
+    database_type: sqlite3
+    database: spire
+    host: ""
+    # -- likely 5432 for postgresql 3306 for mysql
+    port: 0
+    username: spire
+    password: ""
+    options: []
```
faisal-memon, line 93: Maybe comment ofer this section that this config is only needed for postgres or mysql.

charts/spire/charts/spire-server/values.yaml
```diff
@@ -73,12 +73,28 @@ affinity: {}
...
+dataStore:
+  sql:
+    # -- Other supported databases are "postgres" and "mysql"
+    database_type: sqlite3
+    database: spire
+    host: ""
+    # -- likely 5432 for postgresql 3306 for mysql
+    port: 0
+    username: spire
+    password: ""
```
faisal-memon, line 92: Should the password be read from the secret?

charts/spire/charts/spire-server/templates/secret.yaml
faisal-memon: Prefer to only have the password be a k8s secret

kfox1111: Ok. I'll prepare another pr with that in it so we can review.

Struggling a bit with how all this reconciles with the principal of keeping things "simple". All this combined is the most complicated way of setting "connection_string" possible.

charts/spire/charts/spire-server/README.md
kfox1111: It doesn't have a 1:1 relationship with dataStore. 

Its potentially used by sqlite3 for spire-server, by the disk KeyStore, and/or by Tornjak.

So when mysql/postgresql is used that has its own storage, it still may be needed.

charts/spire/charts/spire-server/templates/secret.yaml
kfox1111: Here it is. https://github.com/spiffe/helm-charts/pull/255

charts/spire/charts/spire-server/values.yaml
kfox1111: Are you asking for the option of reading the password out of an existing k8s secret?

charts/spire/charts/spire-server/README.md
kfox1111: See https://github.com/spiffe/helm-charts/pull/211
As its not merged, and this kind of depends on it, some of the changes appear here too.

charts/spire/charts/spire-server/values.yaml
kfox1111: Done

charts/spire/charts/spire-server/README.md
```diff
@@ -80,6 +84,7 @@ A Helm chart to install the SPIRE server.
 | nodeAttestor.k8sPsat.serviceAccountAllowList | list | `[]` |  |
 | nodeSelector | object | `{}` |  |
 | notifier.k8sbundle.namespace | string | `""` | Namespace to push the bundle into, if blank will default to SPIRE Server namespace |
+| persistence | object | `{"accessMode":"ReadWriteOnce","enabled":true,"size":"1Gi","storageClass":null}` | Enable persistence used by sqlite3 for spire-server, by the disk KeyStore, and/or by Tornjak. |
```
faisal-memon, line 87: Why did this get all condensed?

kfox1111:
Somehow that merge still made a dco issue. :/ Going to manually squash...

charts/spire/charts/spire-server/README.md
kfox1111: Merge issue. Should be fixed.

examples/external-mysql/values.yaml
```diff
@@ -0,0 +1,11 @@
+spire-server:
+  dataStore:
+    sql:
+      database_type: mysql
+      host: mysql
+      port: 3306
+      database: spire
+      username: spire
+      password: CHANGEME
```
marcofranssen, line 9: Maybe add a disclaimer, folks probably want to inject this differently so it doesn't end up in git repo. To also mitigate @faisal-memon his question.

marcofranssen (changes requested):
Few small nits.

charts/spire/charts/spire-server/README.md
marcofranssen: This gets condensed not because of merge issue, but because of the helm docs introduced.

See https://github.com/norwoodj/helm-docs#valuesyaml-metadata for more details.

In essence adding a documentation tag on any of the nested properties would resolve this again.

marcofranssen:
> Somehow that merge still made a dco issue. :/ Going to manually squash...

Thats because merges don't automatically add the signof. It is a missing feature in github for merging prs into prs. You could have added the dco during the merge dialog by typing it yourself. However, the squash is also fine in this case 👍

charts/spire/charts/spire-server/README.md
kfox1111: No, there is that other persistence pr, and I manually merged part of it here to get them looking the same again, but put it in the wrong place. :/

I fixed the manual merge issue location and it should be back to split out again.

examples/external-mysql/values.yaml
kfox1111: oh, great point... It should be removed here and documentation updated.... Will fix.

kfox1111: Fix committed.

kfox1111:
> Thats because merges don't automatically add the signof. It is a missing feature in github for merging prs into prs. You could have added the dco during the merge dialog by typing it yourself. However, the squash is also fine in this case +1

I could have sworn I reviewed that and verified my signoff was in it. but somehow it still failed. But, yeah. fixed for now.

charts/spire/charts/spire-server/README.md
```diff
@@ -80,6 +84,10 @@ A Helm chart to install the SPIRE server.
 | nodeAttestor.k8sPsat.serviceAccountAllowList | list | `[]` |  |
 | nodeSelector | object | `{}` |  |
 | notifier.k8sbundle.namespace | string | `""` | Namespace to push the bundle into, if blank will default to SPIRE Server namespace |
+| persistence.accessMode | string | `"ReadWriteOnce"` |  |
+| persistence.enabled | bool | `true` | Enable persistence used by sqlite3 for spire-server, by the disk KeyStore, and/or by Tornjak. |
```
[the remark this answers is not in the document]
kfox1111: And a few advantages. less volumes for admins to have to snapshot/backup.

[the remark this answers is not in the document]
faisal-memon: I agree with @kfox1111 on having 1 volume. Keeps things simple. If we need to have multiple volumes we can add in the future.

charts/spire/charts/spire-server/templates/_helpers.tpl
```diff
@@ -105,3 +105,49 @@ Create the name of the service account to use
 {{- end }}
 {{- end }}
 
+{{- define "spire-server.config-mysql-query" }}
```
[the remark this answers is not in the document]
kfox1111: So, 
{{- define "spire-server.config-mysql-options" }}
?

The general idea I was thinking, was to have the reader of the code pay attention a little bit to the difference between the two. the function is converting the options into a url query in the one case, but a key=value (more straight through options case?) when it comes to postgresql.

Another option might be to spell that out more explicitly...
"spire-server.config-mysql-options-2-url-query"
"spire-server.config-postgresql-options-2-key-value" ?

charts/spire/charts/spire-server/templates/_helpers.tpl
```diff
@@ -105,3 +105,49 @@ Create the name of the service account to use
...
+{{- $lst := list }}
+{{- range . }}
+{{- range $key, $value := . }}
+{{- $eValue := toString $value }}
+{{- $entry := printf "%s=%s" (urlquery $key) (urlquery $eValue) }}
```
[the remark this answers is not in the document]
kfox1111: If it was just the = it would be a lot easier. the = and the urlquery makes it much trickier. Not sure how to reduce it without being very similar in line count (maybe slightly less) but with more convolution. The printf join line is a bit different too.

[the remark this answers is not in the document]
faisal-memon: got it, no need to change anything. thanks for the explanation.

charts/spire/charts/spire-server/README.md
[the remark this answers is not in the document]
faisal-memon: Can we auto set the `persistence` if `sqllite3` is the db or `keymanager` is `disk`?

charts/spire/charts/spire-server/values.yaml
```diff
@@ -73,12 +73,29 @@ affinity: {}
...
+  # -- Enable persistence used by sqlite3 for spire-server, by the disk KeyStore, and/or by Tornjak.
   enabled: true
   size: 1Gi
   accessMode: ReadWriteOnce
   storageClass: null
 
+dataStore:
+  sql:
+    # -- Other supported databases are "postgres" and "mysql"
+    database_type: sqlite3
```
faisal-memon, line 86:
```suggestion
    databaseType: sqlite3
```

charts/spire/charts/spire-server/values.yaml
```diff
@@ -73,12 +73,29 @@ affinity: {}
...
+    database: spire
```
faisal-memon, line 87:
```suggestion
    databaseName: spire
```

charts/spire/charts/spire-server/templates/_helpers.tpl
```diff
@@ -105,3 +105,49 @@ Create the name of the service account to use
...
+  {{- $_ := set $config "database_type" "sqlite3" }}
+  {{- $_ := set $config "connection_string" "/run/spire/data/datastore.sqlite3" }}
+{{- else if eq .Values.dataStore.sql.database_type "mysql" }}
+  {{- $_ := set $config "database_type" "mysql" }}
+  {{- $query := include "spire-server.config-mysql-query" .Values.dataStore.sql.options }}
+  {{- $_ := set $config "connection_string" (printf "%s:${DBPW}@tcp(%s:%d)/%s%s" .Values.dataStore.sql.username .Values.dataStore.sql.host (int .Values.dataStore.sql.port) .Values.dataStore.sql.database $query) }}
+{{- else if eq .Values.dataStore.sql.database_type "postgresql" }}
+  {{- $_ := set $config "database_type" "postgresql" }}
+  {{- $options:= include "spire-server.config-postgresql-options" .Values.dataStore.sql.options }}
+  {{- $_ := set $config "connection_string" (printf "dbname=%s user=%s password=${DBPW} host=%s port=%d%s" .Values.dataStore.sql.database .Values.dataStore.sql.username .Values.dataStore.sql.host (int .Values.dataStore.sql.port) $options) }}
```
faisal-memon, line 148: Will the rendered configMap have the password visible in it?

faisal-memon: nm i see it now, it pulls from the env var.

charts/spire/charts/spire-server/values.yaml
```diff
@@ -73,12 +73,29 @@ affinity: {}
...
+    host: ""
+    # -- likely 5432 for postgresql 3306 for mysql
```
faisal-memon, line 89:
```suggestion
    # -- Likely 5432 for postgres and 3306 for mysql
```

charts/spire/charts/spire-server/values.yaml
faisal-memon, line 86:
```suggestion
    database_type: sqlite3
    
    # The remaining config in this section only applies to postgres and mysql databases

```

charts/spire/charts/spire-server/values.yaml
```diff
@@ -73,12 +73,29 @@ affinity: {}
...
+    port: 0
+    username: spire
+    password: ""
+    # -- Only used by "postgres" or "mysql"
+    options: []
```
faisal-memon, line 94:
```suggestion
    options: []
```

charts/spire/charts/spire-server/values.yaml
```diff
@@ -73,12 +73,29 @@ affinity: {}
...
+
+    # -- Set any setting from https://github.com/spiffe/spire/blob/v1.6.3/doc/plugin_server_datastore_sql.md under this section.
```
faisal-memon, line 96:
```suggestion
    # -- Set any setting from https://github.com/spiffe/spire/blob/main/doc/plugin_server_datastore_sql.md under this section.
```

faisal-memon: So we don't have to keep the version up to date.

charts/spire/charts/spire-server/values.yaml
kfox1111: I can't find the comment now. probably lost in the squash, but I believe the comment there was requested by Marco.

kfox1111: oh... I see what you did... this with the other patch puts a comment on the whole section instead...

But that isn't quite equivalent, because, while you can see the comment when looking a the raw values, it doesn't show up in the README. Not sure how to do the same thing in a way that would show up in the README

kfox1111:
Added port auto setting, so if the user is using the default db ports, it will "just work"

charts/spire/charts/spire-server/values.yaml
faisal-memon: Yea, not sure if thats possible with helm docs. Can alternatively add the only used by to each of the variable if you want it to show up in the readme.

charts/spire/charts/spire-server/README.md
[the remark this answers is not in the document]
faisal-memon: Whats the value of disabling persistence?

charts/spire/charts/spire-server/templates/_helpers.tpl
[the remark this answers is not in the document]
faisal-memon: Its fine with me to keep as is. seems like its doing something different.

charts/spire/charts/spire-server/README.md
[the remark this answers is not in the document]
kfox1111: I've used it in several, related ways on other charts (there may be others)

1. In ci pipelines to speed up deployment of testing your own software, that uses spire. Provisioning the volume takes a bit to do, and being able to skip it reduces the overall runtime of the pipeline.  This enabled more rapid development/testing cycles.
2. In a ci pipeline where your deploying a test k8s. Similar to 1, not needing to provision a persistent volume driver at all can speed up deployment of the test cluster. Most valuable when dynamically allocating clusters in the pipeline
3. Some ci systems don't clean up after themselves very well (EX: cancellations). Its much easier to clean up helm charts that don't have pvc's.
4. Complex testing: while minikube has pv support in standalone node mode, it does not in its multinode mode. So if you want to do some nice multinode testing with it, just disabling persistence lets you deploy test instances of things such as spire, mysql, your code, etc in a nice scailable way, with stuff like antiaffinity, for quick testing of something that looks production like.

examples/bin/readpw.sh
```diff
@@ -0,0 +1,3 @@
+# shellcheck shell=bash disable=SC2034
```
marcofranssen, line 1:
```suggestion
#!/usr/bin/env bash

# shellcheck shell=bash disable=SC2034
```

marcofranssen (commented):
Few more nitpicks, disclaimer, reviewed from the mobile app, hopefully suggestions without typos.

examples/bin/readpw.sh
kfox1111: I did not put a shell on it because its not really executable. It can only be sourced. If you execute it, it runs, but the env var gets undefined at the end.

charts/spire/charts/spire-server/README.md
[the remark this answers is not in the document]
marcofranssen: Due the way these PVCs work any other workloads then spire-server will already define their own `persistentVolumeClaim`

See spire-server statefulset.

```yaml
  volumeClaimTemplates:
    {{- if eq (.Values.persistence.enabled | toString) "true" }}
    - metadata:
        name: spire-data
      spec:
        accessModes:
          - {{ .Values.persistence.accessMode | default "ReadWriteOnce" }}
        resources:
          requests:
            storage: {{ .Values.persistence.size }}
        {{- if .Values.persistence.storageClass }}
        storageClassName: {{ .Values.persistence.storageClass }}
        {{- end }}
    {{ end }}
```

If tornjak needs a PVC it will have to implement its own meaning you would also want to be able to choose its own storageClass and accessModes etc. That is just how it works. Going with this API simply means we will have to break the API as soon we add a PVC to tornjak. IMHO the persistence flag is to generalized, hence why it previously was at dataStorage specifically for the dataStorage volume.

maybe it should be 

```
dataStore:
  persistence:
    enabled: true
    …
```

So we can later do things like

```
tornjak:
  persistence:
    enabled: true
```

Oh and if we are about to disable persistence, then probably the statefulset should also be changed to a deployment if persistence is disabled. That is something I have seen in other charts as well. Happy to create a PR to patch those changes on top of this PR.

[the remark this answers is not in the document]
kfox1111: > maybe it should be
> ```yaml
> dataStore:
>  persistence:
>    enabled: true
>    …
> ```
> So we can later do things like
> 
> ```yaml
> tornjak:
>  persistence:
>    enabled: true
> ```
and one for the disk keystore too. But, yeah, there'd be 3x the storage waste I think. I don't see the benefit in having so many volumes, especially once you have multiple replicas. But there is benefit in combining them for the end user.

> Oh and if we are about to disable persistence, then probably the statefulset should also be changed to a deployment if persistence is disabled. That is something I have seen in other charts as well. Happy to create a PR to patch those changes on top of this PR.

Other then for testing purposes, I don't think the user should ever be disabling persistence on the disk keystore. So at least 1 pvc is needed. Having the test case deployment look like the prod deployment in every way but deploy a little faster and not save state permanently is the ideal, so it to still be a statefulset is useful. Its totally ok for it to remain a statefulset and have no volumes. There are other properties of a statefulset that are useful outside pvs, such as persistent pod identity and lifecycle management that still make it a thing. I do volumeless statefulsets all the time.

> Imagine following usecase. I do want to run with an external database, that means I want to disable the persistence for sqlite, but I still want to run persistence for the keys and tornjak. How would I achieve that with this change?

The same way as today. You just don't turn off persistence. Nothing related to this has changed in this pr. Just the name of the key used to do the enabling/disabling as been renamed to better reflect the way its always worked, and to follow helm convention.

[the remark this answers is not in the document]
kfox1111: Thought of one other thing, and I think its really important....

We can't add more volumes to the statefulset without a lot of upgrade pain on the users. Apparently volumeClaimTemplates is immutable. So if the user wanted to start out without, say, Tornjak, and then enable it later they would have to destroy their statefulset if it had sepaerate pvc's. 
```
The StatefulSet "test" is invalid: spec: Forbidden: updates to statefulset spec for fields other than
'replicas','template', 'updateStrategy', 'persistentVolumeClaimRetentionPolicy' and
'minReadySeconds' are forbidden
```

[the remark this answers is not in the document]
faisal-memon: We can still use a single pvc as long as they all store persistent data to `/run/spire/data` correct?  Thats what we are doing currently for sqllite3 and the disk keymanager. 

Im not familiar with Tornjak. If Tornjak is in same Pod then it would just need to do the same.

> Oh and if we are about to disable persistence, then probably the statefulset should also be changed to a deployment if persistence is disabled. That is something I have seen in other charts as well. Happy to create a PR to patch those changes on top of this PR.

I think this is a good idea to merge after this PR.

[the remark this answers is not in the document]
marcofranssen: > Thought of one other thing, and I think its really important....
> 
> We can't add more volumes to the statefulset without a lot of upgrade pain on the users. Apparently volumeClaimTemplates is immutable. So if the user wanted to start out without, say, Tornjak, and then enable it later they would have to destroy their statefulset if it had sepaerate pvc's.
> 
> ```
> The StatefulSet "test" is invalid: spec: Forbidden: updates to statefulset spec for fields other than
> 'replicas','template', 'updateStrategy', 'persistentVolumeClaimRetentionPolicy' and
> 'minReadySeconds' are forbidden
> ```

Fair point. Maybe we should switch to using a deployment with a separate PVC like they do in this chart. Probably for the reason you mentioning here? 
https://github.com/sigstore/helm-charts/blob/main/charts/rekor/templates/server/pvc.yaml

What do you think? Just trying to prevent API changes in followup releases.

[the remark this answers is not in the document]
faisal-memon: I think the discussion on StatefulSet vs Deployment is out of scope for this pr. I think its worth investigating separately as a follow on. For the changes in the pr, are we set on a single PV? If so then i think the changes here are good to go.

marcofranssen (approved):
Since there is concensus in the approach by other maintainers I'll go ahead merging this one. 🤞 we won't have to change the values structure for persistence later to not introduce another breaking change on the public API.

PR #226, opened 2023-04-19
Fix tornjak rebase errors

mrsabath:


kfox1111 (approved):
Strange it duped that. Fixes look good to me.

PR #227, opened 2023-04-20
Spire-server plugin config (alternate)

kfox1111:
This patch enables end users to configure plugins in the the spire-server config.

Alternate approach to https://github.com/spiffe/helm-charts/pull/198


charts/spire/charts/spire-server/templates/configmap.yaml
```diff
@@ -31,58 +31,58 @@ server:
...
-        plugin_data:
-          database_type: "sqlite3"
-          connection_string: "/run/spire/data/datastore.sqlite3"
+    sql:
+      plugin_data:
+        database_type: "sqlite3"
+        connection_string: "/run/spire/data/datastore.sqlite3"
 
-  {{- with .Values.nodeAttestor.k8sPsat }}
+  {{- with .Values.nodeAttestor.k8s_psat }}
```
marcofranssen, line 39: Do we have to change the camelCase convention in values.yaml?

edwbuck: I think point is about if we are to support camel case, snake case, or something else.  I'm going to put the need to document the policy in an issue; but, I'm basically calling for the following general rules:

1.  Proper names are to be represented in the case they come in, preserving all "Helm supported" characters per Helm guidelines, with unsupported / problematic characters replaced by underscores.  The rational is that it helps one identify the item, increasing usability.  (This replaces the hyphen '-' with underscores, making some items look like snake case; but they aren't snake case, for example freedom-FooFigthers would be freedom_FooFighters, not freedom_Foo_Fighters).

2. All other names are to be snake case, starting with a lower case letter, per Helm's documented general conventions.  

3.  All compound words including Acronyms (NASA, for National Aeronautics and Space Administration) and Initialisms (NFL, for National Football League) are to maintain an all-uppercase spelling, in accordance with the English language conventions.  If the camel case causes confusion "NASARAM", rename the word, ideally putting any still-existing Acronym or Initialism at the end "memoryNASA"

4. All compound words including abbreviations (shortening of words) will have the abbreviation in all lowecase, with the exception of when camel case indicates the abbreviation would be capitalized.  Examples include apptBuilding, destinationAppt, usersGroupsEtc.
 
5. If an abbreviation, initialisizm, or acronym would noromally be used with periods in the name (for example P.M.) we remove the periods without supplying an "alternate" character to maintain the spaceing (P.M. becomes PM, not P_M_ or P_M)

Exceptions to these guildelines should be rare; but are permitted, provided the current set of maintainers agree to violate the established guidelines.

(camelcase) 'k8sPsat' is wrong, because Psat isn't a word, it's an acronym, shortening "Projected Service Account Token", the appropriate name according to camelcase is 'k8sPSAT

FYi:
correct snakecase would be 'k8s_PSAT'.  Correct camel_case would be 'k8sPSAT' and correct "name case" would be "k8s_psat"

The last one is the one that most closely resembles the name of the node attestor plugin, we should use it.

edwbuck: https://github.com/spiffe/helm-charts/issues/329

edwbuck: If there are no objections, I'll resolve this conversation.  Until then, let's let 'k8s_psat' go forward, as there is only a convention to use camel case, and this seems the correct time to break the convention as it's referring to proper name (of the plugin).

edwbuck: @marcofranssen apologies for resolving this conversation without hearing back from you, but a week has passed, so I am working under the assumption that you're ok with what I have said.  If you are not, please reopen the conversation.

edwbuck:
@kfox1111 please resolve the conflicts so we can review this in the context of the current chart offerings.

charts/spire/charts/spire-server/templates/configmap.yaml
```diff
@@ -32,59 +32,59 @@ server:
...
   {{- $upstreamAuthorityUsed = add1 $upstreamAuthorityUsed }}
   UpstreamAuthority:
-    - disk:
-        plugin_data:
-          cert_file_path: "/run/spire/upstream_ca/tls.crt"
-          key_file_path: "/run/spire/upstream_ca/tls.key"
-          {{- if ne .secret.data.bundle "" }}
-          bundle_file_path: "/run/spire/upstream_ca/bundle.crt"
-          {{- end }}
+    disk:
```
marcofranssen, line 65: Isn't this breaking, I thought we required those `-` earlier.

.github/tests/upstream-authority-cert-manager/values.yaml
```diff
@@ -1,5 +1,5 @@
 spire-server:
   upstreamAuthority:
-    certManager:
+    cert-manager:
```
marcofranssen, line 3: Why are we swithcing this from camelCasing?

I prefer to continue with the camelCasing convention in the Helm API as opposed to mixing in the spire plugin naming on the API which doesn't make it very consistent.

kfox1111: See https://github.com/spiffe/helm-charts/pull/227#discussion_r1173020275

charts/spire/charts/spire-server/templates/configmap.yaml
kfox1111: The macro makes it optional, so values merging works better.

kfox1111: All the tests pass too

edwbuck:
@marcofranssen Please come to a resolution of whether #198 or #227 is the solution to go forward with within a week, or propose an alternative. We have been deadlocked in this choice for more than a month, and currently Ken reports that you're not fully pleased with either solution.  It's ok to not like specific solutions, but we need a path forward because plugins are a thing we need to support.

marcofranssen:
See https://github.com/spiffe/helm-charts/pull/198#issuecomment-1618147499

Issue #195, opened 2023-04-06
Add test for global config overrides


marcofranssen:

kfox1111:
Should we test it as part of the production example one as we'd expect any production deployment to be actually overriding the values? Would save an extra set of tests.

PR #229, opened 2023-04-21
Deny production runs of example.org trust domains

kfox1111:
This pr adds some tests to ensure that if the user is deploying for production that they set appropriate settings.

fixes: https://github.com/spiffe/helm-charts/issues/195


marcofranssen (commented):
Do we really want to enforce people to override? If they are just doing a quick install for testing, shouldn't that be possible with the default values?

If we want people to explicitly set those it would be better to remove the defaults.

This feels a bit like a unnecessary additional config value.

kfox1111:
> Do we really want to enforce people to override? If they are just doing a quick install for testing, shouldn't that be possible with the default values?

That's the point of the pr. If you just helm install the chart, it just works with example.org for a quick install for testing as it does today.

If you say explicitly, your trying to deploy for production by adding examples/production/values.yaml, then it forces you to specify the things you should be specifying. No one should be deploying a production spire as example.org? The pr helps you find all the things you need to override to get a valid production site by enforcing you set them.
 
> If we want people to explicitly set those it would be better to remove the defaults.

That would make the, "just kick the tires" install much harder. But I could get behind that if there is agreement that is easier. It would be less complicated in the chart, just harder to get an example going. We could put the example values in examples/testing-deployment/values.yaml or something and then have people need to use that for a test deployment?

> This feels a bit like a unnecessary additional config value.

Necessary or unnecessary depends on the usability path question above.

charts/spire/templates/_spire-lib.tpl
```diff
@@ -50,3 +50,15 @@
...
 {{- end }}
+
+{{/*
+Take in an array of, '.', a failure string to display, and boolean to to display it,
+if strict checking is enabled and the boolean is true
+*/}}
+{{- define "spire-lib.strict-check" }}
+{{- if (dig "spire" "strictChecking" false (index . 0).Values.global) }}
+{{- if (index . 2) }}
+{{- fail (index . 1) }}
```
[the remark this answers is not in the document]
kfox1111: Yeah... A usage documentation item would be good. Will add.

Ordering of the yaml doesn't matter. useage is like:
```
{{- include "spire-lib.strict-check" (list . "trustDomain must be set" (eq (include "spire-lib.trust-domain" .) "example.org"))}}
```

where the arguments are specified in a list so wont be reordered, but takes up less space then doing it with dictionary passing.

args are root, an error to display, and a boolean on if to fail and display when true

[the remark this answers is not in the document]
kfox1111: actually, the arguments are kind of described here:
https://github.com/spiffe/helm-charts/pull/229/files#diff-789477a4f2a924ca09b7f80a469fc8827159a3d3d8a6281c37018dc16b60c094R55-R56

There any particular wording changes that might make it more understandable?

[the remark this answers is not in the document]
kfox1111: > * What is `index . 2`?
> * What is `index . 1`?

Oh. Maybe I'm not understanding the question either....
```index thing 1``` is what might be done in other languages like ```thing[1]``` or sometimes ```thing.1```. its looking up the second argument passed to the macro in a list. (0 based indexing)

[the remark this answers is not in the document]
Comment (marcofranssen) [reply charts/spire/templates/_spire-lib.tpl:63]: I'm also having some difficulty processing this one

```
{{- include "spire-lib.strict-check" (list . "trustDomain must be set" (eq (include "spire-lib.trust-domain" .) "example.org"))}}
```

I think even though this might be less syntax a dict could make it more descriptive and understandable. After reading this example 3 times I'm still not sure I get it.

File: charts/spire/charts/spire-agent/templates/configmap.yaml [hunk 2]
```diff
@@ -1,3 +1,5 @@
+{{- include "spire-lib.strict-check" (list . "clusterName must be set" (eq (include "spire-lib.cluster-name" .) "example-cluster"))}}
```
edwbuck, line 1: We should not provide a default cluster name or trust domain.

Out of all of the settings, these should be specified by the user 100% of the time.

edwbuck (dismissed):

charts/spire/templates/_spire-lib.tpl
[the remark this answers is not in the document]
marcofranssen: Ah got it now. Probably a good suggestion to assign them to variables that makes it more understanable indeed.

I would suggest following names.

```
{{ $root := index 0 }}
{{ $message := index 1 }}
{{ $condition := index 2 }}
```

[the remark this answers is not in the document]
kfox1111: Done

examples/production/values.yaml
```diff
@@ -1,4 +1,6 @@
 global:
+  spire:
+    strictChecking: true
```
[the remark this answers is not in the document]
kfox1111: Done

charts/spire/charts/spire-agent/templates/configmap.yaml
kfox1111: What about macro's suggestion of still keeping the ability to easily kick the tires, but turn it off by default?

examples/production/values.yaml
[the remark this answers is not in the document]
kfox1111: Actually, it looks like a bigger lift to make sure all the tests are then specifying the newly required config too. I can do that, but we should decide that's the way we want to go, first.

charts/spire/charts/spire-agent/templates/configmap.yaml
edwbuck: I'm willing to be flexible, but I really don't like the precedent that we either have already set, or are committing to setting.

If a installation using our Helm charts isn't production quality, then we should fix it to be production quality, adjust what we decide is production quality, or adjust our documentation to indicate that initially the installation won't be production quality, but after you set a specific property, it will be?

Personally I don't care which one is the path forward; and in the interest of getting some movement, and @marcofranssen call that we easily "kick the tires, but turn (I'm not sure if it is "production quality" or "easy non-production prototype installations", I'm going to withdraw my objection and clear this for the other maintainers to move in whatever direction they deem fit.

kfox1111 (commented):
Update to include new global.

Also, set default back to false until we determine that tests should be rewritten

faisal-memon:
> Something like productionChecks for the value instead.

That sounds good to me.

charts/spire/values.yaml
```diff
@@ -24,6 +24,9 @@ global:
       ## @param global.spire.image.registry Override all Spire image registries at once
       registry: ""
 
+    ## @param global.spire.productionChecks Set to false to disable production requirements to enable you to kick the tires with spire without needing a lot of configuration.
```
faisal-memon, line 27:
```suggestion
    ## @param global.spire.productionChecks Check values, such as trustDomain, are suitable for production deployment
```

faisal-memon: I like the description to say what enabling the flag does

faisal-memon (approved):

charts/spire/charts/spire-server/values.yaml
```diff
@@ -225,7 +225,7 @@ federation:
 
 ca_subject:
   ## @param ca_subject.country Country for Spire server CA
-  country: NL
+  country: ARPA
```
marcofranssen, line 228: https://en.wikipedia.org/wiki/List_of_ISO_3166_country_codes

Which country does ARPA represent?

charts/spire/values.yaml
```diff
@@ -24,6 +24,9 @@ global:
       ## @param global.spire.image.registry Override all Spire image registries at once
       registry: ""
 
+    ## @param global.spire.productionChecks Check values, such as trustDomain, are suitable for production deployment
+    productionChecks: false
```
marcofranssen, line 28: Shall we call this `strictMode`? that is also something used quite commonly in Javascript, React, Typescript world.

charts/spire/charts/spire-server/values.yaml
kfox1111: https://www.digicert.com/kb/ssl-certificate-country-codes.htm

ARPA should be valid, but shouldn't be used in any org I think with spire. so easy to check if its been set or not by the end user?

charts/spire/values.yaml
kfox1111: @faisal-memon ?

examples/production/values.yaml
```diff
@@ -2,6 +2,8 @@ global:
   telemetry:
     prometheus:
       enabled: true
+  spire:
+   productionChecks: false
```
marcofranssen, line 6:
```suggestion
   strictMode: true
```

As suggested earlier, maybe call it strictMode. Shouldn't the value be true for this test?

charts/spire/templates/_spire-lib.tpl
```diff
@@ -107,3 +107,18 @@ rules:
...
+
+{{/*
+Take in an array of, '.', a failure string to display, and boolean to to display it,
+if production checking is enabled and the boolean is true
+*/}}
+{{- define "spire-lib.production-check" }}
+{{ $root := index . 0 }}
+{{ $message := index . 1 }}
+{{ $condition := index . 2 }}
+{{- if (dig "spire" "productionChecks" false $root.Values.global) }}
```
marcofranssen, line 119:
```suggestion
{{- if (dig "spire" "strictMode" false $root.Values.global) }}
```

charts/spire/templates/_spire-lib.tpl
```diff
@@ -107,3 +107,18 @@ rules:
 {{- end }}
 {{- include "spire-lib.image" $root }}
 {{- end }}
+
+{{/*
+Take in an array of, '.', a failure string to display, and boolean to to display it,
+if production checking is enabled and the boolean is true
+*/}}
+{{- define "spire-lib.production-check" }}
```
marcofranssen, line 115:
```suggestion
{{- define "spire-lib.check-strict-mode" }}
```

charts/spire/templates/_spire-lib.tpl
```diff
@@ -107,3 +107,18 @@ rules:
 {{- end }}
 {{- include "spire-lib.image" $root }}
 {{- end }}
+
+{{/*
+Take in an array of, '.', a failure string to display, and boolean to to display it,
+if production checking is enabled and the boolean is true
```
marcofranssen, line 113:
```suggestion
if strictMode is enabled and the boolean is true
```

charts/spire/charts/spire-server/templates/configmap.yaml
```diff
@@ -1,3 +1,9 @@
+{{- include "spire-lib.production-check" (list . "clusterName must be set" (eq (include "spire-lib.cluster-name" .) "example-cluster"))}}
+{{- include "spire-lib.production-check" (list . "trustDomain must be set" (eq (include "spire-lib.trust-domain" .) "example.org"))}}
+{{- include "spire-lib.production-check" (list . "jwtIssuer must be set" (eq (include "spire-lib.jwt-issuer" .) "https://oidc-discovery.example.org"))}}
+{{- include "spire-lib.production-check" (list . "ca_subject.county must be set" (eq .Values.ca_subject.country "ARPA"))}}
+{{- include "spire-lib.production-check" (list . "ca_subject.organization must be set" (eq .Values.ca_subject.organization "Example"))}}
+{{- include "spire-lib.production-check" (list . "ca_subject.common_name must be set" (eq .Values.ca_subject.common_name "example.org"))}}
```
marcofranssen, line 6:
```suggestion
{{- include "spire-lib.check-strict-mode" (list . "clusterName must be set" (eq (include "spire-lib.cluster-name" .) "example-cluster"))}}
{{- include "spire-lib.check-strict-mode" (list . "trustDomain must be set" (eq (include "spire-lib.trust-domain" .) "example.org"))}}
{{- include "spire-lib.check-strict-mode" (list . "jwtIssuer must be set" (eq (include "spire-lib.jwt-issuer" .) "https://oidc-discovery.example.org"))}}
{{- include "spire-lib.check-strict-mode" (list . "ca_subject.county must be set" (eq .Values.ca_subject.country "ARPA"))}}
{{- include "spire-lib.check-strict-mode" (list . "ca_subject.organization must be set" (eq .Values.ca_subject.organization "Example"))}}
{{- include "spire-lib.check-strict-mode" (list . "ca_subject.common_name must be set" (eq .Values.ca_subject.common_name "example.org"))}}
```

charts/spire/charts/spire-agent/templates/configmap.yaml
```diff
@@ -1,3 +1,5 @@
+{{- include "spire-lib.production-check" (list . "clusterName must be set" (eq (include "spire-lib.cluster-name" .) "example-cluster"))}}
+{{- include "spire-lib.production-check" (list . "trustDomain must be set" (eq (include "spire-lib.trust-domain" .) "example.org"))}}
```
marcofranssen, line 2:
```suggestion
{{- include "spire-lib.check-strict-mode" (list . "clusterName must be set" (eq (include "spire-lib.cluster-name" .) "example-cluster"))}}
{{- include "spire-lib.check-strict-mode" (list . "trustDomain must be set" (eq (include "spire-lib.trust-domain" .) "example.org"))}}
```

charts/spire/charts/spiffe-oidc-discovery-provider/templates/configmap.yaml
```diff
@@ -1,3 +1,5 @@
+{{- include "spire-lib.production-check" (list . "trustDomain must be set" (eq (include "spire-lib.trust-domain" .) "example.org"))}}
+{{- include "spire-lib.production-check" (list . "jwtIssuer must be set" (eq (include "spire-lib.jwt-issuer" .) "https://oidc-discovery.example.org"))}}
```
marcofranssen, line 2:
```suggestion
{{- include "spire-lib.check-strict-mode" (list . "trustDomain must be set" (eq (include "spire-lib.trust-domain" .) "example.org"))}}
{{- include "spire-lib.check-strict-mode" (list . "jwtIssuer must be set" (eq (include "spire-lib.jwt-issuer" .) "https://oidc-discovery.example.org"))}}
```

charts/spire/README.md
```diff
@@ -98,6 +98,7 @@ Now you can interact with the Spire agent socket from your own application. The
 | `global.spire.trustDomain`              | The trust domain for Spire install                                                                         | `example.org`                |
 | `global.spire.upstreamServerAddress`    | Set what address to use for the upstream server when using nested spire                                    | `""`                         |
 | `global.spire.image.registry`           | Override all Spire image registries at once                                                                | `""`                         |
+| `global.spire.productionChecks`         | Check values, such as trustDomain, are suitable for production deployment                                  | `false`                      |
```
marcofranssen, line 101:
```suggestion
| `global.spire.strictMode`         | Check values, such as trustDomain, are suitable for production deployment                                  | `false`                      |
```

examples/production/values.yaml
kfox1111: Good catch. broke in the last rename I guess.

marcofranssen (changes requested):

marcofranssen (approved):
LGTM 🚀

PR #230, opened 2023-04-21
Add additional domains to JWT issued items.

kfox1111:
This patch makes it possible to override all the jwt issuers globally.

## Release Notes

the spiffe-oidc-discovery-provider values has a breaking change due to a rename of the value:

```diff
- domains: []
+ additionalDomains: []
```


charts/spire/charts/spiffe-oidc-discovery-provider/README.md
```diff
@@ -26,8 +26,7 @@ A Helm chart to install the SPIFFE OIDC discovery provider.
 | config.acme.directoryUrl | string | `"https://acme-v02.api.letsencrypt.org/directory"` |  |
 | config.acme.emailAddress | string | `"letsencrypt@example.org"` |  |
 | config.acme.tosAccepted | bool | `false` |  |
-| config.domains[0] | string | `"localhost"` |  |
-| config.domains[1] | string | `"oidc-discovery.example.org"` |  |
+| config.domains | list | `[]` | Override the domains to use. Defaults to jwtIssuer and localhost. |
```
marcofranssen, line 29: I would still prefer to have those defaults as real defaults in values.yaml.

kfox1111: I'm trying to simplify what a user would have to do to deploy a production system.

```
global:
  jwtIssuer: thingy
```

is much easier then:
```
spire-server:
  jwtIssuer: thingy
spiffe-oidc-provider:
  domains:
  - localhost
  - thingy
```

---

The distinct(concat... thing could help though. What if we set config.domain to be just ['local'] by default and have the rest of the patch as is? Might meet both needs a little better?

kfox1111: Added the uniq(concat thing.

charts/spire/charts/spiffe-oidc-discovery-provider/templates/configmap.yaml
```diff
@@ -8,9 +8,8 @@ domains:
   - "{{ include "spiffe-oidc-discovery-provider.fullname" . }}"
   - "{{ include "spiffe-oidc-discovery-provider.fullname" . }}.{{ include "spiffe-oidc-discovery-provider.namespace" . }}"
   - "{{ include "spiffe-oidc-discovery-provider.fullname" . }}.{{ include "spiffe-oidc-discovery-provider.namespace" . }}.svc.{{ include "spire-lib.cluster-domain" . }}"
-  {{- if gt (len .Values.config.domains) 0 }}
-  {{- .Values.config.domains | toYaml | nindent 2 }}
-  {{- end }}
+  {{- $jwtIssuer := include "spire-lib.jwt-issuer" . }}
```
marcofranssen, line 11: This will tightly couple the sub chart with the root level chart. If we break the chart out it all of a sudden depends on this global that will not exist when using the chart individually. Basically a dependency cycle.

charts/spire/charts/spiffe-oidc-discovery-provider/values.yaml
```diff
@@ -89,12 +89,14 @@ insecureScheme:
       #   cpu: 100m
       #   memory: 64Mi
 
+jwtIssuer: oidc-discovery.example.org
+
 config:
   # -- The log level, valid values are "debug", "info", "warn", and "error"
   logLevel: info
+  # -- Override the domains to use. Defaults to jwtIssuer and localhost.
```
marcofranssen, line 97:
```suggestion
  # -- Override the domains to use. (jwtIssuer is automatically added to this list).
```

edwbuck:
@kfox1111 Please peek into the CI failures.  I feel that if we can get this passing in CI, reviews will then follow.

edwbuck:
> I stopped clicking the merge button because it wasnt getting reviews in so long. It takes effort and no effort was being put in. If it will get reviews then, I'll fix it, again.

I understand, but it is a bit ironic that such a statement is in direct reply to a review attempt. :)

We have a lot of stuff in the backlog, but we are burning it down.  It was up to about 30 open issues not too long ago, now it's really close to 15.  This means that some of these older items are finally getting the attention they deserve.

In any case, we need to apply work to the old PRs to get them out of the system, or they'll just stay in limbo forever.

charts/spire/charts/spiffe-oidc-discovery-provider/README.md
```diff
@@ -26,8 +26,7 @@ A Helm chart to install the SPIFFE OIDC discovery provider.
 | config.acme.directoryUrl | string | `"https://acme-v02.api.letsencrypt.org/directory"` |  |
 | config.acme.emailAddress | string | `"letsencrypt@example.org"` |  |
 | config.acme.tosAccepted | bool | `false` |  |
-| config.domains[0] | string | `"localhost"` |  |
-| config.domains[1] | string | `"oidc-discovery.example.org"` |  |
+| config.domains | list | `["localhost"]` | Override the domains to use. (jwtIssuer is automatically added to this list). |
```
edwbuck, line 29: The need for the comment here is an indication that the API is not expressive enough to hold up without the additional commentary.

I can see two paths forward on the API side.

1. Keep the "config.domains" key, and add a _helpers.tpl test to ensure that it includes the required matching jwtIssuer domain.
2. Alter the "config.domains" key to be named "config.additionalDomains" or something like that.

One is clearly more work than the other, but I haven't thought about the other aspects of each choice yet.  There are likely other options available too.  If you address this in a different way, that's fine by me; I'd like you to consider which way would service your needs the best.

charts/spire/charts/spiffe-oidc-discovery-provider/templates/_helpers.tpl
```diff
@@ -85,4 +85,3 @@ Create the name of the service account to use
 {{- define "spiffe-oidc-discovery-provider.workload-api-socket-path" -}}
 {{- printf "/spiffe-workload-api/%s" .Values.agentSocketName }}
 {{- end }}
```
edwbuck, line 87: This additional dash is just a bug, and not part of this effort.  Next time submit a fix for it in a different PR.

charts/spire/charts/spiffe-oidc-discovery-provider/values.yaml
```diff
@@ -101,12 +101,14 @@ insecureScheme:
       #   cpu: 100m
       #   memory: 64Mi
 
+jwtIssuer: oidc-discovery.example.org
+
 config:
   # -- The log level, valid values are "debug", "info", "warn", and "error"
   logLevel: info
+  # -- Override the domains to use. (jwtIssuer is automatically added to this list).
```
edwbuck, line 109: This kind of comment also reinforces my previous statement that the change makes the configuration "API" not align with what it says it does.  Consider calling the list "additionalDomains".

edwbuck (changes requested):
See if additionalDomains is a better name, because all the comments about how these are in addition to the jwtIssuer domain hints that the actual name you're using isn't the right one.

charts/spire/charts/spiffe-oidc-discovery-provider/templates/_helpers.tpl
kfox1111: Its not a dash. it looks like an editor removed the blank line at the end of the file. I can undo that.

charts/spire/charts/spiffe-oidc-discovery-provider/README.md
kfox1111: I like the idea of renaming the option to be more representative of what it does. additionalDomains makes sense to me. The rename would be easy. I'll submit that and if we want to go the other way, we can do that instead.

kfox1111: Done

edwbuck: Verified changed to additional domains

edwbuck (approved):
Updated to clarify the purpose of the parameter

charts/spire/charts/spiffe-oidc-discovery-provider/templates/configmap.yaml
marcofranssen: I think this hyperlink moved to wrong line by updates on the code. Was trying to see which part resolves it, but can't relate that in the line linked.

kfox1111: Yeah. the review drifted. Try:

https://github.com/spiffe/helm-charts/pull/230/files#diff-789477a4f2a924ca09b7f80a469fc8827159a3d3d8a6281c37018dc16b60c094R18-R21
And 
https://github.com/spiffe/helm-charts/pull/230/files#diff-7be84ce176a12ba3b77f98055916d01a5c97e1688a3a32e361495839b5ef69caR107

Should cause it to still work when split.

marcofranssen (approved):
LGTM :rocket:

Issue #96, opened 2023-03-06
Add support for spire-server admin_ids


marcofranssen:
See https://github.com/spiffe/spire/blob/main/doc/spire_server.md#server-configuration-file

Would be great if we somehow can configure the admin_ids.

E.g. via spire-controller-manager CRDs? Maybe as a separate Helm chart to provision admin_ids CRDs so we can configure the same on spire-server.

PR #231, opened 2023-04-22
Add spire-server config section to allow customization

kfox1111:
This patch adds a config section to spire-server that gets merged into the server block of the config. This allows things such as admin_ids to be easily configured.

fixes: https://github.com/spiffe/helm-charts/issues/96


mrsabath (approved):
lgtm

faisal-memon:
I think its confusing to have this `config{}` block and then to have some of the config broken out. We should probably settle on an approach.

marcofranssen:
> I think its confusing to have this `config{}` block and then to have some of the config broken out. We should probably settle on an approach.

For the public API I think we have discussed a couple of times, it is probably better to code all the explicit options as opposed to relying on this magic variable that accepts all the spire configs.

It will take a bit more work to explicitly add all those options, but it will also add the documentation at chart level.

kfox1111:
The documentation aspect is nice though. Although with passthrough, we can also point at the spire docs rather then our own.

I think its going to take more then a bit more work though. This pr is just a piece of the overall problem. Looking at the documentation, there are tons of options users may need to set.  In config, in the various plugins section for the agent and spire server. (Documented configuration here https://github.com/spiffe/spire/tree/main/doc)

https://github.com/spiffe/helm-charts/blob/main/project/overview.md#what-is-the-scope-of-this-project says
> in any SPIRE documented configuration

I do think we want the users to be able to use the options.

Do we really want to be explicitly including code for all the options though instead of allowing some amount of passthrough? Its going to be a very large pr, and much more work to maintain. Every time spire adds a new one, we need to too. If we don't stay on top of it, users won't have the options they need. it being a lot of work will also put pressure on us not to add the options which is also not good.

marcofranssen:
Like we did with upstream authority we can do for more type of plugins

kfox1111:
I'm afraid we'd end up with thousands of lines of code we'd have to maintain that way then? Or is there another way I'm missing?

faisal-memon:
Can we take a hybrid approach where we explicitly define the most common options in the values.yaml? The template can just be range under the hood to generate the rest. We can add a comment that says any config defined in the spire docs but not explicitly specified here is still supported.

marcofranssen:
> Can we take a hybrid approach where we explicitly define the most common options in the values.yaml? The template can just be range under the hood to generate the rest. We can add a comment that says any config defined in the spire docs but not explicitly specified here is still supported.

Yeah same like upstreamAuthority we can define such a thing for each plugin type

edwbuck:
> > I think its confusing to have this `config{}` block and then to have some of the config broken out. We should probably settle on an approach.
> 
> For the public API I think we have discussed a couple of times, it is probably better to code all the explicit options as opposed to relying on this magic variable that accepts all the spire configs. 

I agree that this is a generally better approach.  Of course, this can then over-populate the values.yaml file with many items that are of little consequence; but, such things are more easily fixed by having "if not set, use this default" style handling of values.yaml.

This is a more detailed description than the prior "hybrid approach" reference made by Faisal because it proposes a policy rather than a feature. (The feature being not too many values.yaml entries).  Such an approach will likely require updates to the generated README.md documents, but I think we can manage that.  Until then, let's error on detailing all the values, as it "poises" us for the next step, which is to determine which ones have reasonable defaults and thus can be trimmed with "if not set, then set the value to X" style directives in the templates.

faisal-memon:
We decided to not go forward with this config merge concept at the last sync meeting. Is it ok to close this? The issue can stay open to track the greater issue of how to handle all the missing config options.

PR #233, opened 2023-04-22
Internal database suport

kfox1111:
This pr adds built in mysql and postgresql support to the spire-server chart.


marcofranssen:
Maybe adding documentation on using these charts is sufficient for now:

- https://artifacthub.io/packages/helm/bitnami/postgresql
- https://artifacthub.io/packages/helm/bitnami/mysql

I guess many people prefer to use their own way of installing and managing the databases or are relying on charts like the bitnami ones etc already. In my case I would probably even Go with a managed version from the Cloud vendors.

Therefore I think it is better to just document how to integrate with another helm chart rather then embedding a part of of a database setup in our chart. E.g. those dedicated charts also handle other database specific topics, which are not in our implementation just now.

kfox1111:
There is another pr that does external databases support. This one just adds the option to optionally use the spire chart to nest the db chart. This is very commonly done in other charts that need a database. It does make it easier to configure the database, only having to specify the config once rather then twice. Once in spire-server instead of once to the external db chart and once to the spire-server chart. And certain properties can be automaticaly correctly configured such as db hostname.

kfox1111:
uh... where did the macro code go?

edwbuck:
I've been looking at this for the better part of a day.  I had this long post about the pros and cons, but I realized nobody would read it.  Apologies, but this is the short version.

I'm convinced the approach will technically work.  I also feel we should avoid it because the only answer I can imagine to the question "would users of our charts come to us when encountering issues with the nested database chart?" is "yes".

When they come to us, we'll have to tell them that we aren't really in the database business, they need to go elsewhere.  That wastes their time and makes us look less competent (because we're putting on airs of managing stuff we don't manage).  We can attempt to document this in our wiki, but just saying we're not supporting some portion of what we deliver is (in my words, not yours) a cop-out.

And that point side-steps the complexity of having two ways to configure MySQL / PostGreSQL, as we will still need the non-nested approach for all of those users that already have database instances deployed (or want to serve multiple schemas from the same engine).

Finally, after a day free-time in talking with Kevin, we haven't even described the transition from sqlite3 to a nested chart MySQL, or a nested chart MySQL to an external MySQL or any of the other database-to-database transitions (or if we would support them).  I don't think anyone would care much if we don't support transitions changing the database engine, but I do think that a few people would get upset if we didn't support "I chose the wrong MySQL setup, so now I want to migrate to the right one"

kfox1111:
@edwbuck thanks for taking the time to do a review, ensure you had all he information to make an informed decision, have a discussion about the pro's and cons, and consider if there were any other approaches that could work.

There's a bunch of tradeoffs to be made and I think we weight them slightly differently. And that's ok. Your concern about users coming to us for database support is valid. This early into the life of the project, we cant afford the risk of that. So, after this discussion, I agree that this is probably not the right time to introduce this functionality, if we ever do.

I'm closing this pr.

PR #237, opened 2023-04-25
Move check versions script + fix a bug

marcofranssen:
See https://github.com/spiffe/helm-charts/pull/236 for the result.

https://github.com/spiffe/helm-charts/actions/runs/4795275892 for the workflow execution.

### DIFF
--- a/.github/scripts/update-versions.sh
+++ b/.github/scripts/update-versions.sh
@@ -18,7 +18,7 @@ jq -r ".[].name" "${CHARTJSON}" | while read -r CHART; do
   echo "  latest version: ${LATEST_VERSION}"
   if [ "x${VERSION}" != "x${LATEST_VERSION}" ]; then
     echo "  New version found!"
-    jq "(${ENTRYQUERY}).version |= "'"${LATEST_VERSION}"' "${CHARTJSON}" > /tmp/$$
+    jq "(${ENTRYQUERY}).version |= \"${LATEST_VERSION}\"" "${CHARTJSON}" > /tmp/$$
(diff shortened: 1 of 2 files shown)


kfox1111:
Not seeing any fixes to the job. Just better formatting.

faisal-memon:
> ah. I think I finally understand after doing some debugging. Can we break this pr up into two for posterity (or 2 commits)? Its really hard to follow that there is actually one small fix in here to make sure the new version gets into the json file with it being hidden by all the other changes going on that's just format changes without any functional difference, along with a file rename.

@marcofranssen Or do a move of the file to the new directory instead of delete and create a new file.

kfox1111:
Here's the isolated fix. Reverted back to the code that was reviewed by @faisal-memon and me.
https://github.com/spiffe/helm-charts/pull/239

We can then merge this on top to move the file around and get the better formatting.

marcofranssen:
I don't get why you pr got reverted. I reviewed it as well. When I approved it merged and I figured all the patches where missing.

kfox1111:
> I don't get why you pr got reverted. I reviewed it as well. When I approved it merged and I figured all the patches where missing.

It happened here at line 21

kfox1111 (approved):
Works locally. Looks ok.

marcofranssen:
https://github.com/spiffe/helm-charts/pull/237/commits splitted the changes across 3 commits, still GitHub does not render the file as a rename due to the changes in the file.

To capture the specific changes in Git I splitted the commits, so `git blame`  can always show us why certain changes have been made.

PR #238, opened 2023-04-25
Build database connection_string from values

kfox1111:
This patch makes it possible to build up the connection_string for mysql and postgresql via values.


marcofranssen (approved):
LGTM :rocket:

PR #239, opened 2023-04-25
Fix the helm version update script

kfox1111:
The quoting was broken before a merge. This patch sets it back to what was tested before the change.


kfox1111:
We're just going to merge the fix and a bunch of reformatting of the script at once.

PR #242, opened 2023-04-26
Add workflow with shellcheck to do some sanity checks on shell scripts

marcofranssen:


.github/workflows/scripts/update-versions.sh
```diff
@@ -1,24 +1,24 @@
...
+  echo "  repo: ${REPO_URL}"
+  echo "  current version: ${VERSION}"
+  helm repo add "${CHART}" "${REPO_URL}" > /dev/null
+  helm repo update "${CHART}" > /dev/null
+  LATEST_VERSION=$(helm search repo --regexp "${CHART}/${CHART}\v" -o json | jq -r '.[0].version')
+  echo "  latest version: ${LATEST_VERSION}"
+  if [ "x${VERSION}" != "x${LATEST_VERSION}" ]; then
     echo "  New version found!"
-    jq "( $ENTRYQUERY ).version |= "'"${LATEST_VERSION}"' "${CHARTJSON}" > /tmp/$$
+    jq "(${ENTRYQUERY}).version |= "'"${LATEST_VERSION}"' "${CHARTJSON}" > /tmp/$$
```
kfox1111, line 21: The quoting here is still wrong.

.github/tests/namespace-override/post-install.sh
```diff
@@ -13,7 +14,7 @@ k_rollout_status () {
   kubectl rollout status --watch --timeout 30s --namespace "$1" "$2" "$3" | tail -n 1
 }
 
-RELEASE=$(helm ls --no-headers -n "${scenario}" | awk '{print $1}' | grep 'spire-[^-]*$')
+RELEASE=$(helm ls --no-headers -n "${NS}" | awk '{print $1}' | grep 'spire-[^-]*$')
```
kfox1111, line 17: Where is NS coming from?

.github/tests/namespace-override/post-install.sh
```diff
@@ -30,7 +31,7 @@ cat <<EOF >>"$GITHUB_STEP_SUMMARY"
 | spire-spiffe-oidc-discovery-provider | <pre>$(k_wait spire-server deployments.apps "${RELEASE}-spiffe-oidc-discovery-provider")</pre> |
 EOF
 
-if [ $1 -ne 0 ]; then
+if [ "$1" -ne '0' ]; then
```
kfox1111, line 34: I don't understand this one. != is used for strings, -ne for integers.

.github/tests/production-example/post-install.sh
```diff
@@ -38,14 +44,14 @@ cat <<EOF >>"$GITHUB_STEP_SUMMARY"
...
-| spire-agent                          | "$("${k_rollout_status[@]}" spire-system daemonset spire-agent)" |
+| spire-server                         | "$(k_rollout_status spire-server statefulset spire-server)" |
+| spire-controller-manager             | "$(k_rollout_status spire-server statefulset spire-controller-manager)" |
+| spire-spiffe-oidc-discovery-provider | "$(k_wait spire-server deployments.apps spire-spiffe-oidc-discovery-provider)" |
+| spire-spiffe-csi-driver              | "$(k_rollout_status spire-system daemonset spire-spiffe-csi-driver)" |
+| spire-agent                          | "$(k_rollout_status spire-system daemonset spire-agent)" |
 EOF
 
-if [ $1 -ne 0 ]; then
+if [ "$1" -ne '0' ]; then
```
kfox1111, line 54: same

kfox1111 (commented):
Most of this looks good. A couple of issues.

.github/workflows/scripts/update-versions.sh
marcofranssen: Yes, correct will resolve that in that other PR once this is merged, that other PR will have a much smaller diff.

.github/tests/namespace-override/post-install.sh
marcofranssen: woops good catch

.github/tests/prometheus/post-install.sh
```diff
@@ -0,0 +1,17 @@
+#!/usr/bin/env bash
+
+set -x
+
+SCRIPT=$(readlink -f "$0")
+SCRIPTPATH=$(dirname "$SCRIPT")
+scenario="${SCRIPTPATH##*/}"
```
kfox1111, line 7: This line is inconsistent then some of the rest? Should they all be the same?

.github/tests/production-example/post-install.sh
```diff
@@ -38,14 +44,14 @@ cat <<EOF >>"$GITHUB_STEP_SUMMARY"
 
 | workload                             | Status |
 | ------------------------------------ | ------ |
-| spire-server                         | "$("${k_rollout_status[@]}" spire-server statefulset spire-server)" |
-| spire-controller-manager             | "$("${k_rollout_status[@]}" spire-server statefulset spire-controller-manager)" |
-| spire-spiffe-oidc-discovery-provider | "$("${k_wait[@]}" spire-server deployments.apps spire-spiffe-oidc-discovery-provider)" |
-| spire-spiffe-csi-driver              | "$("${k_rollout_status[@]}" spire-system daemonset spire-spiffe-csi-driver)" |
-| spire-agent                          | "$("${k_rollout_status[@]}" spire-system daemonset spire-agent)" |
+| spire-server                         | "$(k_rollout_status spire-server statefulset spire-server)" |
```
kfox1111, line 47: Does this need the pre stuff?

kfox1111 (commented):
Couple more things

.github/tests/production-example/post-install.sh
marcofranssen: good catch

marcofranssen: this is resolved in #88

kfox1111 (approved):
Close enough. It makes things way better and future pr's already proposed will fix the rest.
LGTM

PR #245, opened 2023-04-27
Switch image.version to image.tag

kfox1111:
The convention in most charts is to use image.tag. This pr updates the values to use it instead of the less standard image.version.

### DIFF
--- a/charts/spire/charts/spire-agent/values.yaml
+++ b/charts/spire/charts/spire-agent/values.yaml
@@ -10,7 +10,7 @@ image:
   # -- The image pull policy
   pullPolicy: IfNotPresent
   # Overrides the image tag whose default is the chart appVersion.
-  version: ""
+  tag: ""
 
 imagePullSecrets: []
 nameOverride: ""
@@ -86,7 +86,7 @@ waitForIt:
     repository: chainguard/wait-for-it
     # -- The image pull policy
     pullPolicy: IfNotPresent
-    version: latest-20230113
+    tag: latest-20230113
   resources: {}
 
 # workloadAttestors determine a workload's properties and then generate a set of selectors associated with it.
--- a/charts/spire/charts/spire-agent/values.yaml
+++ b/charts/spire/charts/spire-agent/values.yaml
@@ -9,6 +9,8 @@ image:
   repository: spiffe/spire-agent
   # -- The image pull policy
   pullPolicy: IfNotPresent
+  # -- This value is deprecated in favor of tag. (Will be removed in a future release)
+  version: ""
   # -- Overrides the image tag whose default is the chart appVersion.
   tag: ""
 
@@ -86,8 +88,10 @@ waitForIt:
     repository: chainguard/wait-for-it
     # -- The image pull policy
     pullPolicy: IfNotPresent
+    # -- This value is deprecated in favor of tag. (Will be removed in a future release)
+    version: ""
     # -- Overrides the image tag
-    tag: latest
+    tag: latest-20230517
   resources: {}
 
 # workloadAttestors determine a workload's properties and then generate a set of selectors associated with it.
--- a/charts/spire/charts/spire-agent/values.yaml
+++ b/charts/spire/charts/spire-agent/values.yaml
@@ -9,7 +9,7 @@ image:
   repository: spiffe/spire-agent
   # -- The image pull policy
   pullPolicy: IfNotPresent
-  # Overrides the image tag whose default is the chart appVersion.
+  # -- Overrides the image tag whose default is the chart appVersion.
   tag: ""
 
 imagePullSecrets: []
@@ -86,7 +86,8 @@ waitForIt:
     repository: chainguard/wait-for-it
     # -- The image pull policy
     pullPolicy: IfNotPresent
-    tag: latest-20230113
+    # -- Overrides the image tag
+    tag: latest
   resources: {}
 
 # workloadAttestors determine a workload's properties and then generate a set of selectors associated with it.
--- a/charts/spire/charts/spire-server/values.yaml
+++ b/charts/spire/charts/spire-server/values.yaml
@@ -12,6 +12,8 @@ image:
   repository: spiffe/spire-server
   # -- The image pull policy
   pullPolicy: IfNotPresent
+  # -- This value is deprecated in favor of tag. (Will be removed in a future release)
+  version: ""
   # -- Overrides the image tag whose default is the chart appVersion.
   tag: ""
 
@@ -169,6 +171,8 @@ controllerManager:
     repository: spiffe/spire-controller-manager
     # -- The image pull policy
     pullPolicy: IfNotPresent
+    # -- This value is deprecated in favor of tag. (Will be removed in a future release)
+    version: ""
     # -- Overrides the image tag
     tag: "0.2.2"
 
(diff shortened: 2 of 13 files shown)


marcofranssen:
Most charts depends on what charts you are used to.

https://github.com/sigstore/helm-charts/blob/main/charts/common/templates/_images.tpl

The sigstore charts all use the convention I have adopted here as well. Don't really see to value of changing that.

kfox1111:
> Most charts depends on what charts you are used to.

That's fair.

I'm recommending the convention that helm itself, at least as far back as 2.0.0 (Nov 16, 2016), recommended whatever you issue 'helm create foo". (I don't have access to download 1.0.0)

It really isn't a very big deal at the end of the day. But it is a very very old/common convention.

Its a tiny thing I agree and largely doesn't matter. But lots of little tiny improvements add up to bigger usability gains in the long run.

edwbuck:
@marcofranssen You seemed to reject the request in your clear explanation; but, I can't determine if you actually rejected it because I see an explanation but not the course of action you wish to take.

I would urge you to change your mind, as tags are a superset of versions conceptually, such that a tag may contain a version, or any other text that helps someone manage which specific container should be deployed ("qa", "prod", etc.)

@kfox1111 If you feel that the benefit of using the word "tag" is greater that the benefit of using the word "version".  Please ensure that we don't use the two as synonyms of each other.  In the examples below, tags are being attached to tagged TARGET_IMAGES.

`docker tag 0e5574283393 fedora/httpd:version1.0`
`docker tag http fedora/httpd:version1.0`
`docker tag version1.0 fedora/httpd:version1.0`

This tounge twister of an explanation is based on the usage statement of
`docker tag SOURCE_IMAGE[:TAG] TARGET_IMAGE[:TAG]` as noted [in the documentation](https://docs.docker.com/engine/reference/commandline/tag).

As tags only _SOMETIMES_ contain text that seem to look like traditional (or even non-traditional) versions, I would prefer we shift to the use of the word "tag" for consistency, even though older documentation primarily demonstrated tagging with version numbers.  This use of "tag" permits hints that the tracked item can change, based on the tag's change, as is common with "latest", "nightly", or even "passed_acceptance" (where such a tag has no meaning outside of the organization).

marcofranssen:
The downside of changing this thing is we have a (unnecessary) breaking change in the API. If everyone feels we need to change this I will follow, but still don't see the value of really changing that.

edwbuck:
I think we've been thinking of this as, if we have \"tag\", we can't have \"version\". 

This is one of those rare cases where we can have both.

1. If the more-modern "tag" is set, we use it's value.
2. If the traditional "version" is set, we use it's value.
3. If both are set, we issue an error and fail.
4. If neither are set, we issue an error to set the "tag".

I understand the rules established to deprecate items; and, I understand the rules established that consider the values.yaml as part of the "API" that if changed, needs changed in ways that APIs are managed.  What I think all of those rules have done is combine [in a way that presents a false dichotomy](https://en.wikipedia.org/wiki/False_dilemma).

Basically, having one value key doesn't mean we must eject the other, or migrate the other.  There's no need to take a stand on the "right" one.  There's not even a need to stop supporting "version".  We can support both for many years to come, as it is a relatively clear cut case of one being an alias of the other.  If this "dual" support causes problems, we can also submit future issues to push the users to migrate into the last "key" standing.

kfox1111:
Yeah. That's a great point, that makes sense to me and feels like a good compromise.

faisal-memon:
So in this scenario `version` would be a hidden configurable? Not listed in the docs.

marcofranssen:
> So in this scenario `version` would be a hidden configurable? Not listed in the docs.

We can't hide it from the docs, what we can do is add a description `Deprecated, please use tag from now on, this setting will be removed in a future release`. This offers people a migration path while we shortly support one or the other.

edwbuck:
Another option for the docs, is "version" - "the tag value to pull if tag is not set, to support legacy conventions", assuming we aren't intent on deprecation.   If we are intent on deprecation, then something like Marco's suggestion is fine.

edwbuck:
@kfox1111 Now that we have enough alignment to unblock this one, can you provide a code change so we can get this merged with priority?

kfox1111:
Marco has volunteered to do the agreed changes. Thanks Marco! :)

marcofranssen:
> Marco has volunteered to do the agreed changes. Thanks Marco! :)

? 🤔

kfox1111:
> > Marco has volunteered to do the agreed changes. Thanks Marco! :)
> 
> ? thinking

This is what you wrote to me on slack. Did I misunderstand? I'm sorry if I did. I don't want to be putting words in other peoples mouths.

```
Marco Franssen
https://github.com/spiffe/helm-charts/pull/245 for this one, if we can add a deprecated message to version and do a fallback to version in the macro for now
we will remove it later in 2 or 3 releases
I’m also happy to update on your behalf to take away some load
```

It sounds good to me.

charts/spire/charts/spire-agent/values.yaml
```diff
@@ -86,7 +88,10 @@ waitForIt:
     repository: chainguard/wait-for-it
     # -- The image pull policy
     pullPolicy: IfNotPresent
-    version: latest-20230113
+    # -- This value is deprecated in favor of tag. (Will be removed in a future release)
+    version: ""
+    # -- Overrides the image tag
+    tag: latest
```
mrsabath, line 94: I am not in favor using "latest" tag since they can change the content without our control, but in this specific scenario I assume the `chainguard` would not do anything that dramatic to break the test, so it's acceptable, I guess...

charts/spire/charts/spire-server/values.yaml
```diff
@@ -229,7 +233,10 @@ controllerManager:
         repository: rancher/kubectl
         # -- The image pull policy
         pullPolicy: IfNotPresent
+        # -- This value is deprecated in favor of tag. (Will be removed in a future release)
         version: ""
+        # -- Overrides the image tag
+        tag: latest
```
mrsabath, line 239: same as before, but acceptable

mrsabath (approved):
Thank you for cleaning up and providing a easy migration process. I reviewed the PR and it looks correctly to me. I commented on a nit related to usage of `latest` tag, but it's not a show stopper.

marcofranssen:
> Thank you for cleaning up and providing a easy migration process. I reviewed the PR and it looks correctly to me. I commented on a nit related to usage of `latest` tag, but it's not a show stopper.

Handled those, good catch, copy paste 😞

marcofranssen:
> This looks very good but has one issue I can see. Due to a merge happening this line needs an additional check to look for tag as well as version: [`main`/charts/spire/charts/spire-server/templates/_helpers.tpl#L110](https://github.com/spiffe/helm-charts/blob/main/charts/spire/charts/spire-server/templates/_helpers.tpl?rgh-link-date=2023-05-17T15%3A52%3A36Z#L110)

@kfox1111 Would this be sufficient?

https://github.com/spiffe/helm-charts/pull/245/files#diff-f9ef77203566edfa336f7bcc66f084886b74cabff3fb5af5f64b13b3cb2deb3dR110-R112

kfox1111 (commented):
LGTM

faisal-memon (approved):

PR #246, opened 2023-04-28
Ship the examples/profiles

kfox1111:
Rework examples and profiles so that they are shipped with the helm chart so the recommendations match the chart version.


kfox1111:
Found out that all files shipped in the package are also uploaded to k8s in the release secret, bloating it. Therefore, only the profiles should be included, but no documentation/examples. This needs to be reworked a bit. marking as draft for now.

README.md
```diff
@@ -1,3 +1,5 @@
+> **Note**: We do not recommend using the charts in this repository directly except for development. Instead, please see https://artifacthub.io/packages/search?repo=spiffe
+
 > **Note**: All the helm charts in this repo are beta. We encourage you to try them out and contribute. The API may change as we move towards a production ready release.
```
edwbuck, line 4: I think we're putting things in the wrong order, It's beta software, and that's why you don't use them directly.
```suggestion
**Note**: We do not recommend using the charts in this repository directly. Instead, please see https://artifacthub.io/packages/search?repo=spiffe

```

README.md
```diff
@@ -1,3 +1,5 @@
+> **Note**: We do not recommend using the charts in this repository directly except for development. Instead, please see https://artifacthub.io/packages/search?repo=spiffe
```
edwbuck, line 1: Remove what to do here, so we can say "this is beta" and THEN "don't directly link"

```suggestion
```

charts/spire/README.md
```diff
@@ -47,6 +47,21 @@ spec:
         - --service-account-signing-key-file=/run/config/pki/sa.key
 ```
 
+## Deployment - Non Production
```
edwbuck, line 50: Reorder this to lead with the approach we want people to take first.

This means that we want the "Production" deployment first, because that's our mission to deliver, and we want people to do it this way.

Which brings the question, do we even want to advertise a different way?  Do we need to?  Savvy developers of Helm will know what to do if we don't tell them, and the dangers that come with their own choices.

We should be very careful advertising how to do something we don't really want to support.

charts/spire/README.md.gotmpl
```diff
@@ -49,6 +49,21 @@ spec:
         - --service-account-signing-key-file=/run/config/pki/sa.key
 ```
 
+## Deployment - Non Production
```
edwbuck, line 52: Same commentary as before (but in the template)

I don't think we should lead with our least desirable deployment approach, we want to bury the items we don't want to be the primary approach after the first option.

charts/spire/examples/production/README.md
```diff
@@ -0,0 +1,36 @@
+# Recommended production setup
+
+> **Note**: You should use all these files from your particular helm chart release. Don't use directly from the upstream
```
edwbuck, line 3: Avoid using directly?  These are instructions.  Let's avoid being vague.  If possible let's give them the commands, or at least command for "an example" of how to avoid what we're asking them to avoid.

charts/spire/examples/production/your-values.yaml
```diff
@@ -0,0 +1,20 @@
+# These values should be in your own configuration values.yaml
```
edwbuck, line 1: I would prefer we had comments, perhaps with "# CHANGE_ME: " labels on what to update, so we can give guidance on what to set

charts/spire/examples/production/your-values.yaml
```diff
@@ -0,0 +1,20 @@
...
+    trustDomain: changeme.yourdomain
+
+spire-server:
+  jwtIssuer: oidc-discovery.changeme.yourdomain
+  ca_subject:
+    country: NL
+    organization: ChangeMe
+    common_name: chageme.yourdomain
+
+# If using spiffe-oidc-discovery-provider
```
edwbuck, line 16: I would prefer that we have the oidc-discovery-provider disabled by default for a generalized production example.

charts/spire/profiles/production-values.yaml
```diff
@@ -1,3 +1,6 @@
+# NOTE: Do not edit the contents of this file. We will update it as we find
```
edwbuck, line 1: This is an odd request to make.  Can you share the motivation behind the request.  With that, I can better review it.

edwbuck (changes requested):

edwbuck:
@kfox1111 Please clean up this PR, or decide it's time to retire it.  It's been in draft status for three weeks.

Personally, I think shipping the examples in some manner is a good idea, but that "some manner" might just be a "user manual / howto guide".  Without such a guide, it seems like concrete examples are only going to be useful to those who already understand our environment / approach.

kfox1111:
Its on my todo list. We've been getting lots of other pr's updated/merged, and this one hasn't been the highest priority in the last few days. Its marked as Draft to deal with the fact it bitrotted a bit. Please give it a little time to make it back to the top of the pile. it should be soon at this rate.

Thanks for the reviews. I appreciate it.

kfox1111:
Its only been in draft status for a week due to https://github.com/spiffe/helm-charts/pull/246#issuecomment-1540814092

edwbuck:
> Its only been in draft status for a week due to [#246 (comment)](https://github.com/spiffe/helm-charts/pull/246#issuecomment-1540814092)

Ok.  I guess I imagined it was in draft status from the beginning.    The reason I'm mentioning this one is more because I have a goal of reviewing all the PRs with an eye to get the number down to 15.  Yes, that was 20 just a week ago, but that's the idea, fewer PRs in the pipeline (sometimes) helps efficiency.

I'm ok with a little waiting.

kfox1111:
This needs such a thorough reworking I think I'm just going to close this one for now and restart in a new pr.

Issue #221, opened 2023-04-17
Cron job to look for out of date images


kfox1111:
We have some images that are pinned but need to be updated periodically such as:
waitForIt.image.version: "latest-20230113"

We should have a job look for newer versions and submit pr's to bump the tags as appropriate.

kfox1111:
latest-xxxx in chainguard seems to be just a nightly build. So its always going to be changing. Probably not the best candidate. Need to find a more versioned tag.

Can see it with something like:
```
podman run -it --rm bitnami/oras:latest repo tags cgr.dev/chainguard/bash | grep '^latest-'
```

kfox1111:
Fetching the sbom:
```bash
cosign verify-attestation --certificate-identity https://github.com/chainguard-images/images/.github/workflows/release.yaml@refs/heads/main --certificate-oidc-issuer https://token.actions.githubusercontent.com --type spdx cgr.dev/chainguard/bash:latest | jq  '.payload | @base64d | fromjson.predicate'
```

kfox1111:
Finally figured out why I wasn't getting good info in the sbom. its because its multiarch. Pulling the sbom for a platform works better:
cosign download sbom cgr.dev/chainguard/bash:latest-20230425 --platform linux/amd64

kfox1111:
Ok. Looking at the changes in bash over time, its constantly changing. Most of the time its changing due to it including a reference to the github repo hash. If we throw that away, we get:
```
[kfox@zathras diff]$ grep 'Version changes:' * | grep -v 'Version changes:  1' | grep -v 'Version changes:  0'
latest-20230307.json-latest-20230308.json:Version changes:  3
latest-20230317.json-latest-20230319.json:Version changes:  3
latest-20230320.json-latest-20230321.json:Version changes:  3
latest-20230323.json-latest-20230324.json:Version changes:  3
latest-20230325.json-latest-20230326.json:Version changes:  3
latest-20230329.json-latest-20230330.json:Version changes:  4
latest-20230411.json-latest-20230412.json:Version changes:  3
latest-20230412.json-latest-20230413.json:Version changes:  4
latest-20230413.json-latest-20230414.json:Version changes:  3
latest-20230420.json-latest-20230421.json:Version changes:  4
latest-20230421.json-latest-20230422.json:Version changes:  2
latest-20230424.json-latest-20230425.json:Version changes:  2
```

So, sometimes it is a week or so between changes, and sometimes a few days. Still very frequent. :/

[...]

PR #249, opened 2023-04-29
Cron job to check for and update images

kfox1111:
This patch adds a cron job to check images for updates and propose a PR with updated versions when one is found and a new version is available.

Running it, it already discovers 3 updates at the time of this writing.

It does have an issue with reformatting comments and removing white space. We may want to just go along with the reformatting the first time.

fixes: https://github.com/spiffe/helm-charts/issues/221


kfox1111:
@marcofranssen Its failing shellcheck with:
./.github/scripts/update-tags.sh:38:85: note: Double quote to prevent globbing and word splitting. [SC2086]

But the non double quoted variable substitution is actually critical to its proper functioning.

kfox1111:
Fixed

.github/workflows/check-versions.yaml
```diff
@@ -32,6 +32,25 @@ jobs:
...
+
+      - name: Update image tags
+        run: |
+          go install github.com/mikefarah/yq/v4@latest
+
+          sudo apt-get install wget apt-transport-https gnupg lsb-release
+          wget -qO - https://aquasecurity.github.io/trivy-repo/deb/public.key | sudo apt-key add -
+          echo deb https://aquasecurity.github.io/trivy-repo/deb $(lsb_release -sc) main | sudo tee -a /etc/apt/sources.list.d/trivy.list
+          sudo apt-get update
+          sudo apt-get install trivy
```
marcofranssen, line 49: There is a trivy action

https://github.com/aquasecurity/trivy-action

.github/workflows/check-versions.yaml
```diff
@@ -32,6 +32,25 @@ jobs:
           ./.github/scripts/update-versions.sh
           git diff
 
+      - name: Setup go
+        uses: actions/setup-go@v4
```
marcofranssen, line 36: Can we lock the action on the patch release.

```suggestion
        uses: actions/setup-go@v4.0.0
```

.github/workflows/check-versions.yaml
kfox1111: The action runs trivy itself. We need to script the execution of it. Not sure how to do that with the upstream action.

.github/workflows/check-versions.yaml
kfox1111: good idea.

.github/workflows/check-versions.yaml
faisal-memon: +1 for the trivy action if possible.

kfox1111: I looked into this some more. I don't think we can use the action:

1. It doesn't support just the install of trivy. its an action to run trivy.
2. We would no longer be able to just run the script locally on our own machines
3. We'd have to change things to be much much more complicated. It would have to do a dynamic matrix job that then runs all the trivy actions, then somehow tell it to not block on action failures, then run another job gathering up all the results of the trivy actions to do the building of the changes & pr.

So, might technically be possible to do, but feels undesirable.

kfox1111: I've been running this script manually since its not merged for a while now and its been very handy to be able to run it outside of gate.

.github/scripts/update-tags.sh
```diff
@@ -0,0 +1,63 @@
...
+    QUERY=$(jq -r ".\"${CHART}\"[${IDX}].query" "$IMAGEJSON")
+    FILTER=$(jq -r ".\"${CHART}\"[${IDX}].filter" "$IMAGEJSON")
+    SORTFLAGS=$(jq -r ".\"${CHART}\"[${IDX}].\"sort-flags\"" "$IMAGEJSON")
+
+    VALUES="${SCRIPTPATH}/../../charts/spire/charts/${CHART}"
+    REGISTRY=$(yq e ".${QUERY}.registry" "$VALUES")
+    REPOSITORY=$(yq e ".${QUERY}.repository" "$VALUES")
+    VERSION=$(yq e ".${QUERY}.tag" "$VALUES")
+    # shellcheck disable=SC2086
+    LATEST_VERSION=$(crane ls "${REGISTRY}/${REPOSITORY}" | grep "${FILTER}" | sort ${SORTFLAGS}| tail -n 1)
```
marcofranssen, line 49:
```suggestion
    LATEST_VERSION=$(crane ls "${REGISTRY}/${REPOSITORY}" | grep "${FILTER}" | sort "${SORTFLAGS[@]}"| tail -n 1)
```

This might do the trick without disabling the linting. Not sure though in this usecase.

.github/workflows/check-versions.yaml
```diff
@@ -2,7 +2,7 @@ name: Check versions
 
 on:
   schedule:
-    - cron: '0 8 * * 1'
+    - cron: '0 8 * * *'
```
marcofranssen, line 5: Do we really want to increase the interval? Once a week seems sufficient to me.

.github/scripts/update-tags.sh
kfox1111: No. the nonquoting is critical to its functioning. multiple args are passed through the variable and have to be handled separately by bash. The linter is trying to catch a common case of accidental misconfiguration but in this case, it really is intentional.

faisal-memon: Can we put in a comment on why we have to disable the linter here?

kfox1111: Done

.github/scripts/update-tags.sh
```diff
@@ -0,0 +1,64 @@
...
+
+    if trivy image "${REGISTRY}/${REPOSITORY}:${VERSION}" --exit-code 1; then
+      echo No CVE found. Skipping.
+      continue
+    fi
+
+    if [ "${VERSION}" != "${LATEST_VERSION}" ]; then
+      echo "New image version found: ${REGISTRY}/${REPOSITORY}:${LATEST_VERSION}"
+      python3 -c "import sys; from dict_deep import deep_set; import ruamel.yaml; y = ruamel.yaml.YAML(); y.indent(mapping=2, sequence=4, offset=2); y.preserve_quotes = True; d = y.load(open('${VALUES}')); deep_set(d, '${QUERY}.tag', '${LATEST_VERSION}'); y.dump(d, sys.stdout);" > /tmp/$$
```
faisal-memon, line 59: Is it better to put this in a file or at least have the statements on different lines so its easier to read?

kfox1111:
Fixed.

.github/scripts/update-tags.sh
kfox1111: > This might do the trick without disabling the linting. Not sure though in this usecase.

I tried it again just to be sure... the problem is the variable is a string with spaces, not an array, so the [@] form doesn't work.

.github/workflows/check-versions.yaml
kfox1111: I'll revert it because weekly is still better then random, which we are doing now, and there is contention.

but, how do we know we have cve's without someone looking? This running is the thing that looks and lets us know. So if its weekly, we're only looking for cve fixes weekly. I'd rather know sooner.

The resources for the job are quite minimal, so its not a very high cost.

.github/scripts/update-tags.sh
kfox1111: Will try.

kfox1111: same issue. its trying to build one array argument out of a string intended to hold multiple:
sort: multi-character tab ‘ . -k1,1n -k2,2n -k3,3n’

I believe unquoted really is the correct usage.

You can test for yourself by setting spire-agent.waitForIt.image.tag  to latest-20220517
and running ./.github/scripts/update-tags.sh

edwbuck: Both of these approaches work, and both work without shellcheck reporting an error, or needing quoting disabled.

```
SORTFLAGS=( "-k1,1n" "-k2,2n" "-k3,3n" )

sort "${SORTFLAGS[@]}" << 'EOF'
a, b, c, 1
a, c, c, 2
a, b, d, 3
EOF

printf 'a, b, c, 1\na, c, c, 2\na, b, d, 3\n' | sort "${SORTFLAGS[@]}"
```

The idea that "${VARIABLE[@]}" expands to "-k1,1n -k2,2n -k3,3n" isn't true.  For arrays, expansion generally puts the double quotes around each element of an array expansion "-k1,1" "-k2,2" "-k3,3".

to illustrate, I use this bit of code to dump an array.  (It's from a shell library project I wrote, to manage function extensions in an "object_oriented" like manner)

```
function dump_object() {
        declare OBJECT_VAR=$1
        declare -n OBJECT_REF=$1

        SORTED_KEYS=()
        PREVIOUS_IFS=${IFS}
        while IFS= read -rd '' KEY; do
                SORTED_KEYS+=( "${KEY}" )
        done < <(printf '%s\0' "${!OBJECT_REF[@]}" | sort -z)
        for KEY in ${SORTED_KEYS[*]}; do
                echo ${OBJECT_VAR}[$KEY]=${OBJECT_REF[${KEY}]}
        done
        IFS=${PREVIOUS_IFS}
}
```

to use it one does
```
SORTFLAGS=( "-k1,1n" "-k2,2n" "-k3,3n" )
dump_object SORTFLAGS
```

to get 
```
SORTFLAGS[0]=-k1,1n
SORTFLAGS[1]=-k2,2n
SORTFLAGS[2]=-k3,3n
```

in the definition of SORTFLAGS above, the double quoting in the sort works perfectly.

```
#!/bin/bash

function dump_object() {
        declare OBJECT_VAR=$1
        declare -n OBJECT_REF=$1

        SORTED_KEYS=()
        PREVIOUS_IFS=${IFS}
        while IFS= read -rd '' KEY; do
                SORTED_KEYS+=( "${KEY}" )
        done < <(printf '%s\0' "${!OBJECT_REF[@]}" | sort -z)
        for KEY in ${SORTED_KEYS[*]}; do
                echo ${OBJECT_VAR}[$KEY]=${OBJECT_REF[${KEY}]}
        done
        IFS=${PREVIOUS_IFS}
}

SORTFLAGS=( "-k1,1n" "-k2,2n" "-k3,3n" )

dump_object SORTFLAGS

sort "${SORTFLAGS[@]}" << 'EOF'
a, b, c, 1
a, c, c, 2
a, b, d, 3
EOF
```

but if I do
```
BAD_SORTFLAGS=( "-k1,1n -k2,2n -k3,3n" )

dump_object BAD_SORTFLAGS
```
which shows as output
```
BAD_SORTFLAGS[0]=-k1,1n -k2,2n -k3,3n
```

Then running
```
sort "${BAD_SORTFLAGS[@]}" << 'EOF'
a, b, c, 1
a, c, c, 2
a, b, d, 3
EOF
```
or
```
printf 'a, b, c, 1\na, c, c, 2\na, b, d, 3\n' | sort "${SORTFLAGS[@]}"
```
doesn't work but the unquoted versions
```
sort ${BAD_SORTFLAGS[@]} << 'EOF'
a, b, c, 1
a, c, c, 2
a, b, d, 3
EOF
```
and
```
printf 'a, b, c, 1\na, c, c, 2\na, b, d, 3\n' | sort ${SORTFLAGS[@]}
```
do work.

Odds are you are misidentifying the root cause.  It's not because the quotes around ${SORTFLAGS[@]} are combining the parameters, it's because you only set one value in ${SORTFLAGS[@]} that has spaces separating the text within that single value.

Please never disable the quoting rules in shellcheck, of all the rules, it's the most important one.

.github/scripts/update-tags.sh
edwbuck, line 1: Need to check with the object_dump the actual values within SORTFLAGS, which are probably not set the way they are thought to be set, which is why the removed quoting works when the proper quoting doesn't.

edwbuck (changes requested):

.github/scripts/update-tags.sh
kfox1111: The source of the string is not a bash array though, but a json document. How do you suggest we pull it into a bash array?

kfox1111: Its defined in the json config like:
https://github.com/spiffe/helm-charts/pull/249/files#diff-4c6c57c7cd483f8116b836fc3762a95f0f347b5f1f5e4cb67002f247d50ff14bR24

Then pulled into bash here:
https://github.com/spiffe/helm-charts/pull/249/files#diff-e800c023ce454983c68686a85f2ea49f6114c2cd433a239eb9637c84ae3ddeefR42

There could be a better way to structure that, but no idea what that should look like

marcofranssen:
Thinking again on this, isn't it better if the upstream project checks for vulnerabilities in the images? Maybe contribute it there on the images we are using. 

In the end it doesn't help us to detect the vulnerability if the upstream doesn't make a release. I think we should just check for new version like dependabot does. Checking for vulnerabilities seems to be something that shouldn't be part of this helm-charts repo, but should be handled in the upstream projects we depend on (hence why I advised to use chainguard images), then we can have a job that checks for new releases.

IMHO we should close this PR as image scanning shouldn't be our responsibility as in the end it doesn't give us value in the sense of having patches for those things we detect.

faisal-memon:
> Thinking again on this, isn't it better if the upstream project checks for vulnerabilities in the images? Maybe contribute it there on the images we are using.
> 
> In the end it doesn't help us to detect the vulnerability if the upstream doesn't make a release. I think we should just check for new version like dependabot does. Checking for vulnerabilities seems to be something that shouldn't be part of this helm-charts repo, but should be handled in the upstream projects we depend on (hence why I advised to use chainguard images), then we can have a job that checks for new releases.
> 
> IMHO we should close this PR as image scanning shouldn't be our responsibility as in the end it doesn't give us value in the sense of having patches for those things we detect.

I agree with the above. In the end if we are just staying up to date on the latest versions then we are just as good. The code to do that should be much simpler. I spent a good chunk of time reviewing but could not wrap my head around the scripts. I wish I had your scripting skills @kfox1111. 

With so many other pr's to review i'm not sure how much more time i can spend on this one.

kfox1111:
> Thinking again on this, isn't it better if the upstream project checks for vulnerabilities in the images? Maybe contribute it there on the images we are using.
> 
> In the end it doesn't help us to detect the vulnerability if the upstream doesn't make a release. I think we should just check for new version like dependabot does. Checking for vulnerabilities seems to be something that shouldn't be part of this helm-charts repo, but should be handled in the upstream projects we depend on (hence why I advised to use chainguard images), then we can have a job that checks for new releases.
> 
> IMHO we should close this PR as image scanning shouldn't be our responsibility as in the end it doesn't give us value in the sense of having patches for those things we detect, those patches will only arrive if the upstream detects them, hence why I advise to contribute this over there in the upstream projects.

Chainguard releases daily. So if we do what you suggest, we will be patching our charts, daily. The container scanner part of the pr does just one thing. Squash patching version when no cve's are fixed so we're not having to cut releases constantly to deal with the fact that chainguard releases daily and we can't easily tell which ones are important to update (actual cve's fixed) vs just a new daily image was built. We could consider not doing the trivy stuff if we stopped using chainguard as they are the only images doing such constant releases.

kfox1111:
Are we trying to throw the baby out with the bathwater? When I joined up, I think the only one that knew how the .github tests worked was Marco. It was ok. Things were better with the tests then without them.

Perhaps this is in that category. We are only better off with this in place then without it. That not everyone knows all the gory details of how it works in its entirety, is not great, but we're still better off with it IMO.

Based on how complex it became in bash, its probably better to rewrite this in something more well known such as python. I propose though that we can merge this, as is, and when other, more pressing issues are out of the way, we can do that then. I'm afraid we're letting the perfect stand in the way of the good enough here.

kfox1111:
Ok. With Edwin's help, we got it converted over to read into a bash array. This fixes the shellcheck issue so we don't have to squash it. Thanks Edwin!

edwbuck (approved):
Worked with Kevin to get all BASH passing shellcheck.

Led to actually documenting the different flags in the input YAML as different flags (previously there were not an array, but a string).

Led to changes to pull each array item into a bash array,

Led to the ability to properly quote that array on expansion.

edwbuck:
@faisal-memon @marcofranssen @mrsabath Can we get a second review here?  This is nearly ready to merge.

faisal-memon:
> @faisal-memon @marcofranssen @mrsabath Can we get a second review here?  This is nearly ready to merge.

@edwbuck Any thoughts on the concerns posted earlier and the alternatives presented?

kfox1111:
Anything left to do?

faisal-memon:
@kfox1111 I appreciate proactively focusing on security. However there are alot of concerns presented above with this pr. What value is the trivy scanning providing vs just keep versions up to date?

kfox1111:
> @kfox1111 I appreciate proactively focusing on security. However there are alot of concerns presented above with this pr. What value is the trivy scanning providing vs just keeping versions up to date?

I already explained that here:
https://github.com/spiffe/helm-charts/pull/249#issuecomment-1582629766

Please respond to that along with this: 
https://github.com/spiffe/helm-charts/pull/249#issuecomment-1582770047

.github/workflows/check-versions.yaml
```diff
@@ -32,6 +32,30 @@ jobs:
...
+      - name: Update image tags
+        run: |
+          go install github.com/mikefarah/yq/v4@latest
+
+          sudo apt-get install wget apt-transport-https gnupg lsb-release
+          wget -qO - https://aquasecurity.github.io/trivy-repo/deb/public.key | sudo apt-key add -
+          echo deb https://aquasecurity.github.io/trivy-repo/deb $(lsb_release -sc) main | sudo tee -a /etc/apt/sources.list.d/trivy.list
+          sudo apt-get update
+          sudo apt-get install trivy
+          sudo pip install ruamel.yaml dict_deep
```
marcofranssen, line 54: Shouldn't we use a requirements.txt so we can configure dependabot to keep these deps updated? This is likely to break if those deps change.

https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#package-ecosystem

.github/workflows/check-versions.yaml
```diff
@@ -32,6 +32,30 @@ jobs:
           ./.github/scripts/update-versions.sh
           git diff
 
+      - name: Setup go
+        uses: actions/setup-go@v4.0.0
+
+      - name: Setup crane
+        uses: imjasonh/setup-crane@v0.1
+
+      - uses: actions/setup-python@v4
```
marcofranssen, line 41: Lock on the patch release like we do with other actions.

.github/workflows/check-versions.yaml
```diff
@@ -32,6 +32,30 @@ jobs:
           ./.github/scripts/update-versions.sh
           git diff
 
+      - name: Setup go
+        uses: actions/setup-go@v4.0.0
```
marcofranssen, line 36: Why do we need Go in this pipeline? Doesn't seem like we need to compile any Go here.

.github/workflows/check-versions.yaml
```diff
@@ -32,6 +32,30 @@ jobs:
...
+      - name: Setup crane
+        uses: imjasonh/setup-crane@v0.1
+
+      - uses: actions/setup-python@v4
+        with:
+          python-version: '3.9'
+
+      - name: Update image tags
+        run: |
+          go install github.com/mikefarah/yq/v4@latest
```
marcofranssen, line 47: Since we are doing a python script anyhow, why not resolve this in python? so we don't need to do the yq part.

.github/tests/images.json
```diff
@@ -0,0 +1,46 @@
+{
```
marcofranssen, line 1: Some newly added images are missing here.

.github/tests/images.json
```diff
@@ -0,0 +1,46 @@
...
+      "filter": "^[0-9]\\+\\.[0-9]\\+\\.[0-9]\\+-alpine$",
+      "sort-flags": []
+    }, {
+      "query": "telemetry.prometheus.nginxExporter.image",
+      "filter": "^[0-9]\\+\\.[0-9]\\+\\.[0-9]\\+$",
+      "sort-flags": ["-t", ".", "-k1,1n", "-k2,2n", "-k3,3n"]
+    },
+    {
+      "query": "tests.bash.image",
+      "filter": "^latest-",
```
marcofranssen, line 35: We switched to use the semver tag here. So this also abstracts away the need of daily patching.

.github/tests/images.json
```diff
@@ -0,0 +1,46 @@
...
+      "sort-flags": []
+    }
+  ],
+  "tornjak-frontend/values.yaml": [
+    {
+      "query": "tests.bash.image",
```
marcofranssen, line 41: We switched to use the semver tag here. So this also abstracts away the need of daily patching.

.github/tests/images.json
```diff
@@ -0,0 +1,46 @@
+{
+  "spire-server/values.yaml": [
+    {
+      "query": "tests.bash.image",
+      "filter": "^latest-",
```
marcofranssen, line 5: We switched to use the semver tag here. So this also abstracts away the need of daily patching.

marcofranssen (commented):
I gave it one more review, still think this is a lot of hassle to keep things patched, while we actually should rely on the upstream projects to provide those patches.

Also see the inline comments on the bash image. The chart now uses the semver version which is not releasing daily. It releases whenever there is a new patch for bash, meaning only when a CVE is resolved.

So I still heavily doubt we need all of this stuff in our charts repo. In the end we can't patch anyhow if there is no patch released.

marcofranssen (changes requested):
I gave it one more review, still think this is a lot of hassle to keep things patched, while we actually should rely on the upstream projects to provide those patches.

Also see the inline comments on the bash image. The chart now uses the semver version which is not releasing daily. It releases whenever there is a new patch for bash, meaning only when a CVE is resolved.

So I still heavily doubt we need all of this stuff in our charts repo. In the end we can't patch anyhow if there is no patch released.

marcofranssen (changes requested):
I gave it one more review, still think this is a lot of hassle to keep things patched, while we actually should rely on the upstream projects to provide those patches.

Also see the inline comments on the bash image. The chart now uses the semver version which is not releasing daily. It releases whenever there is a new patch for bash, meaning only when a CVE is resolved.

So I still heavily doubt we need all of this stuff in our charts repo. In the end we can't patch anyhow if there is no patch released.

.github/workflows/check-versions.yaml
kfox1111: We could. Where should we put the file so that dependabot could find them though?

.github/workflows/check-versions.yaml
kfox1111: For crane, they only have major.minor

.github/workflows/check-versions.yaml
kfox1111: left over from yq I think. Will remove.

kfox1111: oh. yq is still used.

.github/workflows/check-versions.yaml
kfox1111: See comment about maybe rewriting this whole feature to be all python later. It started life in bash/yq and then gained python just to do the editing without messing up the comments.

.github/tests/images.json
kfox1111: Not sure which ones you mean? the kubectl ones don't have versions pinned.

.github/tests/images.json
kfox1111: can switch to semver for bash. waitForIt has no non latest tags.

kfox1111: Have a look at https://github.com/spiffe/helm-charts/pull/356#pullrequestreview-1484828752
I think we probably want to stick with latest variants because they are immutable.

.github/tests/images.json
kfox1111: same

.github/tests/images.json
kfox1111: same

kfox1111:
> I gave it one more review, still think this is a lot of hassle to keep things patched, while we actually should rely on the upstream projects to provide those patches.

How do we do that without actually looking for updated versions? If we can find a way to use versioned tags with chainguard we could drop the need for the trivy scanner from the script. That doesn't negate the need for the rest of the script?

> Also see the inline comments on the bash image. The chart now uses the semver version which is not releasing daily. It releases whenever there is a new patch for bash, meaning only when a CVE is resolved.

see https://github.com/spiffe/helm-charts/pull/356#pullrequestreview-1484828752

> So I still heavily doubt we need all of this stuff in our charts repo. In the end we can't patch anyhow if there is no patch released.

How do we know we need to patch without this script? Are you looking? I see, just manually running the script, there is another image needing an update. I don't know how long its been needed as its not running periodically and I"m not checking daily.

faisal-memon:
> Chainguard releases daily. So if we do what you suggest, we will be patching our charts, daily. 

We can just run the version check cron once a week, so we wont be patching daily.

> So I still heavily doubt we need all of this stuff in our charts repo. In the end we can't patch anyhow if there is no patch released.

Agree. 

> How do we know we need to patch without this script? Are you looking?

Same as we do with dependabot. Run a simple version checker once a week.

kfox1111:
> Same as we do with dependabot. Run a simple version checker once a week.

How is that not this script, minus the few lines that do the trivy run?

faisal-memon:
> How is that not this script, minus the few lines that do the trivy run?

Its the trivy part that we are both questioning. New versions of images with critical bug fixes but no cve's will be filtered out by it. So we will still need to manually check for new images anyways.

kfox1111:
> > How is that not this script, minus the few lines that do the trivy run?
> 
> Its the trivy part that we are both questioning. New versions of images with critical bug fixes but no cve's will be filtered out by it. So we will still need to manually check for new images anyways.

Ok, it sounds like we may be closer then I was thinking. I was confused when @marcofranssen  said:
> IMHO we should close this PR as image scanning shouldn't be our responsibility

And you seemed to agree with him.

The trivy part is just a few lines. I'm ok getting rid of it entirely, IF,
waitForIt actually cuts a versioned release and we can figure out how to get immutable tags from chainguard. Until then, the script without trivy would basically cause a patch to be created basically every run, which would cause more work for us, and more work for end users if it didn't actually fix anything.

We also could add a flag to the config to make the cve scanner optional for all images that have a release version if that would help too?

kfox1111:
chainguard image issues:
https://github.com/chainguard-images/images/issues/881
and
https://github.com/chainguard-images/images/issues/880

881 may make 880 moot if they really only support immutable images with the latest-xxx tags.

kfox1111:
Sounds like we need to update the logic to always use :latest@sha256xxxx as per https://github.com/spiffe/helm-charts/issues/442

kfox1111:
Updated to deal with https://github.com/spiffe/helm-charts/issues/442

faisal-memon (approved):
Thanks @kfox1111 for creating this and removing the trivy scanner part.

.github/workflows/check-versions.yaml
marcofranssen: You put it in the same folder as the python script.

See:

https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file

I think it should be using pip as package manager and you can define the path to the folder containing the python code and requirements txt.

.github/tests/images.json
marcofranssen: Oh found the answer further down in this thread.

https://github.com/spiffe/helm-charts/pull/249#issuecomment-1674018842

.github/tests/images.json
kfox1111: fsGroup was added since, so I just added it here.

marcofranssen (approved):
LGTM 🚀

PR #250, opened 2023-04-30
Switch the spire tests to always run

kfox1111:
This patch sets the spire chart tests to always run. This enables changes in tests to be tested and sets a base for split out charts.

### DIFF
--- a/.github/workflows/helm-chart-ci.yaml
+++ b/.github/workflows/helm-chart-ci.yaml
@@ -127,8 +127,6 @@ jobs:
       - lint-chart
       - build-matrix
 
-    if: needs.lint-chart.outputs.changed == 'true'
-
     strategy:
       fail-fast: false
       matrix:
--- a/.github/workflows/helm-chart-ci.yaml
+++ b/.github/workflows/helm-chart-ci.yaml
@@ -92,14 +92,6 @@ jobs:
         with:
           version: ${{ env.CHART_TESTING_VERSION }}
 
-      - name: Run chart-testing (list-changed)
-        id: list-changed
-        run: |
-          changed=$(ct list-changed --target-branch ${{ github.base_ref }})
-          if [[ -n "$changed" ]]; then
-            echo "changed=true" >> $GITHUB_OUTPUT
-          fi
-
       - name: Run chart-testing (lint)
         run: |
           ct lint --debug ${{ github.base_ref != 'release' && '--check-version-increment=false' || '' }} \
@@ -113,8 +105,6 @@ jobs:
     runs-on: ubuntu-22.04
 
     needs: [lint-chart]
-
-    if: needs.lint-chart.outputs.changed == 'true'
 
     steps:
       - name: Checkout
@@ -154,8 +144,6 @@ jobs:
           - v1.23.17
           - v1.22.17
           - v1.21.14
-        values:
-          - ${{ fromJson(needs.build-matrix.outputs.tests) }}
 
     steps:
       - name: Checkout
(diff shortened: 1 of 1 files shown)


.github/workflows/helm-chart-ci.yaml
```diff
@@ -114,8 +106,6 @@ jobs:
 
     needs: [lint-chart]
 
-    if: needs.lint-chart.outputs.changed == 'true'
```
mrsabath, line 117: we could always reverse it, to always run by default and have option not to run when not needed, but I can't think of the case that it would support that scenario

marcofranssen, line 117: I was thinking on running with something like this.

```
if: needs.lint-chart.outputs.changed == 'true' || needs.diff.outputs.workflow-changed || needs.diff.outputs.ci-scripts-changed || needs.diff.outputs.tests-changed
```

or maybe more simple a single output from a diffs job.

```
if: needs.lint-chart.outputs.changed == 'true' || needs.diff.outputs.requires-test-run
```

That way it only runs when required.

kfox1111: Yeah. was looking for a case but hadn't found one, so decided to just remove it all for now.

edwbuck: I think it is acceptable to always check the lint-chart output, even if the output doesn't change.  The primary concerns I have for skipping them are

- In a "skipping" configuration, there are always two reasons the tests were skipped; the tests would have reported the same results AND the tests crashed.  Inability to differentiate the two can hide issues in the testing framework until long after the testing framework was modified.
- Should we want to aggregate pass / fail statistics in testing, the counts will vary widely due to "skipping" tests.  This makes future analysis difficult, as to "smooth" the time series graph, one needs to resort to one of the following undesirable choices
  - Add in the "skipped" test count in scenarios where the tests were missing, to bring the number up to it's "if it was tested" count. (possibly creating a false report, if the testing framework was updated, breaking the tests)
  - Don't add in the skipped test counts, creating spikes in testing that could be misinterpreted as temporary improvements, or as noise, that needs to be filtered in processing quality metrics trends.

If this testing is very expensive, then please provide feedback so my suggestions above don't foul the production pipeline.

.github/workflows/helm-chart-ci.yaml
```diff
@@ -209,6 +197,7 @@ jobs:
             "${TEST_DIR}/install.sh"
           else
             ct install --debug \
+              --charts "charts/spire" \
```
edwbuck, line 200: Can I get a simple explanation of why this line needed to be added to "always run all the tests"?

Seems like if it is necessary, there was / is an undocumented bug in our testing.

edwbuck: From a chat conversation with Kevin, without the `--charts` parameter, the software itself determines that the charts are not requiring an update, so the tests would be effectively skipped without this extra parameter

Closing this question as resolved.

edwbuck (approved):
I'm happy with the changes

The questions I asked were answered satisfactorily.  There will still be a VM spin up to run these tests, but in the grand scheme of things, until this starts costing us human manpower (or time that violates a SLA of some sort) I think we can afford to test each time.

edwbuck:
I'm keeping this open until we get Mariusz to weigh in or at least a second approval.

.github/workflows/helm-chart-ci.yaml
mrsabath: My only concern was that over time the portfolio of test would be growing and that means the duration for the green checkmark would be getting longer, even if there is a trivial change (e.g. spelling correction). But I support your point of view as well.

mrsabath (approved):
LGTM!

faisal-memon (approved):
Thanks @kfox1111 for filling this hole, @edwbuck and @mrsabath for reviewing.

faisal-memon (approved):
Thanks @kfox1111 for filling this hole, @edwbuck and @mrsabath for reviewing.

PR #255, opened 2023-05-03
Use separate secret for db password

kfox1111:
This patch makes the external database config store just the password in its own Kubernetes secret.


marcofranssen (approved):
LGTM 🚀 I'll leave merging to you, not at laptop atm. I'll check tomorrow for the other PR and handle merging if not done yet. 🙏 thanks great work.

PR #262, opened 2023-05-05
Add project glossary.

### LINKED ISSUE #261: Add a project glossary
We need a glossary to document and guide the project's terminology in ways that lead use to using a common language to describe the same items that permit descriptions with different words or word phrases.

edwbuck:
This glossary is not meant to be complete, as
no glossary can be.  It does mean to cover the
basic different deployments.

To keep things clear, federation was also added
to clarify that federation is not a deployment
but a different "thing".

Closes #261

### DIFF
--- a/project/glossary.md
+++ b/project/glossary.md
@@ -15,7 +15,7 @@ Deployment Type
 : One of three supported deployments of the template: Standalone, 
 Primary, or Secondary.
 
-Federation
+**Federation**
 : When one cluster is configured to trust elements of another cluster
 containing a different trust domain.
 
--- a/project/glossary.md
+++ b/project/glossary.md
@@ -23,7 +23,7 @@ containing a different trust domain.
 : A Primary Deployment or Standalone Deployment that is configured
 to federate with one or more Deployments.
 
-Primary Deployment
+**Primary Deployment**
 : A deployment of the Helm Charts where the cluster's purpose is to
 provide certificates to Secondary Deployments.
 
--- a/project/glossary.md
+++ b/project/glossary.md
@@ -36,7 +36,7 @@ Standalone Deployment
 : A deployment of the Helm Charts where the cluster both manages the CA
 and services Workload Identity requests in the same cluster.
 
-Tiered Deployment
+**Tiered Deployment**
 : Use of the Helm Charts two or more times, such that one chart is
 configured as a Primary Deployment Type and the others are configured as
 Secondary Deployment Types.
--- a/project/glossary.md
+++ b/project/glossary.md
@@ -19,7 +19,7 @@ Primary, or Secondary.
 : When one cluster is configured to trust elements of another cluster
 containing a different trust domain.
 
-Federated Deployment
+**Federated Deployment**
 : A Primary Deployment or Standalone Deployment that is configured
 to federate with one or more Deployments.
 
--- a/project/glossary.md
+++ b/project/glossary.md
@@ -11,7 +11,7 @@ team members.
 **Deployment**
 : A use of the Helm Charts to describe a single SPIRE cluster.
 
-Deployment Type
+**Deployment Type**
 : One of three supported deployments of the template: Standalone, 
 Primary, or Secondary.
 
--- a/project/glossary.md
+++ b/project/glossary.md
@@ -12,7 +12,7 @@ team members.
 : A use of the Helm Charts to describe a single SPIRE cluster.
 
 **Deployment Type**
-: One of three supported deployments of the template: Standalone, 
+: One of three supported deployments of the charts: Standalone, 
 Primary, or Secondary.
 
 **Federation**
--- a/project/glossary.md
+++ b/project/glossary.md
@@ -41,6 +41,6 @@ and services Workload Identity requests in the same cluster.
 configured as a Primary Deployment Type and the others are configured as
 Secondary Deployment Types.
 
-Trust Domain
+**Trust Domain**
 : The host field of a SPIFFE ID, naming a minimum footprint of a trust
 bundle's distribution.
--- a/project/glossary.md
+++ b/project/glossary.md
@@ -27,7 +27,7 @@ to federate with one or more Deployments.
 : A deployment of the Helm Charts where the cluster's purpose is to
 provide certificates to Secondary Deployments.
 
-Secondary Deployment
+**Secondary Deployment**
 : A deployment of the Helm Charts where the cluster obtains an
 intermediate CA from a Primary Deployment and uses it to service
 Workload Identity requests.
--- a/project/glossary.md
+++ b/project/glossary.md
@@ -32,7 +32,7 @@ provide certificates to Secondary Deployments.
 intermediate CA from a Primary Deployment and uses it to service
 Workload Identity requests.
 
-Standalone Deployment
+**Standalone Deployment**
 : A deployment of the Helm Charts where the cluster both manages the CA
 and services Workload Identity requests in the same cluster.
 
--- a/project/glossary.md
+++ b/project/glossary.md
@@ -8,7 +8,7 @@ to keep other documents concise and precise.  Documenting terminology
 here helps prevent misundersandings, facilitating easy onboarding of new
 team members.
 
-Deployment
+**Deployment**
 : A use of the Helm Charts to describe a single SPIRE cluster.
 
 Deployment Type
--- a/project/glossary.md
+++ b/project/glossary.md
@@ -1,0 +1,46 @@
+<!-- vim: ft=markdown colorcolumn=72
+-->
+# Glossary
+
+This glossary is a dictionary of terms defined as they are used by the
+spire/helm-charts project.  Writing the definitions of terms here helps
+to keep other documents concise and precise.  Documenting terminology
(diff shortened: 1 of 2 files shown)


project/overview.md
```diff
@@ -41,3 +41,10 @@ within the primary github repository.
 The project is currently in a pre-release status.  While the standalone
 deployment is close to release quality, the tiered deployment is not.
 
+## Project Navigation
+
+Below are links to project documentation useful for the management and
+maintenance of the spire/helm project.
+
+### Coordination
+- [glossary](glossary.md)
```
marcofranssen, line 50:
```suggestion

- [glossary](glossary.md)
```

project/glossary.md
```diff
@@ -0,0 +1,46 @@
...
+-->
+# Glossary
+
+This glossary is a dictionary of terms defined as they are used by the
+spire/helm-charts project.  Writing the definitions of terms here helps
+to keep other documents concise and precise.  Documenting terminology
+here helps prevent misundersandings, facilitating easy onboarding of new
+team members.
+
+Deployment
```
marcofranssen, line 11:
```suggestion
**Deployment**
```

project/glossary.md
```diff
@@ -0,0 +1,46 @@
...
+: A use of the Helm Charts to describe a single SPIRE cluster.
+
+Deployment Type
```
marcofranssen, line 14:
```suggestion
**Deployment Type**
```

project/glossary.md
```diff
@@ -0,0 +1,46 @@
...
+: One of three supported deployments of the template: Standalone, 
+Primary, or Secondary.
+
+Federation
```
marcofranssen, line 18:
```suggestion
**Federation**
```

project/glossary.md
```diff
@@ -0,0 +1,46 @@
...
+: When one cluster is configured to trust elements of another cluster
+containing a different trust domain.
+
+Federated Deployment
```
marcofranssen, line 22:
```suggestion
**Federated Deployment**
```

project/glossary.md
```diff
@@ -0,0 +1,46 @@
...
+: A Primary Deployment or Standalone Deployment that is configured
+to federate with one or more Deployments.
+
+Primary Deployment
```
marcofranssen, line 26:
```suggestion
**Primary Deployment**
```

project/glossary.md
```diff
@@ -0,0 +1,46 @@
...
+: A deployment of the Helm Charts where the cluster's purpose is to
+provide certificates to Secondary Deployments.
+
+Secondary Deployment
```
marcofranssen, line 30:
```suggestion
**Secondary Deployment**
```

project/glossary.md
```diff
@@ -0,0 +1,46 @@
...
+: A deployment of the Helm Charts where the cluster obtains an
+intermediate CA from a Primary Deployment and uses it to service
+Workload Identity requests.
+
+Standalone Deployment
```
marcofranssen, line 35:
```suggestion
**Standalone Deployment**
```

project/glossary.md
```diff
@@ -0,0 +1,46 @@
...
+: A deployment of the Helm Charts where the cluster both manages the CA
+and services Workload Identity requests in the same cluster.
+
+Tiered Deployment
```
marcofranssen, line 39:
```suggestion
**Tiered Deployment**
```

project/glossary.md
```diff
@@ -0,0 +1,46 @@
...
+: Use of the Helm Charts two or more times, such that one chart is
+configured as a Primary Deployment Type and the others are configured as
+Secondary Deployment Types.
+
+Trust Domain
```
marcofranssen, line 44:
```suggestion
**Trust Domain**
```

marcofranssen (commented):
Few nit pick suggestions for the sake of readability.

marcofranssen (approved):
LGTM. For next time. From the files tab suggestions can be combined in a single commit. Probably best to squash this PR.

project/glossary.md
```diff
@@ -0,0 +1,46 @@
...
+to keep other documents concise and precise.  Documenting terminology
+here helps prevent misundersandings, facilitating easy onboarding of new
+team members.
+
+**Deployment**
+: A use of the Helm Charts to describe a single SPIRE cluster.
+
+**Deployment Type**
+: One of three supported deployments of the template: Standalone, 
+Primary, or Secondary.
```
kfox1111, line 16: I can think of at least one more configuration that security conscious users might want... spire-server in one (security focused) k8s cluster and the spire-agent in a workload cluster. What would we call that?

project/glossary.md
```diff
@@ -0,0 +1,46 @@
...
+# Glossary
+
+This glossary is a dictionary of terms defined as they are used by the
+spire/helm-charts project.  Writing the definitions of terms here helps
+to keep other documents concise and precise.  Documenting terminology
+here helps prevent misundersandings, facilitating easy onboarding of new
+team members.
+
+**Deployment**
+: A use of the Helm Charts to describe a single SPIRE cluster.
```
kfox1111, line 12: Since there are multiple "clusters" involved, should we call it out explicitly like done here? Throughout the rest of the document cluster is being used to talk about a SPIRE cluster and not a k8s cluster or sql cluster.

Or document Cluster as its own term? That may be better too, as then we can be explicit about what more then a multi pod spire-server makes up a "spire cluster"?

project/glossary.md
```diff
@@ -0,0 +1,46 @@
...
+
+**Primary Deployment**
+: A deployment of the Helm Charts where the cluster's purpose is to
+provide certificates to Secondary Deployments.
+
+**Secondary Deployment**
+: A deployment of the Helm Charts where the cluster obtains an
+intermediate CA from a Primary Deployment and uses it to service
+Workload Identity requests.
+
```
kfox1111, line 34: Honest question. I could go either way. Primary/Secondary totally works. Our charts are a different way of deploying spire, so different terminology may be the bast way to go to disambiguate things. But spire also has its own terminology that may be better to reuse. They would call it root and nested.

Whats the best approach there?

project/glossary.md
```diff
@@ -0,0 +1,46 @@
...
+**Standalone Deployment**
+: A deployment of the Helm Charts where the cluster both manages the CA
+and services Workload Identity requests in the same cluster.
+
+**Tiered Deployment**
+: Use of the Helm Charts two or more times, such that one chart is
```
kfox1111, line 40: A root chart is likely to contain at least a spire-server and likely a spiffe-oidc-discovery-provider

```suggestion
: Use of the Helm Charts two or more times, such that one or more charts are
```

edwbuck: I would prefer to consider "use of the Helm Charts" in a context where one "of the Charts" matches one cluster, not one use "of the charts" match one of each of the possible components in a cluster.  With the latter, undesired, interpretation, to deploy a standalone cluster, one would use the charts two or three times, which just seems an overly complex (and wrong) way to describe things.

project/glossary.md
edwbuck: Yes, we should be specific where it helps, and non-specific where it hurts.

I wouldn't want a definition of "Cluster" to be more than "A collection of related computers and software acting as part of a system"  

The alternative is "Cluster" means "A collection of related computers and software acting as part of SPIRE deployment" which would then require Kubernetes Cluster to mean one of "A Kubernetes collection of related computers and software acting as part of a SPIRE deployment" (blocking the ability to talk about Kubernetes clusters) or "A collection of related computers and software acting as part of a Kubernetes deployment" (making contextual comments of "Kubernetes Cluster" with follow on "Cluster" comments contextually awkward.

Feel free to suggest "SPIRE Cluster" / "Kubernetes Cluster" where you see a raw "Cluster" and the meaning should be constrained to only one or the other.

project/glossary.md
edwbuck: Two deployments of the charts (the charts might be identical or not, depending on how they decide to deploy)

kfox1111: How would the deployment flag be set for each case? Still standalone or in some cases secondary, where agent is disabled in one release and server is disabled in the other?

I guess, is the deployment type referring to the configuration of the spire-server, or does it include agents. If it includes agents, there probably needs to be a few more permutations.

project/glossary.md
kfox1111: Oh. I read this differently then, so there is some ambiguity. I took "one chart is configured" to be referring to one of the helm charts (source code), and "the others" referring to other charts (more source code). helm chart refers to code.

Your not talking about charts though, your talking about either what helm calls a Release (instantiation of a chart, https://helm.sh/docs/intro/using_helm/#three-big-concepts) or what this document has been calling a Deployment.

edwbuck: Please read the whole sentence, "Use of the Helm Charts two or more times" is not a decorative statement.

Using the word release was intentionally avoided, because we will be re-releasing the Helm Charts for the same product (SPIRE) multiple times.

kfox1111 (approved):
With https://github.com/spiffe/helm-charts/issues/270 being filed to cover the remaining issue, I think this is good to go as is.

faisal-memon (approved):

Issue #268, opened 2023-05-09
Replicas > 1 is allowed with sqlite driver and it shouldn't be.


kfox1111:
We need to check for replicas > 1 when the sqlite driver is being used and block install/upgrades. It is not a supported or even workable configuration.

edwbuck:
Desired behavior here is to issue an error statement that "The SQLite3 Datastore cannot support more than one Server instance, please select another datastore or set the number of server instances to one"

That way the error message embeds the fix.

PR #269, opened 2023-05-09
Replicas > 1 is not allowed with sqlite driver

kfox1111:
This pr enforces replicas of the spire server can not be more then one when the sqlite driver is used.

fixes: https://github.com/spiffe/helm-charts/issues/268

### DIFF
--- a/charts/spire/charts/spire-server/templates/hpa.yaml
+++ b/charts/spire/charts/spire-server/templates/hpa.yaml
@@ -13,7 +13,7 @@ spec:
     name: {{ include "spire-server.fullname" . }}
   minReplicas: {{ .Values.autoscaling.minReplicas }}
   {{- if and (eq .Values.dataStore.sql.databaseType "sqlite3") .Values.autoscaling.enabled (gt (int .Values.autoscaling.maxReplicas) 1) }}
-  {{- fail "sqlite3 database only supports one replica. You can't use autoscaling.maxReplicas > 1" }}
+  {{- fail "When running with sqlite3 database, you can't scale up to more then one instance. 'autoscaling.maxReplicas' MUST be 1" }}
   {{- end }}
   maxReplicas: {{ .Values.autoscaling.maxReplicas }}
   metrics:
--- a/charts/spire/charts/spire-server/templates/hpa.yaml
+++ b/charts/spire/charts/spire-server/templates/hpa.yaml
@@ -12,6 +12,9 @@ spec:
     kind: Deployment
     name: {{ include "spire-server.fullname" . }}
   minReplicas: {{ .Values.autoscaling.minReplicas }}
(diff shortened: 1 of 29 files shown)


kfox1111:
> Might be better to merge this after the database pr as youll have to change it to check for database type.

Was thinking this is standalone enough it should review pretty fast and we can get it in, then I'd extend it in the db pr real quick to work properly with it?

charts/spire/charts/spire-server/templates/hpa.yaml
```diff
@@ -12,6 +12,9 @@ spec:
     kind: Deployment
     name: {{ include "spire-server.fullname" . }}
   minReplicas: {{ .Values.autoscaling.minReplicas }}
+  {{- if gt (int .Values.autoscaling.maxReplicas) 1 }}
```
kfox1111, line 15: If external db merges first, then this becomes something like:
{{- if and (eq .Values.dataStore.sql.databaseType "sqlite3") (gt (int .Values.autoscaling.maxReplicas) 1) }}

faisal-memon: Yes, lets merge the external db first and then this change.

kfox1111: Reordered to come after external db change.

charts/spire/charts/spire-server/templates/hpa.yaml
```diff
@@ -12,6 +12,9 @@ spec:
     kind: Deployment
     name: {{ include "spire-server.fullname" . }}
   minReplicas: {{ .Values.autoscaling.minReplicas }}
+  {{- if and (eq .Values.dataStore.sql.databaseType "sqlite3") .Values.autoscaling.enabled (gt (int .Values.autoscaling.maxReplicas) 1) }}
+  {{- fail "sqlite3 database only supports one replica. You can't use autoscaling.maxReplicas > 1" }}
```
marcofranssen, line 16:
```suggestion
  {{- fail "When running with sqlite3 database, you can't scale up to more then one instance. 'autoscaling.maxReplicas' MUST be 1" }}
```

charts/spire/charts/spire-server/templates/statefulset.yaml
```diff
@@ -11,6 +11,9 @@ metadata:
     {{- include "spire-server.labels" . | nindent 4 }}
 spec:
   {{- if not .Values.autoscaling.enabled }}
+  {{- if and (eq .Values.dataStore.sql.databaseType "sqlite3") (gt (int .Values.replicaCount) 1) }}
+  {{- fail "sqlite3 database only supports one replica." }}
```
marcofranssen, line 15:
```suggestion
  {{- fail "When running with sqlite3 database, you can't scale up to more then one instance. 'replicaCount' MUST be 1" }}
```

marcofranssen (commented):
Small suggestion on the failure message to have it aligned in style accross both scenarios.

I'm also good merging this PR first. It doesn't break anything and is a nice improvement to prevent misusing the chart.

👍

PR #271, opened 2023-05-10
Use deployment if persistence disabled

marcofranssen:
When persistence is disabled the chart will deploy a regular deployment as opposed to a statefulset.

Running this change on my existing chart installation results in no diff, meaning the macro still renders everything exactly the same as it should.

```shell
helm diff upgrade -n spire-system --install --values .local/spire-values.yaml spire ./charts/spire
```

### DIFF
--- a/charts/spire/charts/spire-server/templates/deployment.yaml
+++ b/charts/spire/charts/spire-server/templates/deployment.yaml
@@ -1,9 +1,9 @@
-{{- if eq (.Values.persistence.enabled | toString) "true" }}
+{{- if eq (.Values.persistence.enabled | toString) "false" }}
 {{- $configSum := (include (print $.Template.BasePath "/configmap.yaml") . | sha256sum) }}
 {{- $configSum2 := (include (print $.Template.BasePath "/secret.yaml") . | sha256sum) }}
 {{- $configSum3 := (include (print $.Template.BasePath "/controller-manager-configmap.yaml") . | sha256sum) }}
 apiVersion: apps/v1
-kind: StatefulSet
+kind: Deployment
 metadata:
   name: {{ include "spire-server.fullname" . }}
   namespace: {{ include "spire-server.namespace" . }}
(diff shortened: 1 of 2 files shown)


charts/spire/charts/spire-server/templates/deployment.yaml
```diff
@@ -0,0 +1,34 @@
...
+        checksum/config: {{ $configSum }}
+        checksum/config2: {{ $configSum2 }}
+        checksum/config3: {{ $configSum3 }}
+        {{- with .Values.podAnnotations }}
+          {{- toYaml . | nindent 8 }}
+        {{- end }}
+      labels:
+        {{- include "spire-server.selectorLabels" . | nindent 8 }}
+    spec:
+      {{- include "spire-server.podSpec" . | nindent 6 -}}
```
marcofranssen, line 33: This renders the same podspec as we use in the statefulset

charts/spire/charts/spire-server/templates/statefulset.yaml
```diff
@@ -32,173 +32,8 @@ spec:
...
-        {{- end }}
-        {{- if eq (.Values.controllerManager.enabled | toString) "true" }}
-        - name: controller-manager-config
-          configMap:
-            name: {{ include "spire-controller-manager.fullname" . }}
-        {{- end }}
-        {{- if gt (len .Values.extraVolumes) 0 }}
-        {{- toYaml .Values.extraVolumes | nindent 8 }}
-        {{- end }}
+      {{- include "spire-server.podSpec" . | nindent 6 }}
```
marcofranssen, line 35: This renders the same podspec as we use in the deployment.

charts/spire/charts/spire-server/templates/statefulset.yaml
```diff
@@ -32,173 +32,8 @@ spec:
       labels:
         {{- include "spire-server.selectorLabels" . | nindent 8 }}
     spec:
-      {{- with .Values.imagePullSecrets }}
```
marcofranssen, line 35: Moved to the new macro in `_pod.tpl`

marcofranssen:
Closing this PR as we will always require the persistence for non datastore related assets. In essence we can't ever disable the persistence to run this reliable.

PR #273, opened 2023-05-10
Fix spacing for spire server service annotation

faisal-memon:
Noticed the spacing for this was off.


kfox1111 (approved):
Good catch. LGTM.

PR #274, opened 2023-05-11
Remove persistence.enabled flag

faisal-memon:
Its not safe to disable persistence with the helm chart right now. Remove this option for now, revisit later.

### DIFF
--- a/charts/spire/charts/spire-server/README.md
+++ b/charts/spire/charts/spire-server/README.md
@@ -87,7 +87,6 @@ A Helm chart to install the SPIRE server.
 | nodeSelector | object | `{}` |  |
 | notifier.k8sbundle.namespace | string | `""` | Namespace to push the bundle into, if blank will default to SPIRE Server namespace |
 | persistence.accessMode | string | `"ReadWriteOnce"` |  |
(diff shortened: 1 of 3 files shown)


kfox1111:
It is safe to disable persistence in very particular circumstances. Basically just for testing. Instead of removing it, would updating the documentation on the enabled flag to explicitly state, "this should only be disabled for testing purposes" be enough? The feature has been in tree for months.

marcofranssen (approved):
LGTM 🚀

edwbuck:
Let's remove this for now, because one flag to control three things is just wrong.

This slipped through in a prior pull requests.  It was a design mistake.  Flags that are to be manipulated by the end-users should not control intermediate items like PVCs that the end users don't care about because they are an implementation detail.

This PVC holds the following:

- The backing for the sqlite3 database
- The keystore for the server instance
- The backing store Tornjak

We have flags to turn on and off sqlite3, and the backing store for Tornjak.  We can investigate which approach we want to follow in the future, choosing between the same backing store when anything needs stored being activate or one backing store per activated item.

edwbuck:
I can't add myself as a reviewer, consider this the second accepting review.  Please merge

PR #280, opened 2023-05-15
Add json to test path

faisal-memon:
Adds *.json to the workflow trigger. Hoping it will unstick this pipeline: https://github.com/spiffe/helm-charts/pull/279


.github/workflows/helm-chart-ci.yaml
edwbuck, line 1: I understand the change, and am approving.

Please describe the commit a bit more next time, to achieve a faster review / merge process.

edwbuck (approved):

edwbuck:
As lack of this fix breaks all builds, I compelled Faisal to merge immediately.

The core issue is that without the pattern, generated javascript (or javascript schema files) would break the build as they were not handled properly.   This note is for posterity's sake, as the fix is already merged.

PR #281, opened 2023-05-16
Skip tests for docs folders

faisal-memon:
Our mandatory checks will fail if there are docs only changes. The helm-chart-ci.yaml skips these files, so the workflows never run, and the mandatory checks are stuck waiting for results.

This adds a parallel workflow that runs fake tests for docs changes. So that docs only changes can pass the required status checks. It uses `paths-ignore` instead of `paths`. This is documented [here](https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/managing-protected-branches/troubleshooting-required-status-checks#handling-skipped-but-required-checks).

### DIFF
--- a/.github/workflows/helm-chart-ci-ignore.yaml
+++ b/.github/workflows/helm-chart-ci-ignore.yaml
@@ -3,7 +3,7 @@ name: Helm Chart CI
 on:
   workflow_dispatch:
   pull_request:
-    types: [synchronize, opened, reopened, edited]
+    types: [synchronize, opened, reopened]
     paths-ignore:
       - 'charts/**'
       - '.github/workflows/helm-chart-ci.yaml'
--- a/.github/workflows/helm-chart-ci-ignore.yaml
+++ b/.github/workflows/helm-chart-ci-ignore.yaml
@@ -1,0 +1,61 @@
+name: Helm Chart CI
+
+on:
+  workflow_dispatch:
+  pull_request:
+    types: [synchronize, opened, reopened, edited]
+    paths-ignore:
+      - 'charts/**'
+      - '.github/workflows/helm-chart-ci.yaml'
+      - '.github/kind/conf/kind-config.yaml'
+      - '.github/tests/**/*.yaml'
+      - '.github/tests/**/*.sh'
+      - '.github/tests/**/*.json'
+      - 'examples/**/*.yaml'
+      - 'helm-docs.sh'
+
+jobs:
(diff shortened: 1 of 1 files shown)


.github/workflows/helm-chart-ci-ignore.yaml
```diff
@@ -0,0 +1,61 @@
+name: Helm Chart CI
+
+on:
+  workflow_dispatch:
+  pull_request:
+    types: [synchronize, opened, reopened, edited]
```
marcofranssen, line 6:
```suggestion
    types: [synchronize, opened, reopened]
```

Edited runs when the description or title changes.

kfox1111:
Hmm.. Interesting problem...

Would it make sense to just always run the tests in this case too? Remove: https://github.com/spiffe/helm-charts/blob/main/.github/workflows/helm-chart-ci.yaml#L7-L15

It would mean the test suite would run on doc only changes too, but there has not been very many of those pr's so maybe the simplification would be worth it?

faisal-memon:
> Hmm.. Interesting problem...
> 
> Would it make sense to just always run the tests in this case too? Remove: https://github.com/spiffe/helm-charts/blob/main/.github/workflows/helm-chart-ci.yaml#L7-L15
> 
> It would mean the test suite would run on doc only changes too, but there has not been very many of those pr's so maybe the simplification would be worth it?

Im fine either way though slightly prefer not running tests for docs only changes.

kfox1111:
Ok. There's stuff in here that is kind of duplication from other things and needs to be synchronized between the two or else the pipeline breaks. so maybe some changes could be added to make that stuff common so we can change it just in one place?

Since things are broken, I'm ok with that kind of thing being done in a followup pr though.

kfox1111 (approved):

faisal-memon:
> Ok. There's stuff in here that is kind of duplication from other things and needs to be synchronized between the two or else the pipeline breaks. so maybe some changes could be added to make that stuff common so we can change it just in one place?
> 
> Since things are broken, I'm ok with that kind of thing being done in a followup pr though.

I agree. Is there a good way to do that? I always end up with a lot of duplication with actions.

kfox1111:
Might be able to use an environment variable from a file along with maybe the fromJson thing refereed to here: https://stackoverflow.com/questions/61919141/read-json-file-in-github-actions
for building the path list.

There may be simpler ways though. Maybe @marcofranssen knows better way?

PR #282, opened 2023-05-16
Add nodeSelector for tornjak

marcofranssen:
Tornjak only supports amd64 builds at the moment

To prevent deployment failures on clusters with arm64 nodes I have added a nodeSelector for tornjak frontend, and documentation for spire-server.

Didn't define the nodeSelector on spire-server because by default Tornjak isn't deployed and therefore doesn't require to run entire spire-server on amd64. So resolved it there using a documentation update.

Also see https://github.com/spiffe/tornjak/issues/219

### DIFF
--- a/charts/spire/README.md
+++ b/charts/spire/README.md
@@ -323,7 +323,7 @@ Kubernetes: `>=1.21.0-0`
 | spire-server.namespaceOverride | string | `""` |  |
 | spire-server.nodeAttestor.k8sPsat.enabled | bool | `true` |  |
 | spire-server.nodeAttestor.k8sPsat.serviceAccountAllowList | list | `[]` |  |
-| spire-server.nodeSelector | object | `{}` |  |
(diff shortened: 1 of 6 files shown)


kfox1111:
On the fence on this one. Total +1 for nodeSelectors.

Just the frontend amd64 predefine. There's no way to turn it back off if the user does build their own multiarch images.

But that is likely to be an exceedingly rare case and the feature is marked experimental so incomplete. So its kind of ok.

So, I'm going to approve as is. But consider just making the frontend one a suggestion too rather then required. It would be less to undo later.

kfox1111 (approved):

mrsabath:
We have seen this problem when one of our interns was using Macbook Air and the architecture was not compatible with the pre-built image. This would require us to build a few images with various architecture support. We will fix it on the Tornjak level. No changes needed here other than documenting it for now.

kfox1111:
Ok, then I suggest we make another pr removing the hard coded annotation. I forgot about the new macbooks using arm.

PR #265, opened 2023-05-09
Add maintainer's handbook.

edwbuck:
The maintainer's handbook is a guide to remind the maintainers of the values and approaches that most maintainers already hold.

Specifics of code quality will come later, but putting them into this document would make it too long, and thus unread.

Closes #263

PR #287, opened 2023-05-17
Cleanup maintainer handbook

faisal-memon:
Adds some of the comments from mistakenly closed https://github.com/spiffe/helm-charts/pull/265

### DIFF
--- a/project/maintainers.md
+++ b/project/maintainers.md
@@ -8,12 +8,12 @@ provide a consistent user interface for those going through the review process.
 
 ## The Pull Request
 
-Pull requests are submitted through GitHub.  They are contributions to change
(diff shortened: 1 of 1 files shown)


kfox1111 (approved):
LGTM

edwbuck:
This was not a missed item.  It was an item that was deemed controversial
and thus not accepted.

Even today, books of modern publication have more spacing after a sentence,
and while Slate called for one space in 2012, no study was done to see it
one space was more readable until 2018.  That study said two spaces were
better, but new studies cast doubt on the matter.

In any case, I would like to limit the call for whitespace only edits.  We
have real work to do.

PR #288, opened 2023-05-17
Remove manual dispatch from dummy workflow

marcofranssen:
This jobs just runs on docs updates to satisfy the PR merge requirements, as it doesn't do anything, there is not really a need to manually run it as it is only a requirement for PRs.


edwbuck (approved):
Looks good to me too, I'm a bit surprised that the key didn't have a value, even if it was empty (aka {})

Issue #284, opened 2023-05-16
Tornjak review issues.


kfox1111:
There were a few issues identified by a final review of the branch that still need addressing. Referencing here so we don't forget:
https://github.com/spiffe/helm-charts/pull/234#pullrequestreview-1429118074

For convenience here the unhandled feedback.

- [x] tornjak frontend missing in README.md (https://github.com/spiffe/helm-charts/pull/234/files#r1196910210)
- [x] Question on Tornjak metadata backend (hardcoded at "insert metadata") https://github.com/spiffe/helm-charts/pull/234/files#r1195518417
- [x] #291
- [x] Add "(Not for production)" to the end of the tornjak enabled variables. #297

marcofranssen:
@mrsabath Could you check the open questions see above in this issue description. (middle checkbox).

kfox1111:
The hardcoded bit is needed to get standalone without auth working. It will be enhanced to make it configurable when user auth support is added.

kfox1111:
one more thing from the review...
tornjak-frontend.enabled / spire-server.tornjak.enabled, should we put 'Experimental: ' at the begging of the readme docstring for it?
we have the warning for after its enabled. but maybe one in the docs that way for before?

marcofranssen:
Yes probably add something like `(Not for production)` at the enabled values.

kfox1111:
All done. Thanks! :)

PR #297, opened 2023-05-18
Updated Tornjak documenation with Not-for-production labels

mrsabath:
Resolves the requirements from #284

### DIFF
--- a/charts/spire/charts/spire-server/values.yaml
+++ b/charts/spire/charts/spire-server/values.yaml
@@ -266,8 +266,8 @@ nodeAttestor:
 
 # tornjak - Tornjak specific configuration
 tornjak:
-  # -- Deploys Tornjak API (backend)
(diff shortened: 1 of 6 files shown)


charts/spire/charts/spire-server/values.yaml
```diff
@@ -266,8 +266,8 @@ nodeAttestor:
 
 # tornjak - Tornjak specific configuration
 tornjak:
-  # -- Deploys Tornjak API (backend)
-  enabled: false
+  # -- Deploys Tornjak API (backend) (Not for production)
+  enabled: false # (Not for production)
```
kfox1111, line 270: I'd be ok with just the comment at 269 and not on the enabled line if it makes the linter more happy.

kfox1111 (approved):
LGTM

PR #300, opened 2023-05-22
Add back persistence.enabled flag

faisal-memon:
Reverts spiffe/helm-charts#274

Can create Deployment instead of StatefulSet if needed.

### DIFF
--- a/charts/spire/README.md
+++ b/charts/spire/README.md
@@ -336,6 +336,7 @@ Kubernetes: `>=1.21.0-0`
 | spire-server.nodeSelector | object | `{}` | Select specific nodes to run on (currently only amd64 is supported by Tornjak) |
 | spire-server.notifier.k8sbundle.namespace | string | `""` | Namespace to push the bundle into, if blank will default to SPIRE Server namespace |
 | spire-server.persistence.accessMode | string | `"ReadWriteOnce"` |  |
+| spire-server.persistence.enabled | bool | `true` | Enable persistence used by sqlite3 for spire-server, by the disk KeyStore, and/or by Tornjak. |
 | spire-server.persistence.size | string | `"1Gi"` |  |
 | spire-server.persistence.storageClass | string | `nil` |  |
 | spire-server.podAnnotations | object | `{}` |  |
--- a/charts/spire/charts/spire-server/README.md
+++ b/charts/spire/charts/spire-server/README.md
@@ -90,6 +90,7 @@ A Helm chart to install the SPIRE server.
 | nodeSelector | object | `{}` | Select specific nodes to run on (currently only amd64 is supported by Tornjak) |
 | notifier.k8sbundle.namespace | string | `""` | Namespace to push the bundle into, if blank will default to SPIRE Server namespace |
 | persistence.accessMode | string | `"ReadWriteOnce"` |  |
+| persistence.enabled | bool | `true` | Enable persistence used by sqlite3 for spire-server, by the disk KeyStore, and/or by Tornjak. |
 | persistence.size | string | `"1Gi"` |  |
 | persistence.storageClass | string | `nil` |  |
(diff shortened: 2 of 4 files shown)


edwbuck:
@faisal-memon The reason it was killed is because it's a flag on internal policy, not no customer featuring items, and disabling it is likely to create a non-launching deployment.

If we are adding this back (which was discussed in the prior pull request) we need to do it better than how we removed it.  Instead of exposing our internal policy to the api / values.yaml interface, we should:

1. Check if the datastore is set to sqlite3, and if so, then enable the need for the local volume.
2. Check if the keystore is set to disk, and if so, then enable the need for the local volume.
3. Check if Torjak is enabled, and if so, then enable the need for the local volume.

I'm fine with the local volume being reused across multiple items (I'm also ok with individual local volumes, but that's a bigger change).  I just don't want our users to determine the policy of when we need one, as then they have to coordinate two or more options, which are tightly coupled, with the correct settings on both or the deployment will fail.

This simple "revert" is rejected, as that means we'll have to go through the process again.  If this is an emergency need for the volume, I would be more comfortable having it enabled 100% of the time until we can do the detection in a way that doesn't make the end-user decide if they need persistence in addition to the choices that require persistence.

kfox1111:
Its useful in testing scenarios (for example, test pipelines) where you want the service to function basically normally but don't need/desire persistence.

So the bit that seems to be missing here is how to put a word of warning on the enable flag to ensure its only used appropriately.

How about:
This flag is only to be used in certain test situations where you do not want persistence. For production always leave this enabled.

For testing proposes, there isn't a way to autodetect that use case so a flag would be needed.

kfox1111:
Yeah, that works for me.

kfox1111:
Drafting this for now since there is likely a rewrite needed based on the way the discussion is going.

marcofranssen:
> > Its useful in testing scenarios (for example, test pipelines) where you want the service to function basically normally but don't need/desire persistence.
> 
> After a chat with Kevin, I now better understand his need here. He's disabling persistence to push items that typically persist into the writable layer of his containerization platform, to speed up testing scenarios where SPIRE deployments provide some of the testing environment.
> 
> In my mind, this fixes the wrong layer of the solution, effectively creating two different "kinds" of deployments, one with a volume and one without. We will never formally support both kinds of deployments, which means this is the wrong architectural approach.
> 
> Instead, we should add the necessary code to set the volume type, such that Kevin can use the volume, but specify it to back to RAM, temporary directory, etc. such that he can then achieve his testing without a "special" override on volume creation policy.
> 
> This means that we still will have to apply the policy to make the volume be created / accessed when sqlite, Tornjak, etc. require it; but, that the volume it creates might be temporary, persistent, backed by a remote system, etc. based on the volume type configuration.

I'm really in favor of taking this PR as opposed to doing something deviating from the approach most other charts are using.

Most charts just themplate the podspec in a helper and then either wrap it with a statefulset or a deployment. That is the much cleaner approach as opposed to adding a persistence type which is done in the other PR. It gives a much cleaner user facing API as well leverages more of the core kubernetes object types in a way they are intended.

faisal-memon:
> I'm really in favor of taking this PR as opposed to doing something deviating from the approach most other charts are using.
> 
> Most charts just themplate the podspec in a helper and then either wrap it with a statefulset or a deployment. That is the much cleaner approach as opposed to adding a persistence type which is done in the other PR. It gives a much cleaner user facing API as well leverages more of the core kubernetes object types in a way they are intended. It seems we are reinventing the wheel in a non Helm/k8s way of doing things. Why do we want to approach all different then the majority of the community?

+1, this is good enough. If we need something more complex later we can revisit.

edwbuck:
I think the team has agreed to not merge this.  Can we close this PR?

faisal-memon:
Closing in favor of #315

PR #303, opened 2023-05-22
Fix the generated pr so that it runs jobs too

kfox1111:


edwbuck (commented):
Seems reasonable.  Is there any exception for documentation files that are in the test directory (aka *.md or *.adoc)?

kfox1111:
> Seems reasonable. Is there any exception for documentation files that are in the test directory (aka *.md or *.adoc)?

There aren't any documents there at the moment. We can add exceptions if we ever do?

edwbuck:
> > Seems reasonable. Is there any exception for documentation files that are in the test directory (aka *.md or *.adoc)?

> There aren't any documents there at the moment. We can add exceptions if we ever do?

Good enough for me then, accpeting

edwbuck (approved):

faisal-memon:
How will this resolve the tests not running on https://github.com/spiffe/helm-charts/pull/301? That is a .json file which is already included.

kfox1111:
I'm thinking .github/tests/**/*.json isn't matching .github/tests/*.json? Maybe I'm missing something

faisal-memon:
> I'm thinking `.github/tests/**/*.json` isn't matching `.github/tests/*.json`? Maybe I'm missing something

It should match. I think we are missing something. 

```
➜  .github git:(main) pwd
/Users/fmemon/github.com/spiffe/helm-charts/.github
➜  .github git:(main) ls tests/**/*.json
tests/charts.json
```

faisal-memon:
Another thing to consider is if it doesnt match a path it should have fallen to the paths-ignore path and skipped the tests

faisal-memon:
Is this possibly the reason? https://stackoverflow.com/questions/73079924/github-workflows-not-triggered-by-automatically-created-prs

kfox1111:
> Is this possibly the reason? https://stackoverflow.com/questions/73079924/github-workflows-not-triggered-by-automatically-created-prs

Hmm... yeah. That does sound like it could be the real issue. Will update the pr here to try it.

marcofranssen (approved):

PR #304, opened 2023-05-22
Allow to use spire-server as an upstream authority

kfox1111:
These are the bare minimum configurables required for the nested spire use case:

### DIFF
--- a/charts/spire/charts/spire-server/templates/configmap.yaml
+++ b/charts/spire/charts/spire-server/templates/configmap.yaml
@@ -1,4 +1,5 @@
 {{- define "spire-server.yaml-config" -}}
+{{- $upstreamAuthorityUsed := 0 }}
 {{- $root := . }}
 server:
   bind_address: "0.0.0.0"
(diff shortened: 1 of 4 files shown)


faisal-memon (approved):

faisal-memon:
We should probably error if multiple upstreams are enable at some point.

charts/spire/charts/spire-server/templates/configmap.yaml
```diff
@@ -85,6 +88,21 @@ plugins:
...
+  UpstreamAuthority:
+    - spire:
+        plugin_data:
+          server_address: {{ .server.address | quote }}
+          server_port: {{ .server.port }}
+          workload_api_socket: "/run/spire/upstream_agent/spire-agent.sock"
+  {{- end }}
+  {{- end }}
+{{- if gt $upstreamAuthorityUsed 1 }}
+{{- fail "Too many Upstream Authorities used. You can only enable one." }}
```
marcofranssen, line 103:
```suggestion
{{- fail "You can only enable a single Upstream Authority." }}
```

marcofranssen (approved):
LGTM 🚀

PR #260, opened 2023-05-04
Add experimental flag for nested configuration

kfox1111:
This patch adds a non working flag for specifying a mode of nested spire along with some unit tests to ensure output.

PR #306, opened 2023-05-23
Remove 1.21.x testing

kfox1111:
Other tests now are depending on 1.22+.

### DIFF
--- a/charts/spire/README.md.gotmpl
+++ b/charts/spire/README.md.gotmpl
@@ -23,7 +23,7 @@
 | Helm       | `3.x`              |
 | Kubernetes | `1.22+`            |
 
-> **Note**: For Kubernetes, we will officially support the last 3 versions as described in [k8s versioning](https://kubernetes.io/releases/version-skew-policy/#supported-versions). Any version before the last 3 we will try to support as long it doesn't bring security issues or any big maintenance burden. *The first version we tested this chart with is `1.22`.*
+> **Note**: For Kubernetes, we will officially support the last 3 versions as described in [k8s versioning](https://kubernetes.io/releases/version-skew-policy/#supported-versions). Any version before the last 3 we will try to support as long it doesn't bring security issues or any big maintenance burden. 
 
 ## Prerequisites
 
--- a/charts/spire/README.md.gotmpl
+++ b/charts/spire/README.md.gotmpl
@@ -87,7 +87,9 @@ Now you can interact with the Spire agent socket from your own application. The 
 
 {{ template "chart.sourcesSection" . }}
 
-{{ template "chart.requirementsSection" . }}
+{{ template "chart.requirementsHeader" . }}
+
+{{ template "chart.requirementsTable" . }}
 
 {{ template "chart.valuesSection" . }}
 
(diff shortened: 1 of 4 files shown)


faisal-memon:
Thanks @kfox1111. What do you think of removing 1.22-1.24 as well to get within the n-2 range?

faisal-memon:
Can you update the base README.md and remove k8s 1.21 from there?

kfox1111:
I do think some of the integration tests we do are somewhat excessive and could be handled better with unit testing instead. That's where something like the unit test framework in https://github.com/spiffe/helm-charts/pull/260 could come in handy. We could then remove some of the integration tests reducing the number of permutations that way?

kfox1111:
> Also need to make the same change to `helm-chart-ci-ignore.yaml`

I looked for it there but couldn't find it?

faisal-memon:
> > Also need to make the same change to `helm-chart-ci-ignore.yaml`
> 
> I looked for it there but couldn't find it?

Yes, sorry, not needed there.

charts/spire/README.md.gotmpl
```diff
@@ -21,9 +21,9 @@
 |:-----------|:-------------------|
 | SPIRE      | `1.5.3+`, `1.6.3+` |
 | Helm       | `3.x`              |
-| Kubernetes | `1.21+`            |
+| Kubernetes | `1.22+`            |
 
-> **Note**: For Kubernetes, we will officially support the last 3 versions as described in [k8s versioning](https://kubernetes.io/releases/version-skew-policy/#supported-versions). Any version before the last 3 we will try to support as long it doesn't bring security issues or any big maintenance burden. *The first version we tested this chart with is `1.21`.*
+> **Note**: For Kubernetes, we will officially support the last 3 versions as described in [k8s versioning](https://kubernetes.io/releases/version-skew-policy/#supported-versions). Any version before the last 3 we will try to support as long it doesn't bring security issues or any big maintenance burden. *The first version we tested this chart with is `1.22`.*
```
faisal-memon, line 26:
```suggestion
> **Note**: For Kubernetes, we will officially support the last 3 versions as described in [k8s versioning](https://kubernetes.io/releases/version-skew-policy/#supported-versions). Any version before the last 3 we will try to support as long it doesn't bring security issues or any big maintenance burden. 
```

faisal-memon: Dont think we need that bit of history anymore.

faisal-memon (approved):
Thanks @kfox1111

edwbuck (changes requested):
What about the generated README.md in the root of the repo?

What about the kubeVersion in charts/spire/Chart.yaml?

kfox1111:
> What about the generated README.md in the root of the repo?

Fixed
 
> What about the kubeVersion in charts/spire/Chart.yaml?

Fixed. It needs to stay that way to not block installs. But I was able to fix up the docs.

edwbuck (approved):
With the last two commits, all docs now properly report k8s versions 1.22+ are supported, and the Charts.yaml will still deploy on 1.21+, for those that have 1.21 clusters and desire to take their own risks in running on something we might not fully test.

Approved

PR #308, opened 2023-05-23
Upgrade to spire 1.6.4

kfox1111:


faisal-memon (approved):

edwbuck (approved):
charts/spire/README.md.gotmpl also contains a reference to 1.6.3+, which technically contains 1.6.4, so I think this is good to go

PR #309, opened 2023-05-23
Fix bundle role/rolebinding naming conflict

kfox1111:
When you install a primary and a secondary deployment in the same Kubernetes cluster, use the same release name (spire), and configure the primary to give the bundle to the secondary, both releases want to own the same role/rolebinding name. This patch adds the release namespace into the names to prevent conflict.

### DIFF
--- a/charts/spire/charts/spire-server/templates/roles.yaml
+++ b/charts/spire/charts/spire-server/templates/roles.yaml
@@ -3,7 +3,7 @@
 kind: Role
 apiVersion: rbac.authorization.k8s.io/v1
 metadata:
-  name: {{ include "spire-server.fullname" . }}-bundle
+  name: {{ $namespace }}-{{ include "spire-server.fullname" . }}-bundle
   namespace: {{ .Values.notifier.k8sbundle.namespace | default $namespace }}
 rules:
   - apiGroups: [""]
@@ -50,7 +50,7 @@ roleRef:
 kind: RoleBinding
 apiVersion: rbac.authorization.k8s.io/v1
 metadata:
-  name: {{ include "spire-server.fullname" . }}-bundle
+  name: {{ $namespace }}-{{ include "spire-server.fullname" . }}-bundle
   namespace: {{ .Values.notifier.k8sbundle.namespace | default $namespace }}
 subjects:
   - kind: ServiceAccount
@@ -58,7 +58,7 @@ subjects:
     namespace: {{ $namespace }}
 roleRef:
   kind: Role
-  name: {{ include "spire-server.fullname" . }}-bundle
+  name: {{ $namespace }}-{{ include "spire-server.fullname" . }}-bundle
   apiGroup: rbac.authorization.k8s.io
 
 {{- if and .Values.nodeAttestor.k8sPsat.enabled }}


faisal-memon:
Is this workflow common practice? I thought using release name is how to differentiate between two deployments?

kfox1111:
Release name is only unique within a namespace. It is common to use the same release name across multiple namespaces.

faisal-memon:
> Release name is only unique within a namespace. It is common to use the same release name across multiple namespaces.

Thanks, one more question. Do you know what other charts do to handle this scenario?

kfox1111:
Most apps don't integrate to the point where one instance talks to another via kubernetes objects. Usually just via the network. But spire needs trust bundles and we do that via k8s configmaps.

marcofranssen (changes requested):
Why do we need to namespace these role names? I assume you are doing this for the nested setup, but isn't it cleaner to just use different release names when deploying spire?

The `spire-server.fullname` macro already handles that.

So

```shell
helm install -n spire-system spire ./charts/spire
helm install -n spire-upstream upstream-spire ./charts/spire
```

this would already give 2 distinct names for the roles being created.

namely `spire-bundle` and `upstream-spire-bundle`.

I find this change really bloating the clean naming for a single usecase where it can already be resolved by accordingly naming the release. This also gives the advantage to have a better overview of the helm releases.

```shell
$ helm list -A
NAME                            NAMESPACE               REVISION        UPDATED                                         STATUS          CHART                                    APP VERSION
spire                           spire-system            1               2023-05-16 15:47:32.871591 +0200 CEST           deployed        spire-0.6.3                              1.6.3
upstream-spire                  spire-upstream          1               2023-05-16 15:47:32.871591 +0200 CEST           deployed        spire-0.6.3                              1.6.3
```

Lastly in practice I see the nested approach mostly being used in 2 different clusters. Having the nested setup in a single cluster to me seems something you do in test clusters to save on costs where the different helm release name can already make it happen.

This change would in all cases gives these roles a bloated name.

`spire-system-spire-bundle` and `spire-upstream-upstream-spire-bundle` respectavily.

Do you have a link to that best-practice to give helm releases the same name in different namespaces? We always give releases unique names, which is our practice (not claiming our practice is the best).

faisal-memon:
> Just renaming the release name will work around the issue. But some users will be confused by it. And some will report it to us. Fixing it now will prevent later work for us IMO.

Most people probably won't do this. If anyone else does, we can let them know its not a supported config and direct them to change the release name. If it really does become a problem down the road we can revisit and make a change like this.

faisal-memon:
>  So, when I install a mysql in namespace a, I helm install -n a mysql bitnami/mysql. I need another one in namespace b, I helm install -n b mysql bitnami/mysql.

How does mysql handle this?

marcofranssen:
> > So, when I install a mysql in namespace a, I helm install -n a mysql bitnami/mysql. I need another one in namespace b, I helm install -n b mysql bitnami/mysql.
> 
> How does mysql handle this?

Exactly the way we already have it implemented. Same for basically any other charts.

https://artifacthub.io/packages/helm/bitnami/mysql?modal=template&template=role.yaml

See here for that macro function implementation
https://artifacthub.io/packages/helm/bitnami/common?modal=template&template=_names.tpl

Which is identical to what we do.

IMHO we close this PR and leave things as they are.

kfox1111:
Please reconsider and think of it from a different angle...

We need to go from one of, thinking why do we need to accept a change (it costs us to review)
to one more of like: what does this pr hurt, and what does it help?

It hurts, nothing to accept this pr. Anyone not using nested spire won't notice it one bit.

It helps:
* users that helm install this stuff and happen to use the same name. This is not unreasonable for them to do IMO.
* it helps us not have to deal with support tickets and explain to those users why they "did it wrong"

So, please explain, why block it? I see no reason to. "mysql doesn't do it" isn't a reason, when mysql doesn't do the same things nested spire does.

faisal-memon:
> We need to go from one of, thinking why do we need to accept a change (it costs us to review)
to one more of like: what does this pr hurt, and what does it help?

I get that. Taking in this change would enable this use case and its a small change. Where it does hurt is the maintenance overhead of supporting a nonstandard usage. It seems like you are trying to get around limitations in helm with this. Is there a reason why they don't support the same name for resources in different namespace?

kfox1111:
Is the ongoing maintenance burden of having the namespace name in the 2 objects (3 lines) really that much more then the ongoing maintenance burden we will face when users try and use it in this way and we have to keep telling them, do't it that way, along with the inconvenience we place on them?

Its not a helm limitation at all, but due to the way spire nesting needs a configmap passed between the instances (namespaces) and the way rbac is designed in k8s.

Its not common for most charts to be writing k8s objects into other namespaces so rarely hit naming conflicts across namespaces. But spire is not a common app either.

marcofranssen:
For me it is not the maintenance burden, it is the clean naming of resources that bothers me.

Maybe we can do something like this https://github.com/spiffe/helm-charts/blob/main/charts/spire/charts/spire-server/templates/_helpers.tpl#L13-L24

To prevent the double naming if it is not about the nested spire.

kfox1111:
Had an idea of another way of guaranteeing uniqueness that may be better. Please see https://github.com/spiffe/helm-charts/pull/333

edwbuck:
We have a new item in review, to address the same item here.  See #333 .

For this reason, I'm going ask that @kfox1111 close this PR.

This two PRs for one issue is a common pattern, one that I ask we avoid in the future.  If a PR is old new, and not to be merged, then close it with a comment saying so.  If it is still current, and just being updated to be merged (including updates that amount to full rewrites), then don't open a new one, update the current one.

If you are seeking guidance on what approaches people like, don't submit multiple PRs for the effort.  Reach out the team via slack or other channels and just ask.

Issue #307, opened 2023-05-23
Clean up values.yaml files


faisal-memon:
The files are getting a little messy and can be organized better.

mrsabath:
I volunteer to add the probes details parameters as I did with Tornjak frontend: https://github.com/spiffe/helm-charts/blob/main/charts/spire/charts/tornjak-frontend/values.yaml#L78-L90

edwbuck:
Can we describe the goals here?  This seems to be an unbounded task without them, and unless we can set a few goals, it might even be an entirely opinion driven task.

krishnakv:
Hi @edwbuck , happy to pick this up. Will generate a PR this week with the approach/ understanding that `helm show values` from the command line should generate all the relevant values with comments. This will help an operator understand the overall chart and generate a modified YAML for their purpose, all from the command line.

kfox1111:
Just to double check, I think the plan was to find the most common config options most users will need and copy those in with comments. Lots of the options will be rarely used and we don't plan on putting those in the parent chart?

marcofranssen:
I suggested to put a single comment to the readme.md section with values which holds all options already so we don't have to duplicate the info. Just a hyperlink so people searching there are directed to location where they can find it.

Duplicating then is a bad idea IMHO.

kfox1111:
It wasn't to copy all the options, it was to copy the few options that users would likely need in the normal useage of the chart, then remove the flag that sets all the subchart values into the parent's readme?

marcofranssen:
> It wasn't to copy all the options, it was to copy the few options that users would likely need in the normal useage of the chart, then remove the flag that sets all the subchart values into the parent's readme?

That would be a bad choice IMHO. then we would end up with a half documented chart + the fact for advanced users that have to dive deeper things will get more complicated. What we should do is add those `--@ignore` annotations for things we really want to keep out of the docs.

krishnakv:
Hi Folks, just to add the reasoning behind why this values format i.e. all allowed values with commentary and examples is typically maintained in the default values file. 

1. Helm charts are self describing and most operators expect this format
2. A workflow involving cutting and pasting from a README file is messy as YAML is whitespace sensitive - with this default, operators just to have uncomment a few lines and fill in their values
3. Again, most charts have all values in the public interface because users will just skip past or delete the infrequently used options (e.g. tolerations??)
[...]

PR #310, opened 2023-05-23
Parametrize probes

mrsabath:
This PR addresses #307 by parametrizing Probes and moving them to values.yaml

### DIFF
--- a/charts/spire/README.md
+++ b/charts/spire/README.md
@@ -133,6 +133,8 @@ Kubernetes: `>=1.21.0-0`
 | spiffe-csi-driver.image.version | string | `""` | This value is deprecated in favor of tag. (Will be removed in a future release) |
 | spiffe-csi-driver.imagePullSecrets | list | `[]` |  |
 | spiffe-csi-driver.kubeletPath | string | `"/var/lib/kubelet"` |  |
+| spiffe-csi-driver.livenessProbe.initialDelaySeconds | int | `5` | Initial delay seconds for livenessProbe |
+| spiffe-csi-driver.livenessProbe.timeoutSeconds | int | `5` | Timeout value in seconds for livenessProbe |
 | spiffe-csi-driver.nameOverride | string | `""` |  |
 | spiffe-csi-driver.namespaceOverride | string | `""` |  |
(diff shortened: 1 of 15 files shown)


charts/spire/README.md
```diff
@@ -292,6 +292,10 @@ Kubernetes: `>=1.21.0-0`
 | spire-server.controllerManager.image.repository | string | `"spiffe/spire-controller-manager"` | The repository within the registry |
 | spire-server.controllerManager.image.tag | string | `"0.2.2"` | Overrides the image tag |
 | spire-server.controllerManager.image.version | string | `""` | This value is deprecated in favor of tag. (Will be removed in a future release) |
+| spire-server.controllerManager.livenessProbe.httpGet.path | string | `"/healthz"` |  |
```
kfox1111, line 295: This whole pr looks very nice. Thanks. :)

The one thing I think we may want to do/not do is to make the paths/ports/scheme overridable. there isn't a different setting the user would ever need to set for those? so maybe we hard code those still.

kfox1111: And maybe make the rest of the missing liveness/readyness/startup probe settings configurable?

mrsabath: So @kfox1111 , are you suggesting format something like this?:

```
          readinessProbe:
            httpGet:
              scheme: HTTP
              path: /ready
              port: healthz
            failureThreshold: {{ .Values.startupProbe.failureThreshold }}
            initialDelaySeconds: {{ .Values.startupProbe.initialDelaySeconds }}
            periodSeconds: {{ .Values.startupProbe.periodSeconds }}
            successThreshold: {{ .Values.startupProbe.successThreshold }}
            timeoutSeconds: {{ .Values.startupProbe.timeoutSeconds }}
 ```

kfox1111: yeah. that kind of thing.

kfox1111: That way, the user can configure the settings that may differ cluster to cluster (slower cluster, higher timeouts) but not have to worry about the stuff that is internal to the pod that won't change on them.

marcofranssen (approved):
LGTM 🚀

kfox1111 (approved):
LGTM

PR #312, opened 2023-05-24
Allow for having no registry specified

faisal-memon:
Trying to use a local registry gives the following error:

```
  Warning  InspectFailed  54s (x6 over 63s)  kubelet            Failed to apply default image tag "/spire-server:latest-local": couldn't parse image reference "/spire-server:latest-local": invalid reference format
  Warning  Failed         54s (x6 over 63s)  kubelet            Error: InvalidImageName
```

Config I used:
```
  image:
    registry: ""
    repository: "spire-server"
    tag: "latest-local"
```

This PR allow the `image.registry` field to be empty for local registries.

### DIFF
--- a/charts/spire/templates/_spire-lib.tpl
+++ b/charts/spire/templates/_spire-lib.tpl
@@ -32,9 +32,11 @@
 
 {{- define "spire-lib.registry" }}
 {{- if ne (len (dig "spire" "image" "registry" "" .global)) 0 }}
-{{- .global.spire.image.registry }}
+{{- print .global.spire.image.registry "/"}}
+{{- else if ne (len (.image.registry)) 0 }}
+{{- print .image.registry "/"}}
 {{- else }}
-{{- .image.registry }}
+{{- print ""}}
 {{- end }}
 {{- end }}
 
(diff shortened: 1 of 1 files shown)


faisal-memon:
> For local registries, it should be set to localhost or whatever the local server name is?

Sorry bad description. there is no registry. im just trying run a locally compiled spire server.

kfox1111:
Oh, gotcha.

charts/spire/templates/_spire-lib.tpl
```diff
@@ -32,9 +32,11 @@
 
 {{- define "spire-lib.registry" }}
 {{- if ne (len (dig "spire" "image" "registry" "" .global)) 0 }}
-{{- .global.spire.image.registry }}
+{{- print .global.spire.image.registry "/"}}
+{{- else if ne (len (.image.registry)) 0 }}
+{{- print .image.registry "/"}}
 {{- else }}
-{{- .image.registry }}
+{{- print ""}}
```
kfox1111, line 39: Can leave out the else and print.

kfox1111 (approved):
One possible change but not a deal breaker. Looks good to me.

charts/spire/templates/_spire-lib.tpl
faisal-memon: how would that work? Need the else for the empty registry field case.

kfox1111: the macro's just printing stuff out. you have logic to print nothing out, but actually not doing anything also has the same affect? Basically the else and print are doing nothing but taking cycles?

faisal-memon: Got it, thanks for clarifying

marcofranssen: Is this giving an image like `/spire-server` is that actually a valid image name starting with a `/` haven't seen that before.

kfox1111: Without the change it would have an extra / in front. Depending on the container runtime, that may or may not work. Dropping out the / in the case where registry is overridden to be "" seems reasonable to me.

faisal-memon: What Kevin said. Without this change docker complains about the extra slash in front. 

Removed the extra else statement.

kfox1111 (approved):
LGTM

marcofranssen (approved):

PR #315, opened 2023-05-25
Add persistence type flag

kfox1111:
This patch adds a type flag to the persistence settings to enable specifying the backing volume's type.


kfox1111:
Implements the suggestion from https://github.com/spiffe/helm-charts/pull/300

mrsabath (approved):
This is very useful, especially when the `pvc` is not directly available, e.g. default EKS on AWS without block storage. 
LGTM

faisal-memon:
Is  https://github.com/spiffe/helm-charts/pull/300 still needed with this change?

kfox1111:
> Is #300 still needed with this change?

This would remove the need for #300.

marcofranssen (changes requested):
See https://github.com/spiffe/helm-charts/pull/300#issuecomment-1565305701

I think it is better to do a pod template in a macro and reuse that in the Deployment vs Statefulset based on if Persistence is enabled. That is the way I have seen it in many Helm charts and keeps the api much cleaner and leverage more kubernetes native concepts as opposed to reinventing a new approach. I really dislike bloating the public API for test case performance reasons.

faisal-memon:
> See [#300 (comment)](https://github.com/spiffe/helm-charts/pull/300#issuecomment-1565305701)
> 
> I think it is better to do a pod template in a macro and reuse that in the Deployment vs Statefulset based on if Persistence is enabled. That is the way I have seen it in many Helm charts and keeps the api much cleaner and leverage more kubernetes native concepts as opposed to reinventing a new approach. I really dislike bloating the public API for test case performance reasons.

Agree with the above.

faisal-memon:
Been thinking over this one. At the last meeting someone mentioned that alot this can be done at the storage class level. This seems to me a better place to decide your persistence type, rather than here at the SPIRE level. We should just plug into whatever persistence is provided to us.

kfox1111:
> Been thinking over this one. At the last meeting someone mentioned that alot this can be done at the storage class level. This seems to me a better place to decide your persistence type, rather than here at the SPIRE level. We should just plug into whatever persistence is provided to us.

That works really well for just about any csi driver and the recommended path forward that direction. emtpydir and hostpath are kind of the exception to that rule.

edwbuck:
> > Been thinking over this one. At the last meeting someone mentioned that alot this can be done at the storage class level. This seems to me a better place to decide your persistence type, rather than here at the SPIRE level. We should just plug into whatever persistence is provided to us.
> 
> That works really well for just about any csi driver and the recommended path forward that direction. emtpydir and hostpath are kind of the exception to that rule.

Hostpath and emptydir are not exceptions to the design precept that storage should be configurable, even though k8s has inconsistencies in how they are configured, which makes them us a different mechanisim that the CSI driver flag.

In any case, the PR is not for "configuring CSI drivers" it is for "configuring persistence type".  While the implementation for hostpath and emptydir will yield different YAML than that of CSI driver configuration, setting the persistence type to any of these values will still require the "persistence type flag".

With this alignment, can we please close #300 as not-to-be merged?  We have enough "there's two implementations preventing movement on both" issues in review already.

charts/spire/README.md
```diff
@@ -362,8 +362,10 @@ Now you can interact with the Spire agent socket from your own application. The
 | spire-server.nodeSelector | object | `{}` | Select specific nodes to run on (currently only amd64 is supported by Tornjak) |
 | spire-server.notifier.k8sbundle.namespace | string | `""` | Namespace to push the bundle into, if blank will default to SPIRE Server namespace |
 | spire-server.persistence.accessMode | string | `"ReadWriteOnce"` |  |
+| spire-server.persistence.hostPath | string | `""` | Which path to use on the host when type = hostPath |
 | spire-server.persistence.size | string | `"1Gi"` |  |
 | spire-server.persistence.storageClass | string | `nil` |  |
+| spire-server.persistence.type | string | `"pvc"` | What type of volume to use for persistence. Valid options pvc (recommended), hostPath, emptyDir (testing only) |
```
edwbuck, line 368: We should default to emptyDir.  That's because it is the most secure, and the one type that is more likely to be present in every deployment.  Those that want better can reconfigure it to match their environment.

charts/spire/charts/spire-server/README.md
```diff
@@ -108,8 +108,10 @@ A Helm chart to install the SPIRE server.
 | nodeSelector | object | `{}` | Select specific nodes to run on (currently only amd64 is supported by Tornjak) |
 | notifier.k8sbundle.namespace | string | `""` | Namespace to push the bundle into, if blank will default to SPIRE Server namespace |
 | persistence.accessMode | string | `"ReadWriteOnce"` |  |
+| persistence.hostPath | string | `""` | Which path to use on the host when type = hostPath |
 | persistence.size | string | `"1Gi"` |  |
 | persistence.storageClass | string | `nil` |  |
+| persistence.type | string | `"pvc"` | What type of volume to use for persistence. Valid options pvc (recommended), hostPath, emptyDir (testing only) |
```
edwbuck, line 114: Same comment as elsewhere that this should default to emptyDir.

charts/spire/charts/spire-server/templates/statefulset.yaml
```diff
@@ -3,6 +3,9 @@
 {{- $configSum3 := (include (print $.Template.BasePath "/controller-manager-configmap.yaml") . | sha256sum) }}
 {{- $configSumTornjak := (include (print $.Template.BasePath "/tornjak-config.yaml") . | sha256sum) }}
 {{- $fullname := include "spire-server.fullname" . }}
+{{- if not (has .Values.persistence.type (list "pvc" "hostPath" "emptyDir")) }}
```
edwbuck, line 6: Kudos to you for the excellent value checking on what is effectively an enum.

charts/spire/charts/spire-server/templates/statefulset.yaml
```diff
@@ -238,6 +241,16 @@ spec:
...
+        {{- if eq .Values.persistence.type "emptyDir" }}
+        - name: spire-data
+          emptyDir: {}
+        {{- else if eq .Values.persistence.type "hostPath" }}
+        - name: spire-data
+          hostPath:
+            path: {{ .Values.persistence.hostPath }}
+            type: Directory
+        {{- end }}
+  {{- if eq .Values.persistence.type "pvc" }}
```
edwbuck, line 253: Please make this key off the "if-elseif-end" above, making it an "if-elseif-elseif-end" OR make each statement an "if-end" statement.  By having two values handled one way and the third a different way, it presents more mental load on developers reading it.

I'm fine including each block independently keyed of its own activating value.  I'm also ok with a giant switch statement implemented in if-endif-endif-else patterns.  I'm just not fine with the hybrid of the two on one enum.  If you care to know my preference, it would be for multiple if-end statements, as they can be moved more easily.

charts/spire/charts/spire-server/values.yaml
```diff
@@ -98,9 +98,13 @@ readinessProbe:
   periodSeconds: 5
 
 persistence:
+  # -- What type of volume to use for persistence. Valid options pvc (recommended), hostPath, emptyDir (testing only)
+  type: pvc
```
edwbuck, line 102: please default to emptyDir.  Not everyone will have a backing store, and one should be able to launch the product to demo it without configuration.

under peristence.type, please structure the per-type values like so

persistence.type.hostpath = "/string/to/path"
persistence.type.pvc.size = whatever
persistence.type.pvc.accessMode = whatever
etc.

If you wish to make persistence.type.pvc into an empty object for maximum user flexibility, I feel that would also be acceptable.

edwbuck (changes requested):
A number of changes were requested in the per-file comments.  Please address them individually, I will be closely monitoring this and will clear this message's "request for changes" afterwards.

charts/spire/charts/spire-server/templates/statefulset.yaml
kfox1111: Since the volume mounts and volumes are different blocks, different if's for those blocks makes more sense to me. Then I think the separate if's for the items in volumeMounts would be the better route. What do others think?

charts/spire/README.md
kfox1111: emptydir is only for certain testing cases. It should not be the default. Letting the clusters chosen PVC driver be used by default is the best option. In a lot of cases, the cluster-admin already made the choice of default volume driver.

charts/spire/charts/spire-server/values.yaml
kfox1111: For all but the testing case, persistence is needed. By having it default to pvc, it handles the most common case along with making folks think about changing it should they need to. This is the right thing to do IMO.

Why put pvc under type? putting pvc related settings under persistence is the most common api.

faisal-memon:
> That works really well for just about any csi driver and the recommended path forward that direction. emtpydir and hostpath are kind of the exception to that rule.

You can create a [host path persistent volume](https://kubernetes.io/docs/tasks/configure-pod-container/configure-persistent-volume-storage/#create-a-persistentvolume).

kfox1111:
hostpath pv's could be an interesting feature to add to the chart so we can create it for the user. The implementation would be a little different from whats proposed here, but the need a type field would still be the same in that case.

charts/spire/charts/spire-server/templates/statefulset.yaml
edwbuck: Do we really need a group consensus for a refactoring call?  If it altered functionality, sure; but, this seems to be more a call to ask if we can create contention on a relatively minor review point.

charts/spire/charts/spire-server/values.yaml
edwbuck: If we don't put pvc under type, then hostpath will compete with pvc values and one might mix / match where some of the values are ignored and some are not.  By putting it under type, it is clear in the user's values.yaml file if they are mixing and matching, providing visual feedback of the oddity.

Let's say I'm not using Tornjak (sp?) and my keystore, for additional security, is in RAM.  This seems to be a rather "unexceptional" setup.  Then, the only thing I have in the volume is sqlite3, or I'm running with multiple servers, in which case even that's not in the volume.  Which then raises the question, "why are we forcing a PVC on the system, one that will require explicit reconfiguration to avoid?"  Losing a database is nothing, losing a keystore is not the end of the world either, I'm not sure what losing Tornjak entails.

faisal-memon:
> hostpath pv's could be an interesting feature to add to the chart so we can create it for the user. The implementation would be a little different from whats proposed here, but the need a type field would still be the same in that case.

Im not suggesting we create a hostpath pv. If you wants to use hostpath you can do that right now without this PR. If you just want to test without persistent storage then the flag to disable seems to be sufficient and in line with what other helm charts do. So i just don't understand what this pr is accomplishing.

kfox1111:
> Im not suggesting we create a hostpath pv. If you wants to use hostpath you can do that right now without this PR. If you just want to test without persistent storage then the flag to disable seems to be sufficient and in line with what other helm charts do. So i just don't understand what this pr is accomplishing.

That may be the source of the difference of opinions. I think being able to easily run spire on the edge with single node or 3 node configurations running on bare metal is a use case that may increase in frequency as time goes on. We do have arm support thanks to Marco, so a 1 or 3 node raspberry pi cluster makes a lot of sense to me.

If it ever is to be supported, and if we want to minimize the number of configurables, then I think Edwin is right, it might be best to do it as a enum rather then a boolean.

charts/spire/charts/spire-server/values.yaml
kfox1111: The default for upstream spire is sqlite on the equivalent of a hostpath, IE, persistence. One would not ever expect deleting a pod from a node (common on rolling upgrades of Kubernetes) to break your whole spire chain of trust because it has lost its CA cert/key and everything needs rekeying.

charts/spire/charts/spire-server/templates/statefulset.yaml
kfox1111: This pr is already blocked. Just a bit hesitent to continue to work on it if its going to get killed off. If it opens back up, then yeah, I can make the requested changes.

kfox1111:
> 
> Id rather wait till we have a solid use case in hand rather than guess now what may or may not be needed for the future.

A middle ground might be to drop the hostpath part of this pr until edge spire use cases are fleshed out. but leaves the option open to support it in the future without adding an option?

edwbuck:
@kfox1111 The fact that you submitted two PRs for the same issue is adding to the inability to get this across the line.  It makes all forward effort subject to being distracted with "well there is the other PR".

I don't see any blockers on this, beyond Marco's "well there is the other PR" and my "I'd like to default to be different, and maybe the keys to be reconfigured differently."  The first "blocker" shouldn't exist, because there shouldn't be 2 PRs for 1 issue, and the second could be settled more easily if the team wasn't confused by having two PRs.  

Please take up the work of getting the items closed.  This is close enough to being accepted that it could be in this week or sooner, as long as we keep the commentary on this PR.

Also, I've asked for the other PR to be closed once.  This is me asking for it twice.

kfox1111:
> @kfox1111 The fact that you submitted two PRs for the same issue is adding to the inability to get this across the line. It makes all forward effort subject to being distracted with "well there is the other PR".

<snip>

> Also, I've asked for the other PR to be closed once. This is me asking for it twice.

It was put in at your request.

You've asked twice, and twice I've told you, I don't own the other PR. Faisal submitted it, and I'm not comfortable closing stuff I did not submit.

charts/spire/charts/spire-server/README.md
faisal-memon: Why should it default to a value that is for testing only and not the recommended value?

charts/spire/charts/spire-server/values.yaml
edwbuck: I'll defer this to the current defaults then.

charts/spire/README.md
edwbuck: Understood.  Closing this request

charts/spire/charts/spire-server/templates/statefulset.yaml
edwbuck: @kfox1111 Please don't read into a request that the work is going to be killed off.  Prompt handling of requests is the main reason submissions get merged.

Yes, we need to work on the team's review flexibility, occasionally we all get stuck in items that really don't matter, as we are all emotional creatures inclined to align to our first impressions (right or wrong).  That said, if you lose heart, you will make your failure to get the change merged a self-fulfilling prophecy.

Please give the effort an honest amount of work.  If you really feel you are losing ground on an item, then say so explicitly, asking for the changes be limited to those that really matter, or for the review to be closed as not accepted.  But, I would hope that you don't reach for that option so soon.  In this case, I see this as merging, and most of my blocking questions are resolved now.

edwbuck (approved):

kfox1111:
I'm ok pulling the hostpath stuff for now if we want to design that more intentionally. Would that help?

charts/spire/charts/spire-server/README.md
edwbuck: I'm willing to not take a stand on this item.  I'll let it be

edwbuck:
We have two approvals on this one, and one item blocking it.

@marcofranssen please clarify / update your request, now that the whole #300 item has been decided against.
@kfox1111 please look for Marco's update, so we can put this one to bed.

kfox1111:
@edwbuck @marcofranssen not sure how to reconcile that recommendation with this pr. The suggestion, as a far as I understand it, is not to use emptydir at all, but switch to deployment if persistence is to be disabled. In this case, that would be when type=="emptydir" you wouldn't get an emptydir at all, but a deployment? That would look mostly like this pr: https://github.com/spiffe/helm-charts/pull/271 with a different conditional. If thats the way to go, should we resurrect that pr so that @marcofranssen gets the credit for his work?

marcofranssen:
> @edwbuck @marcofranssen not sure how to reconcile that recommendation with this pr. The suggestion, as a far as I understand it, is not to use emptydir at all, but switch to deployment if persistence is to be disabled. In this case, that would be when type=="emptydir" you wouldn't get an emptydir at all, but a deployment? That would look mostly like this pr: #271 with a different conditional. If thats the way to go, should we resurrect that pr so that @marcofranssen gets the credit for his work?

Have been thinking a bit more about this:

If you don't require persistence and you only need temporary storage for your application, using a Deployment with an emptyDir volume can be a suitable choice.

A Deployment in Kubernetes is primarily designed for stateless applications, where each Pod created by the Deployment is considered interchangeable. An emptyDir volume can be used within a Deployment to provide temporary storage that is shared among all Pods managed by the Deployment.

The advantage of using a Deployment with an emptyDir volume is that it allows you to easily scale and manage replicas of your application without the need for persistent storage. The emptyDir volume will be created and mounted in each Pod of the Deployment, allowing them to share temporary data as needed.

On the other hand, a StatefulSet is more suitable when you have stateful applications that require unique network identities, stable hostnames, or persistent storage. StatefulSets provide guarantees for ordering and uniqueness of Pods, which can be important for applications that rely on stable network identities or require specific initialization or scaling behaviors.

In summary, if your application is stateless and doesn't require persistence, using a Deployment with an emptyDir volume is a lightweight and straightforward option. However, if your application is stateful or requires specific ordering or unique identities, a StatefulSet might be a more appropriate choice.

So IMHO we should take #271 if you want to disable persistence we simply switch to the deployment with an emptyDir mount. Otherwise we use the Statefulset with PVC. Then you have both of best worlds.

kfox1111:
> The advantage of using a Deployment with an emptyDir volume is that it allows you to easily scale and manage replicas of your application without the need for persistent storage. The emptyDir volume will be created and mounted in each Pod of the Deployment, allowing them to share temporary data as needed.

A StatefulSet can do the same thing. A StatefulSet need not have a volumeClaimTemplate.
The disadvantage is we need to template out a whole other kind of object, supporting a lot of extra code.

If it was a primary use case I'd say, yeah, supporting a Deployment would be good. In this case, I'm uncertain how much it really benefits us.

The exception might be, if we wanted to support a DaemonSet too. I could see wanting to do that for the Edge use case. If we do that, then we have to template out the PodSpec anyway, and doing it also for Deployment is not such an additional ongoing burden. What do you think to adding both Deployment and DaemonSet?

edwbuck:
> If you don't require persistence and you only need temporary storage for your application, using a Deployment with an emptyDir volume can be a suitable choice.

This is about persistence of the items, where no persistence can be handled with EmptyDir.

We are not expanding the scope of this issue to change StatefulSets to Deployments.  It may be a good idea, but it needs an independent issue and should not be riding the coat tails of this PR.  https://github.com/spiffe/helm-charts/issues/372 will cover that effort.  Please unblock the issue which is about persistence and re-review with a tighter scope of only addressing the changes are sufficient to deal with persistence.

marcofranssen:
> > If you don't require persistence and you only need temporary storage for your application, using a Deployment with an emptyDir volume can be a suitable choice.
> 
> This is about persistence of the items, where no persistence can be handled with EmptyDir.
> 
> We are not expanding the scope of this issue to change StatefulSets to Deployments. It may be a good idea, but it needs an independent issue and should not be riding the coat tails of this PR. #372 will cover that effort. Please unblock the issue which is about persistence and re-review with a tighter scope of only addressing the changes are sufficient to deal with persistence.

https://github.com/spiffe/helm-charts/pull/271 was also all about the persistence. This is the way most charts out there are implementing this feature. 

Have you also ready this part? https://github.com/spiffe/helm-charts/pull/315#issuecomment-1595682612

This also explains why you want to use a deployment if you don't require persistence (emptyDir). So no I feel we should not just simply merge this.

kfox1111:
@marcofranssen what do you think of https://github.com/spiffe/helm-charts/pull/315#issuecomment-1607800507?

marcofranssen:
> @marcofranssen what do you think of [#315 (comment)](https://github.com/spiffe/helm-charts/pull/315#issuecomment-1607800507)?

we could add indeed in the future a PR to also add DaemonSet, but is that really usable for prod as a daemonset won't be able to do the persistent volumes right?

For now I would just use the types where they are intended for. Deployment no persistence, Statefulset => persistence.

kfox1111:
> we could add indeed in the future a PR to also add DaemonSet, but is that really usable for prod as a daemonset won't be able to do the persistent volumes right?

Thinking something like, having a 3 node rpi cluster where each node is a control plane node. Running the spire-server as a daemonset with type set to hostPath

edwbuck:
> > we could add indeed in the future a PR to also add DaemonSet, but is that really usable for prod as a 
> > daemonset won't be able to do the persistent volumes right?

> Thinking something like, having a 3 node rpi cluster where each node is a control plane node. Running the 
> spire-server as a daemonset with type set to hostPath

DaemonSets (typically used for the agents) can use persistent volumes, and generally do use them when a person doesn't want to put the KeyStore in memory.  It is rare, but occasionally an installation benefits from on-disk keystores, especially in QA/evaulation scenarios where direct inspection of the cached certificates in the agents are used to validate (internal) SPIRE functionality. 

With DaemonSets (which contain a Pod declaration) one puts the persistent volume in the contained Pod template, and in there it too can be set to hostPath, emptyDir, or some CSI backed PVC.

In short, we need to have the setting be 2 settings, one for the storage type of the Server, and one for the Storage type of the Agent, as the storage types between those different processes are not likely to be the same, but are likely to be the same among the Servers or the same among the Agents.

To clarify in Kevin's 3 node cluster, we might have the persistence type for the servers be "hostPath" to permit the Servers faster recovery by allowing them to re-find their previously existing signing certs, while we would indicated that Agents use "emptyDir" persistence types because they effectively would need to reload everything anyway.  For clusters that have only one server among the three nodes, one could use affiinity / anti-affinity to ensure that the same node is used for all server launches.  If this 3 node cluster had access to a NFS server, then one could use a PVC claim instead of "hostPath" and bind the storage to a NFS path.  Likewise, if more sophisticated CSI drivers are available, one could bind to Ceph, MapRFS, or whatever.

If the restart performance doesn't matter, one could also configure all of the systems to use "emptyDir" storage and have the intermediate CA's used for signing acquired fresh with each restart.  In this case, the trade off would be longer restarts and more signing, which might be acceptable should the improved security of never seeing the KeyStores stored in a persistent way.

So, selecting the persistence type provides the most flexibility, but we need to have two settings: one to control the persistence type of the servers, and one to control the persistence type of the agents.  If we want to expand this pattern to Tornjak or other items, I think that those tasks should be handled in a separate PR.

faisal-memon:
You can specify HostPath through a Local Path Provisioner. Thats what K3s does. K3s runs on the rpi and probably what id use to get a cluster going on one. I havent tested on rpi, but I think it should work. So why does HostPath need special handling?

kfox1111:
The local path provisioner works, but when you use that one, care still needs to be taken in some configurations to anit-affinity or daemonset the server to ensure you don't get all instances pile up on the same physical node, reducing the ability to HA the deployment.

faisal-memon:
> The local path provisioner works, but when you use that one, care still needs to be taken in some configurations to anit-affinity or daemonset the server to ensure you don't get all instances pile up on the same physical node, reducing the ability to HA the deployment.

How does this PR avoid that scenario?

kfox1111:
It doesnt. Just stating it doesn't add anything I think that this PR doesn't already do? At this point I've kind of lost all track of what is being discussed for what reasons. Just trying to answer questions

v0lkan:
Folks,

While I am writing this personally and not on behalf of the SPIFFE Steering Committee, I want to express a concern that **the overall health of this project has been drawing SSC’s consistent attention**.

As a reminder, the SPIFFE Steering Committee (or SSC) is a group of volunteers that oversee the strategic direction and growth of all projects falling under the SPIFFE umbrella. It is the only group that presides over the full spectrum of SPIFFE-related work. [You can read more about the mission, vision, and charter of SSC here](https://github.com/spiffe/spiffe/blob/main/ssc/CHARTER.md)

As we strive for progress and improvements, I wanted to bring to your attention not only this PR but also other outstanding PRs that have been stagnating for a long while.

Specifically, this PR has been open since May 25th. I understand the complexity of the matter and appreciate the in-depth discussions and voting that took place to ensure the best possible outcome. 

It is evident from the conversations that a substantial amount of thought and effort has gone into resolving the disputes around this PR.

As it currently stands, all outstanding discussions appear to be addressed satisfactorily, and all checks are passing, indicating the readiness of this PR for integration. 

In light of these, and to keep up with our momentum, it would be beneficial for us to move forward with the merging process. 

I suggest we proceed, as doing so will allow us to close this chapter and shift our focus toward new objectives and endeavors.

I would like to remind us once more that it is equally important for us to remain cognizant of the broader project cadence and our commitment to deadlines: 

While detailed discussion and extensive scrutiny are crucial, they should not lead to a loop of constant deliberation at the cost of progression. All conversations around this PR have been duly resolved, and all checks are currently passing, making it ready for integration.

I want to note that our work on this project is more than a hobby or side project. It's a commitment under the guidance of the SPIFFE Steering Committee, and the project remains under the spiffe GitHub organization. 

Stagnation in PRs such as this can inadvertently impact our ability to manage releases effectively and adhere to set timelines. We must balance our attention between resolving current issues and advancing toward our future objectives.

For this reason, I propose we aim to increase our project cadence by setting definitive goals and timelines for decision-making processes. By doing so, we can ensure productive discussions while still maintaining momentum and meeting our obligations.

Thank you for your understanding and consideration. 

I eagerly look forward to further collaborative efforts and progression.

With respect,

Volkan.

PR #318, opened 2023-05-26
Removed pull_request edited event from ci workflow

marcofranssen:
Edited event is only useful if you want a workflow to act on a PR description or title or labels and such, we are not doing that in this workflow.

Removing the edited type from the event doesn't unnecessarily run the workflow again when title, description and such are changed.


kfox1111 (approved):
LGTM

faisal-memon (approved):

PR #327, opened 2023-05-30
Add Makefile for local testing

marcofranssen:


Makefile
```diff
@@ -0,0 +1,21 @@
+.PHONY: help
+help: ## Display this help.
```
mrsabath, line 2: We probably need a better description when to run this test: before, after, or instead of the deployment via `helm install`

marcofranssen: Added some words in CONTRIBUTING.md + added small improvement on test task documentation

mrsabath (approved):
Tested on a brand-new _kind_ cluster. It works great. Thank you @marcofranssen for this handy tool! 
LGTM

Makefile
```diff
@@ -0,0 +1,22 @@
...
+	@$(MAKE) help
+	@awk 'BEGIN {FS = ":.*##"; printf "\nUsage:\n  make \033[36m<target>\033[0m\n"} /^[a-zA-Z_0-9-]+:.*?##/ { printf "  \033[36m%-15s\033[0m %s\n", $$1, $$2 } /^##@/ { printf "\n\033[1m%s\033[0m\n", substr($$0, 5) } ' $(MAKEFILE_LIST)
+
+
+##@ Testing:
+
+.PHONY: test
+test: ## Run tests using Helm chart-testing (ensure to run on dedicated test cluster)
```
kfox1111, line 10: Maybe consider naming this target test-and-clean or something similar?

marcofranssen: the cleaning parts are only there to make it resilient, when testing I added the option --no-clean-up, and then you end up with lingering stuff, I then decided to keep those in there in case of failures to make it still work.

I'll refactor it to an undocumented cleaning task. As the clean shouldn't be relevant for the end-user.

charts/spire/Chart.lock
```diff
@@ -0,0 +1,18 @@
...
+  repository: file://./charts/spiffe-csi-driver
+  version: 0.1.0
+- name: spiffe-oidc-discovery-provider
+  repository: file://./charts/spiffe-oidc-discovery-provider
+  version: 0.1.0
+- name: tornjak-frontend
+  repository: file://./charts/tornjak-frontend
+  version: 0.1.0
+digest: sha256:0119361a3dd54d1cf088f299ce9adbee55a96c832cb675f57e5e115acf7b17c7
+generated: "2023-05-30T19:28:59.334691+02:00"
```
kfox1111, line 18: Did the Chart.lock file get added by mistake?

kfox1111 (approved):
LGTM

PR #331, opened 2023-06-03
Initial submission of Helm Chart key naming conventions.

### LINKED ISSUE #329: Document Helm key naming conventions.
We need to establish conventions to simplify the review of helm key naming conventions.  The initial set likely should contain

1.     Proper names are to be represented in the case they come in, preserving all "Helm supported" characters per Helm guidelines, with unsupported / problematic characters replaced by underscores. The rational is that it helps one identify the item, increasing usability. (This replaces the hyphen '-' with underscores, making some items look like snake case; but they aren't snake case, for example freedom-FooFigthers would be freedom_FooFighters, not freedom_Foo_Fighters).
 
2.     All other names are to be snake case, starting with a lower case letter, per Helm's documented general conventions.
 
3.     All compound words including Acronyms (NASA, for National Aeronautics and Space Administration) and Initialisms (NFL, for National Football League) are to maintain an all-uppercase spelling, in accordance with the English language conventions. If the camel case causes confusion "NASARAM", rename the word, ideally putting any still-existing Acronym or Initialism at the end "memoryNASA"
 
4.     All compound words including abbreviations (shortening of words) will have the abbreviation in all lowecase, with the exception of when camel case indicates the abbreviation would be capitalized. Examples include apptBuilding, destinationAppt, usersGroupsEtc.
 
5.     If an abbreviation, initialisizm, or acronym would noromally be used with periods in the name (for example P.M.) we remove the periods without supplying an "alternate" character to maintain the spaceing (P.M. becomes PM, not P_M_ or P_M)

The scope of this is to document some general guidelines, not to have a comprehensive rule chart that can cover all of the cases.  I think that the above will be sufficient, and it should go into the projects/ folder under conventions.md

edwbuck:
Closes #329

### DIFF
--- a/project/conventions.md
+++ b/project/conventions.md
@@ -62,7 +62,7 @@ characters, even when such characters are supported by Helm.  ASCII
 letters, upper and lower case, ASCII numerals, and the ASCII underscore
 character are always permitted.  
 
-Other characters, such as `@`, `~`, `-`, `&`, `+` are often used as
+Other characters, such as `@`, `~`, `&`, `+` are often used as
 shorthand to represent, respectively `at`, `similar / not`, 
 `minus / without`, `and / reference`, `plus / with`.  Helm key elements
 MUST NOT contain such symbols, as it burdens the reader to deduce the
--- a/project/conventions.md
+++ b/project/conventions.md
@@ -1,0 +1,108 @@
+<!-- vim: ft=markdown colorcolumn=72
+-->
+# Development Conventions
+
+This document intends to capture the current coding conventions for
+the SPIRE Helm Chart project, permitting multiple developers to align
+to a similar style and standard when offering their contributions.
+
+It is impossible to describe every scenario a developer might encounter,
+likewise it is equally impossible to describe the best choice to make
+under all future circumstances.  For this reason, many conventions also
+have a small explanation of the rationale behind the guidance, allowing
+future development efforts to weigh and balance the conflicting forces
+in ambiguous situations not clearly covered.
+
+This document is intended to be a living document.  If you have an issue
+with a convention, we hope that you will create an issue describing the
+problem, and a pull request suggesting the update.  By addressing the
+convention, we hope to avoid arguments about conventions in the pull
+request review process, which slow the process and are often subject to
+the popular whims of the moment.
+
+## Conventions used in this document
+
+While this is not a software specification, the key words "MUST",
+"MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "SHOULD NOT",
+"RECOMMENDED",  "MAY", and "OPTIONAL" in this document are to be
+interpreted as closely to the convention as it would be to a
+specification described in RFC 2119.
+
+Use of the word in its non-all caps form SHOULD NOT be interpreted in
+the strict definitions user RFC 2119, but SHOULD be interpreted in
+context with a meaning aligned to RFC 2119.
+
+## Helm Chart conventions
+
+Helm values are organized in a tree structure, with each entry having
+a key, the path within the structure, and a value, the contents held
+at that path.  Some of the contents include objects and arrays, which
+are structures that permit multiple values under a particular path.
+
+This design choice mimics the representation of a file system, and we
+may use file system analogies to describe the conventions and choices
+in the project.
+
+### Preference for absolute paths in templates
+
+Outside of looping constructs that require relative paths, templates
+should use absolute paths (paths starting with the `.` character) for
+their value references.  
+
+### Helm Key path elements
+
+Elements refer to items in the configuration.  Some of those items are
+external to the project, while others are components the project choose
+to implement the Helm Chart packaging solution.
+
+### Preferred characters for helm path elements
+
+Elements within a helm chart path SHOULD limit the use of special
+characters, even when such characters are supported by Helm.  ASCII
+letters, upper and lower case, ASCII numerals, and the ASCII underscore
+character are always permitted.  
+
+Other characters, such as `@`, `~`, `-`, `&`, `+` are often used as
+shorthand to represent, respectively `at`, `similar / not`, 
+`minus / without`, `and / reference`, `plus / with`.  Helm key elements
+MUST NOT contain such symbols, as it burdens the reader to deduce the
+intent of the writer.
+
+### Camel Case for created elements.
+Elements that refer to components that the project has created should
+exclusively use camel case, with a the first word being lower case.  To
+illustrate, a "card sort order" path element would read "cardsharper".
+
+### Acronyms and Minimalists
+
+Occasionally the use of an Acronym or Minimalist is used as part of an
+element name.  An acronym is the use of the first letter of a phrase
+or multi-word proper name that is pronounceable, while an initialisim is
+the use of the first letter of a phrase or multi-word proper name 
+pronounced by naming the letters.  `NASA` is an example of an acronym,
+while `NFL` is an example of an Initialisim.
+
+To keep the recognition of the Acronym or Initialisim, an all capital
+representation of the Acronym or Initialisim MUST be maintained.  To
+reduce the confusion this can create in camel case elements, we recommend
+that such Acronyms and Initialisim be placed at the end of the element
+or the element is restructured to avoid confusion.  For example, 
+`mailingListNASA` or `mailingList.NASA` is preferred over
+`NASAMailingList`.
+
+Acronyms and Initialisim occasionally use periods in their presentation.
+When they do, the periods MUST be removed without replacement by
+another character.  To illustrate `U.S.S.R.` should be written as
+`USSR`.
+
+### Abbreviations
+
+Abbreviations shorten words by using some of the letters of the original
+word.  In all cases, element names that contain parts which are
+abbreviations should be represented in the case determined by the part's
+position.  For example, a Junior administrator, should be written as
+`jrAdmin`.
+
+Abbreviations occasionally use periods in their presentation.  When they
+do, the periods MUST be removed without replacement by another character.
+To illustrate, use `acctSchedule` for Acct. Schedule.
--- a/project/conventions.md
+++ b/project/conventions.md
@@ -74,6 +74,17 @@ Elements that refer to components that the project has created should
 exclusively use camel case, with a the first word being lower case.  To
 illustrate, a "card sort order" path element would read "cardSortOrder".
 
+### Exact Case for non-created elements
+
+Elements that reflect components that have well known names should make
+every effort to use the well known name exactly, even if the rules for
+captialization don't follow other naming conventions.  Exceptions to
+this SHOULD include replacing problematic characters with underscores,
+to improve template functionality.  To illustrate, the operating system 
+NeXTSTEP would read as "NeXTSTEP" and "NeXT Computers" SHOULD be written
+as "NeXT_Computers" but may also be written as "NeXTComputers" if clarity
+is maintained.
+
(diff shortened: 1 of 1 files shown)


project/conventions.md
```diff
@@ -0,0 +1,108 @@
...
+Elements within a helm chart path SHOULD limit the use of special
+characters, even when such characters are supported by Helm.  ASCII
+letters, upper and lower case, ASCII numerals, and the ASCII underscore
+character are always permitted.  
+
+Other characters, such as `@`, `~`, `-`, `&`, `+` are often used as
+shorthand to represent, respectively `at`, `similar / not`, 
+`minus / without`, `and / reference`, `plus / with`.  Helm key elements
+MUST NOT contain such symbols, as it burdens the reader to deduce the
+intent of the writer.
```
kfox1111, line 69: '-' is often used as ' ' as well, and one of the few items that is dns compatible for naming k8s objects. in some cases its one of the few options and we should be a little careful preventing it.

project/conventions.md
```diff
@@ -0,0 +1,108 @@
...
+
+### Camel Case for created elements.
+Elements that refer to components that the project has created should
+exclusively use camel case, with a the first word being lower case.  To
+illustrate, a "card sort order" path element would read "cardsharper".
+
```
kfox1111, line 75: We're configuring spire options. maybe some discussion about when it is appropriate to use the option as specified to spire and when to follow a stricter camel casing scheme. For example, plugin names.

kfox1111:
Thanks for working on this. It will be increasingly important as we try and cleanup the api.

project/conventions.md
```diff
@@ -0,0 +1,108 @@
...
+Other characters, such as `@`, `~`, `-`, `&`, `+` are often used as
+shorthand to represent, respectively `at`, `similar / not`, 
+`minus / without`, `and / reference`, `plus / with`.  Helm key elements
+MUST NOT contain such symbols, as it burdens the reader to deduce the
+intent of the writer.
+
+### Camel Case for created elements.
+Elements that refer to components that the project has created should
+exclusively use camel case, with a the first word being lower case.  To
+illustrate, a "card sort order" path element would read "cardsharper".
```
faisal-memon, line 74:
```suggestion
illustrate, a "card sort order" path element would read "cardSortOrder".
```

project/conventions.md
```diff
@@ -0,0 +1,108 @@
...
+pronounced by naming the letters.  `NASA` is an example of an acronym,
+while `NFL` is an example of an Initialisim.
+
+To keep the recognition of the Acronym or Initialisim, an all capital
+representation of the Acronym or Initialisim MUST be maintained.  To
+reduce the confusion this can create in camel case elements, we recommend
+that such Acronyms and Initialisim be placed at the end of the element
+or the element is restructured to avoid confusion.  For example, 
+`mailingListNASA` or `mailingList.NASA` is preferred over
+`NASAMailingList`.
```
faisal-memon, line 91: I see in the spire code base alot when the acronym is in front its all lower case. Not suggesting to or to not adopt that approach, just giving a point of reference.

project/conventions.md
```diff
@@ -0,0 +1,108 @@
...
+
+Acronyms and Initialisim occasionally use periods in their presentation.
+When they do, the periods MUST be removed without replacement by
+another character.  To illustrate `U.S.S.R.` should be written as
+`USSR`.
```
faisal-memon, line 96: Suggest not to use USSR as it no longer exists.

project/conventions.md
```diff
@@ -0,0 +1,108 @@
...
+letters, upper and lower case, ASCII numerals, and the ASCII underscore
+character are always permitted.  
+
+Other characters, such as `@`, `~`, `-`, `&`, `+` are often used as
+shorthand to represent, respectively `at`, `similar / not`, 
+`minus / without`, `and / reference`, `plus / with`.  Helm key elements
+MUST NOT contain such symbols, as it burdens the reader to deduce the
+intent of the writer.
+
+### Camel Case for created elements.
```
marcofranssen, line 71:
```suggestion
### Camel Case for created elements

```

accordiing to markdownlint

project/conventions.md
edwbuck: @kfox1111 there's an entire second paragraph that talks about items we don't control the naming of, like plugin names.  I think your comment is sufficiently handled there, where the answer is "make it nearly identical to an external name, including matching capitalization, only replacing spaces and hyphens with underscores (for obvious reasons of following Helm's own documented best practices)"

project/conventions.md
kfox1111: They are more difficult to deal with in helm templates, yes. But they will be unavoidable in cases where nested charts may be used or in cases where we want to use the same key as the plugin name, for example, from the spire settings. So, 'MUST NOT' is probably too strong a wording with regard to '-'.

project/conventions.md
edwbuck: Faisal, generally acronyms are never lower case, but sometimes people to stuff like `tcpPort`, so there's the occasional item that is both an acronym and improperly pushed to lower case.  I thought we'd just address it beforehand, considering I had to write rules for all of the other upper/lower case scenarios.

project/conventions.md
edwbuck: Sure, give me another example of an Acronym that is typically spelled with periods.

And I'm comfortable referring to stuff that existed in the past, as long as it's not in a context that implies it exists today (when it doesn't); but, I really would like a better example too.  I just can't think of a common one to illustrate the point.

project/conventions.md
edwbuck: @kfox1111 These aren't laws.  They're conventions.  If you can get everyone on board with an exception, then that's fine by me.  Trying to make them into laws will only mean we will never have a guideline document that says what we prefer / desire; because it will be littered with far too many rules to be followed.

project/conventions.md
kfox1111: This? https://github.com/spiffe/helm-charts/pull/331/files#diff-f0ed57313266c949a152fa5790e72e4b36629acf9ae273fe84ac48f779bef52eR9-R14

Its pretty vague with regard to plugin names.

I ask because of this:
https://github.com/spiffe/helm-charts/pull/227/files#diff-44e90d1febb559dfb70b0529bd6b5bc4bc0c51ed995407d2314f4aaf1062cac4L3-R3

The plugin name has a '-' in it, and would be manual code to try and map it in that case. not sure its worth one offing random plugin names into a _ form.

project/conventions.md
kfox1111: Just trying to understand. There was seemly great care paid attention to specify terms to match RFC 2119, then the strongest form used MUST NOT used rather then, say, "SHOULD NOT', which clearly articulates that there is room for reasonable exceptions.

project/conventions.md
mrsabath: U.S., U.N. as United Nations

project/conventions.md
edwbuck: **If you think I used the wrong phrase, I'm open to a request to change it.**

I personally don't want symbols that have two meanings, as `message2Route` can always be represented as `messageTooRoute` or `messageTwoRoute` which makes the context of the previous `2` obvious.  The same goes for the other symbols above.  If you want to open the door to them sometimes being used, I'm ok with that, getting a document into the repo we can reference more important.

After all, we can always then do an Issue to clarify / change / rewrite items we disagree with, as we find the scenarios that make the current document insufficient.  But without a document to change, there's just nothing to refine.  That said, I'd want the document that goes in to be mostly ok with the team, it has no value if nobody agrees to it.

project/conventions.md
edwbuck: I get that you are uncomfortable with the suggestion, but I don't have a clue as to what you are suggesting in it's place.  Is there a better idea you want to put on the table?  I'm all ears.

We could do "absolutely the same word, no replacements", we could permit hyphens, disregarding the Helm Chart's own published best practices, we could make - the replacement character, or we could do something else.  

If you can't decide among the alternatives, I'd suggest that until "something better comes along" let's go with this till an Issue is filed due to an actual problem that we need fixed.

project/conventions.md
edwbuck: @kfox1111 If I wasn't clear on how I would like the request to change it.  Use the review tools to provide the line as you want it written, so I don't have to guess at the desired edit a few times.

project/conventions.md
kfox1111: maybe an explicit sentence to the effect of:

"When the chart is directly referencing properties of the spire config file, such as plugin names that are not easy to control, its acceptable to use the string verbatim rather then try and convert it into camel case in values."

There are likely better wordings then that.

project/conventions.md
edwbuck: @faisal-memon If this comment is good enough for your needs, please resolve this conversation.

project/conventions.md
edwbuck: @kfox1111 In rereading this I see the main point was the strength of the guideline, not the content of the guideline.

I'm of the mind that for every rule, there is a scenario where an exception can reasonably occur.  Because of this viewpoint, I don't try to make rules for others absolute.  If you want this to be absolute, I can change it (but I personally think people will do the right thing without it being made absolute).

Do you want it changed?  It's not a big deal to me.

project/conventions.md
```diff
@@ -0,0 +1,120 @@
...
+to implement the Helm Chart packaging solution.
+
+### Preferred characters for helm path elements
+
+Elements within a helm chart path SHOULD limit the use of special
+characters, even when such characters are supported by Helm.  ASCII
+letters, upper and lower case, ASCII numerals, and the ASCII underscore
+character are always permitted.  
+
+Other characters, such as `@`, `~`, `-`, `&`, `+` are often used as
```
kfox1111, line 65:
```suggestion
Other characters, such as `@`, `~`, `&`, `+` are often used as
```

project/conventions.md
```diff
@@ -0,0 +1,120 @@
...
+shorthand to represent, respectively `at`, `similar / not`, 
+`minus / without`, `and / reference`, `plus / with`.  Helm key elements
+MUST NOT contain such symbols, as it burdens the reader to deduce the
+intent of the writer.
```
kfox1111, line 69:
```suggestion
intent of the writer.

`-` SHOULD NOT be used when there are good alternatives. Its reasonable to use when spire itself uses it for config such as plugin names or required by helm for child charts.
```

project/conventions.md
```diff
@@ -0,0 +1,122 @@
...
+letters, upper and lower case, ASCII numerals, and the ASCII underscore
+character are always permitted.  
+
+Other characters, such as `@`, `~`, `&`, `+` are often used as
+shorthand to represent, respectively `at`, `similar / not`, 
+`minus / without`, `and / reference`, `plus / with`.  Helm key elements
+MUST NOT contain such symbols, as it burdens the reader to deduce the
+intent of the writer.
+
+`-` SHOULD NOT be used when there are good alternatives. Its reasonable to use when spire itself uses it for config such as plugin names or required by helm for child charts.
```
kfox1111, line 71: My apologies. I didn't realize I forgot to manually wrap the text.

kfox1111 (approved):
Just the line wrapping issue. Everything else LGTM.

project/conventions.md
edwbuck: No worries.  Wrapping the text is a "nice to have" in my mind.  Content is a "need to have".

faisal-memon (approved):

PR #338, opened 2023-06-09
Add TLS/mTLS support for Tornjak

mrsabath:
This PR adds TLS and mTLS connection to Tornjak backend. It is assumed users create secret `tornjak-certs` prior to deploying the helm charts when TLS and/or mTLS are enabled. 

Sample secret (to be documented): 
```
kubectl -n spire-server create secret generic tornjak-certs \
--from-file=key.pem="client.key"  \
--from-file=cert.pem="client.crt" \
--from-file=rootCA.pem="CA/rootCA.crt"

# testing TLS: 
curl --cacert CA/rootCA.crt https://localhost:20000/

# testing mTLS: 
curl  --cacert CA/rootCA.crt --key client.key --cert client.crt https://localhost:30000
```

### DIFF
--- a/charts/spire/charts/spire-server/README.md
+++ b/charts/spire/charts/spire-server/README.md
@@ -149,10 +149,15 @@ A Helm chart to install the SPIRE server.
 | tornjak.config.http.enabled | bool | `true` | Enables Tornjak HTTP (insecure) service |
 | tornjak.config.http.port | int | `10000` | Container port value for HTTP |
 | tornjak.config.http.service | object | `{"annotations":{},"port":10000,"type":"ClusterIP"}` | Service to handle Tornjak HTTP connection |
-| tornjak.config.mtls.enabled | bool | `true` | Enables Tornjak TLS service |
+| tornjak.config.mtls.ca | string | `"/opt/spire/sample-keys/rootCA.pem"` | CA for mTLS connection |
+| tornjak.config.mtls.cert | string | `"/opt/spire/sample-keys/cert.pem"` | Certificate for mTLS connection |
+| tornjak.config.mtls.enabled | bool | `false` | Enables Tornjak TLS service |
+| tornjak.config.mtls.key | string | `"/opt/spire/sample-keys/key.pem"` | Key for mTLS connection |
 | tornjak.config.mtls.port | int | `30000` | Container port value for mTLS |
 | tornjak.config.mtls.service | object | `{"annotations":{},"port":30000,"type":"ClusterIP"}` | Service to handle Tornjak mTLS connection |
-| tornjak.config.tls.enabled | bool | `true` | Enables Tornjak TLS service |
+| tornjak.config.tls.cert | string | `"/opt/spire/sample-keys/cert.pem"` | Certificate for TLS connection |
+| tornjak.config.tls.enabled | bool | `false` | Enables Tornjak TLS service |
+| tornjak.config.tls.key | string | `"/opt/spire/sample-keys/key.pem"` | Key for TLS connection |
 | tornjak.config.tls.port | int | `20000` | Container port value for TLS |
 | tornjak.config.tls.service | object | `{"annotations":{},"port":20000,"type":"ClusterIP"}` | Service to handle Tornjak TLS connection |
 | tornjak.enabled | bool | `false` | Deploys Tornjak API (backend) (Not for production) |
--- a/charts/spire/charts/spire-server/README.md
+++ b/charts/spire/charts/spire-server/README.md
@@ -35,11 +35,11 @@ Tornjak is the UI and Control Plane for SPIRE [https://github.com/spiffe/tornjak
 * Backend (this chart) - Tornjak APIs that extend SPIRE APIs with Control Plane functionality
 * [Frontend](../tornjak-frontend/README.md) - Tornjak UI
 
-When Tornjak is enabled, it is exposed on both http and https (if TLS server certs are configured). Tornjak handles a permanent redirect from `http` to `https` to ensure users always use the https endpoint.
+When Tornjak is enabled, it can be reached in one of the three connection types:
 
-In addition to that you can configure a `client certificate authority`, this will make Tornjak backend verify Client certificates signed by this authority to enable mTLS authentication.
-
-> **Warning**: For production ensure to configure the certificates including CA, to protect Tornjak from unauthorized access.
+* HTTP - simple HTTP connection recommended for demonstration purpose only. It should not be use for production.
+* TLS - enables Transport Layer Security (TLS) with HTTPS communication. Tornjak client can validate Tornjak server using TLS Certificate (see instructions below)
+* mTLS - mutual TLS. Both, Tornjak client and the server can validate each other using provided certificates (see instructions below)
 
 Tornjak automatically determines the connection type based on provided information. See below.
 
--- a/charts/spire/charts/spire-server/README.md
+++ b/charts/spire/charts/spire-server/README.md
@@ -145,18 +145,11 @@ A Helm chart to install the SPIRE server.
 | tools.kubectl.image.tag | string | `""` | Overrides the image tag |
 | tools.kubectl.image.version | string | `""` | This value is deprecated in favor of tag. (Will be removed in a future release) |
 | topologySpreadConstraints | list | `[]` |  |
+| tornjak.config.connectionType | string | `"http"` | Tornjak supports 3 connection types: `http`, `tls`, and `mtls`. Select only one. |
 | tornjak.config.dataStore | object | `{"driver":"sqlite3","file":"/run/spire/data/tornjak.sqlite3"}` | Persistent DB for storing Tornjak specific information |
-| tornjak.config.http.enabled | bool | `true` | Enables Tornjak HTTP (insecure) service |
-| tornjak.config.http.port | int | `10000` | Container port value for HTTP |
-| tornjak.config.http.service | object | `{"annotations":{},"port":10000,"type":"ClusterIP"}` | Service to handle Tornjak HTTP connection |
-| tornjak.config.mtls.enabled | bool | `false` | Enables Tornjak TLS service. When 'true', the 'serverSecret' and 'userSecret' must be created prior to installing this chart |
-| tornjak.config.mtls.port | int | `30000` | Container port value for mTLS |
-| tornjak.config.mtls.service | object | `{"annotations":{},"port":30000,"type":"ClusterIP"}` | Service to handle Tornjak mTLS connection |
-| tornjak.config.serverSecret | string | `"tornjak-server-secret"` | Name of the secret containing server side key and certificate for TLS verification |
-| tornjak.config.tls.enabled | bool | `false` | Enables Tornjak TLS service. When 'true', the 'serverSecret' must be created prior to installing this chart |
-| tornjak.config.tls.port | int | `20000` | Container port value for TLS |
-| tornjak.config.tls.service | object | `{"annotations":{},"port":20000,"type":"ClusterIP"}` | Service to handle Tornjak TLS connection |
-| tornjak.config.userSecret | string | `"tornjak-user-secret"` | Name of the secret containing user CA for mTLS verification |
+| tornjak.config.serverSecret | string | `"tornjak-server-secret"` | Name of the secret containing server side key and certificate for TLS verification (required for `tls` or `mtls` connectionType) |
+| tornjak.config.service | object | `{"annotations":{},"port":10000,"type":"ClusterIP"}` | Enables the service for a given `connectionType` |
+| tornjak.config.userSecret | string | `"tornjak-user-secret"` | Name of the secret containing user CA for mTLS verification (required for `mtls` connectionType) |
 | tornjak.enabled | bool | `false` | Deploys Tornjak API (backend) (Not for production) |
 | tornjak.image.pullPolicy | string | `"IfNotPresent"` | The Tornjak image pull policy |
 | tornjak.image.registry | string | `"ghcr.io"` | The OCI registry to pull the Tornjak image from |
--- a/charts/spire/charts/spire-server/README.md
+++ b/charts/spire/charts/spire-server/README.md
@@ -37,13 +37,13 @@ Tornjak is the UI and Control Plane for SPIRE [https://github.com/spiffe/tornjak
 
 When Tornjak is enabled, it can be reached in one of the three connection types:
 
-* http - simple HTTP connection recommended for demonstration purpose only. It should not be use for production.
-* tls - enables Transport Layer Security (TLS) with HTTPS communication. Tornjak client can validate Tornjak server using TLS Certificate (see instructions below)
-* mtls - mutual TLS. Both, Tornjak client and the server can validate each other using provided certificates (see instructions below)
+* HTTP - simple HTTP connection recommended for demonstration purpose only. It should not be use for production.
+* TLS - enables Transport Layer Security (TLS) with HTTPS communication. Tornjak client can validate Tornjak server using TLS Certificate (see instructions below)
+* mTLS - mutual TLS. Both, Tornjak client and the server can validate each other using provided certificates (see instructions below)
 
-### Tornjak with TLS
+Tornjak automatically determines the connection type based on provided information. See below.
 
-`tornjak.config.connectionType=tls`
+### Tornjak with TLS Connection Type
 
 TLS connection requires Tornjak to have access to TLS key and certificate.
 Complete instruction on creating your own TLS certificate can be found [here](https://github.com/spiffe/tornjak/blob/main/examples/tls_mtls/README.md).
@@ -59,9 +59,7 @@ Once the charts are deployed, you can test the TLS connection with the following
 curl --cacert CA/rootCA.crt https://localhost:10443
 ```
 
-### Tornjak with mTLS
-
-`tornjak.config.connectionType=mtls`
+### Tornjak with mTLS Connection Type
 
 mTLS connection allows Tornjak server validation by client and Tornjak client validation by Tornjak server. The server validation is identical to above TLS. Follow the steps to create
 TLS secret with key and the certificate.
@@ -80,6 +78,10 @@ Once the charts are deployed, you can test the mTLS connection with the followin
 ```console
 curl  --cacert CA/rootCA.crt --key client.key --cert client.crt https://localhost:10443
 ```
+
+### Tornjak with HTTP Connection Type
+
+In order to run Tornjak with simple HTTP Connection only, make sure you don't create any `Secrets` or `ConfigMaps` listed above.
 
 ## Values
 
@@ -198,7 +200,6 @@ curl  --cacert CA/rootCA.crt --key client.key --cert client.crt https://localhos
 | tools.kubectl.image.tag | string | `""` | Overrides the image tag |
 | tools.kubectl.image.version | string | `""` | This value is deprecated in favor of tag. (Will be removed in a future release) |
 | topologySpreadConstraints | list | `[]` |  |
-| tornjak.config.connectionType | string | `"http"` | Tornjak supports 3 connection types: `http`, `tls`, and `mtls`. Select only one. |
 | tornjak.config.dataStore | object | `{"driver":"sqlite3","file":"/run/spire/data/tornjak.sqlite3"}` | Persistent DB for storing Tornjak specific information |
 | tornjak.config.service | object | `{"annotations":{},"portHttp":10080,"portHttps":10443,"type":"ClusterIP"}` | Enables the service for a given `connectionType` |
 | tornjak.config.tlsSecret | string | `"tornjak-tls-secret"` | Name of the secret containing server side key and certificate for TLS verification (required for `tls` or `mtls` connectionType) |
--- a/charts/spire/charts/spire-server/README.md
+++ b/charts/spire/charts/spire-server/README.md
@@ -37,13 +37,9 @@ Tornjak is the UI and Control Plane for SPIRE [https://github.com/spiffe/tornjak
 
 When Tornjak is enabled, it is exposed on both http and https (if TLS server certs are configured). Tornjak handles a permanent redirect from `http` to `https` to ensure users always use the https endpoint.
 
-In addition to that you can configure a `client certificate authority`, this will make Tornjak backend verify Client certificates signed by this authority to enable mTLS authentication.
+In addition, you can configure a `client certificate authority`, this will make Tornjak backend verify Client certificates signed by this authority to enable mTLS authentication.
 
-> **Warning**: For production ensure to configure the certificates including CA, to protect Tornjak from unauthorized access.
-
-Tornjak automatically determines the connection type based on provided information. See below.
-
-**Warning**: For production, we recommend configuring tls certificates and client CA to protect Tornjak from unauthorized access.
+**Warning**: For production, we recommend configuring TLS certificates and client CA to protect Tornjak from unauthorized access.
 
 ### Tornjak with TLS Connection Type
 
--- a/charts/spire/charts/spire-server/README.md
+++ b/charts/spire/charts/spire-server/README.md
@@ -147,9 +147,10 @@ A Helm chart to install the SPIRE server.
 | topologySpreadConstraints | list | `[]` |  |
 | tornjak.config.connectionType | string | `"http"` | Tornjak supports 3 connection types: `http`, `tls`, and `mtls`. Select only one. |
 | tornjak.config.dataStore | object | `{"driver":"sqlite3","file":"/run/spire/data/tornjak.sqlite3"}` | Persistent DB for storing Tornjak specific information |
-| tornjak.config.serverSecret | string | `"tornjak-server-secret"` | Name of the secret containing server side key and certificate for TLS verification (required for `tls` or `mtls` connectionType) |
 | tornjak.config.service | object | `{"annotations":{},"port":10000,"type":"ClusterIP"}` | Enables the service for a given `connectionType` |
-| tornjak.config.userSecret | string | `"tornjak-user-secret"` | Name of the secret containing user CA for mTLS verification (required for `mtls` connectionType) |
+| tornjak.config.tlsSecret | string | `"tornjak-tls-secret"` | Name of the secret containing server side key and certificate for TLS verification (required for `tls` or `mtls` connectionType) |
+| tornjak.config.userCA.name | string | `"tornjak-user-ca"` |  |
+| tornjak.config.userCA.type | string | `"Secret"` | Type of delivery for the user CA for mTLS client verification options are `Secret` or `ConfigMap` (required for `mtls` connectionType) |
 | tornjak.enabled | bool | `false` | Deploys Tornjak API (backend) (Not for production) |
 | tornjak.image.pullPolicy | string | `"IfNotPresent"` | The Tornjak image pull policy |
 | tornjak.image.registry | string | `"ghcr.io"` | The OCI registry to pull the Tornjak image from |
--- a/charts/spire/charts/spire-server/README.md
+++ b/charts/spire/charts/spire-server/README.md
@@ -30,7 +30,10 @@ A Helm chart to install the SPIRE server.
 
 ## Tornjak
 
-Tornjak is the UI and Control Plane for SPIRE [https://github.com/spiffe/tornjak](https://github.com/spiffe/tornjak)
+Tornjak is the UI and Control Plane for SPIRE [https://github.com/spiffe/tornjak](https://github.com/spiffe/tornjak) and it is composed of two components:
+
+* Backend (this chart) - Tornjak APIs that extend SPIRE APIs with Control Plane functionality
+* [Frontend](../tornjak-frontend/README.md) - Tornjak UI
 
 When Tornjak is enabled, it can be reached in one of the three connection types:
 
@@ -53,7 +56,7 @@ kubectl -n spire-server create secret tls tornjak-tls-secret --cert=client.crt -
 Once the charts are deployed, you can test the TLS connection with the following command (assuming localhost):
 
 ```console
-curl --cacert CA/rootCA.crt https://localhost:10000
+curl --cacert CA/rootCA.crt https://localhost:10443
 ```
 
 ### Tornjak with mTLS
@@ -75,7 +78,7 @@ kubectl -n spire-server create secret generic tornjak-user-ca --from-file=ca.crt
 Once the charts are deployed, you can test the mTLS connection with the following command (assuming localhost):
 
 ```console
-curl  --cacert CA/rootCA.crt --key client.key --cert client.crt https://localhost:10000
+curl  --cacert CA/rootCA.crt --key client.key --cert client.crt https://localhost:10443
 ```
 
 ## Values
--- a/charts/spire/charts/spire-server/README.md
+++ b/charts/spire/charts/spire-server/README.md
@@ -43,6 +43,8 @@ When Tornjak is enabled, it can be reached in one of the three connection types:
 
 Tornjak automatically determines the connection type based on provided information. See below.
 
+**Warning**: For production, we recommend configuring tls certificates and client CA to protect Tornjak from unauthorized access.
+
 ### Tornjak with TLS Connection Type
 
 TLS connection requires Tornjak to have access to TLS key and certificate.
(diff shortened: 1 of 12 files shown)


charts/spire/charts/spire-server/values.yaml
```diff
@@ -351,11 +347,59 @@ tornjak:
...
+        annotations: {}
+
+    # Tornjak TLS service
+    tls:
+      # -- Enables Tornjak TLS service
+      enabled: false
+      # -- Container port value for TLS
+      port: 20000
+      # -- Certificate for TLS connection
+      cert: "/opt/spire/sample-keys/cert.pem"
```
kfox1111, line 376: Does cert and key filename inside the container need to be configurable? k8s has standardized secrets of type tls.

charts/spire/charts/spire-server/values.yaml
```diff
@@ -351,11 +347,59 @@ tornjak:
...
+        annotations: {}
+
+    # Tornjak mTLS service
+    mtls:
+      # -- Enables Tornjak TLS service
+      enabled: false
+      # -- Container port value for mTLS
+      port: 30000
+      # -- Certificate for mTLS connection
+      cert: "/opt/spire/sample-keys/cert.pem"
```
kfox1111, line 392: Same

charts/spire/charts/spire-server/values.yaml
```diff
@@ -351,11 +347,59 @@ tornjak:
...
+      # -- Key for mTLS connection
+      key:  "/opt/spire/sample-keys/key.pem"
+      # -- CA for mTLS connection
+      ca: "/opt/spire/sample-keys/rootCA.pem"
```
kfox1111, line 396: I'm assuming this is for validating the other side's certs? maybe userCA? Should consider just making this a name within a fixed location in the pod so the user doesn't have to know stuff such as /opt/spire/sample-keys?

charts/spire/charts/spire-server/templates/tornjak-config.yaml
```diff
@@ -8,11 +8,29 @@ data:
...
-        port = "10000" # if HTTP enabled, opens HTTP listen port at container port 10000
+        port = "{{ .Values.tornjak.config.http.port }}" # if HTTP enabled, opens HTTP listen port at container port 10000
+      }
+      {{- end }}
+      {{- if eq (.Values.tornjak.config.tls.enabled | toString) "true" }}
+      tls {
+        enabled = true
+        port = "{{ .Values.tornjak.config.tls.port }}" # container port for TLS connection
+        cert = "{{ .Values.tornjak.config.tls.cert }}" # TLS cert
+        key = "{{ .Values.tornjak.config.tls.key }}"  # TLS key
```
kfox1111, line 22: Maybe hardcode the locations here?

charts/spire/charts/spire-server/templates/statefulset.yaml
```diff
@@ -189,6 +203,10 @@ spec:
             - name: spire-data
               mountPath: /run/spire/data
               readOnly: false
+            {{- if or (eq (.Values.tornjak.config.tls.enabled | toString) "true") (eq (.Values.tornjak.config.mtls.enabled | toString) "true") }}
+            - name: certs
+              mountPath: /opt/spire/sample-keys
```
kfox1111, line 208: some name other then sample-keys perhaps?

charts/spire/charts/spire-server/templates/statefulset.yaml
```diff
@@ -218,6 +236,12 @@ spec:
           emptyDir: {}
         - name: spire-controller-manager-tmp
           emptyDir: {}
+        {{- if or (eq (.Values.tornjak.config.tls.enabled | toString) "true") (eq (.Values.tornjak.config.mtls.enabled | toString) "true") }}
+        - name: certs
+          secret:
+            defaultMode: 256
+            secretName: tornjak-certs
```
kfox1111, line 243: Maybe make secretName configurable.

charts/spire/charts/spire-server/values.yaml
kfox1111: It should probably be in its own configmap/secret. Different systems do that part differently.

charts/spire/charts/spire-server/README.md
```diff
@@ -129,17 +129,28 @@ A Helm chart to install the SPIRE server.
...
+| tornjak.config.mtls.cert | string | `"/opt/spire/sample-keys/cert.pem"` | Certificate for mTLS connection |
+| tornjak.config.mtls.enabled | bool | `false` | Enables Tornjak TLS service |
+| tornjak.config.mtls.key | string | `"/opt/spire/sample-keys/key.pem"` | Key for mTLS connection |
+| tornjak.config.mtls.port | int | `30000` | Container port value for mTLS |
+| tornjak.config.mtls.service | object | `{"annotations":{},"port":30000,"type":"ClusterIP"}` | Service to handle Tornjak mTLS connection |
+| tornjak.config.tls.cert | string | `"/opt/spire/sample-keys/cert.pem"` | Certificate for TLS connection |
+| tornjak.config.tls.enabled | bool | `false` | Enables Tornjak TLS service |
+| tornjak.config.tls.key | string | `"/opt/spire/sample-keys/key.pem"` | Key for TLS connection |
+| tornjak.config.tls.port | int | `20000` | Container port value for TLS |
+| tornjak.config.tls.service | object | `{"annotations":{},"port":20000,"type":"ClusterIP"}` | Service to handle Tornjak TLS connection |
```
kfox1111, line 146: are http/mtls/tls exclusive or can you really enable more then one?

charts/spire/charts/spire-server/values.yaml
mrsabath: I was initially thinking we could configure multiple keys e.g. different keys for TLS and mTLS but that's probably overkill here. I am fine with hardcoding the path to the key and cert

charts/spire/charts/spire-server/README.md
mrsabath: You have to enable AT LEAST one, otherwise there is no point of running it, since you can't communicate, therefore if none is configured, Tornjak startup configuration validation throws an error

charts/spire/charts/spire-server/values.yaml
mrsabath: This is used to build the ConfigMap `tornjak-config`, but you are right, we can hard code it there

charts/spire/charts/spire-server/README.md
kfox1111: Is there reason to have more then one though? Just wondering if this should be a mode enum rather then a set of bools.

kfox1111: maybe an alternate angle... The mtls part, is that for user auth?

mrsabath: As is, it can be used for user authentication, however I am preparing a next PR to bring a proper user authentication and authorization via Keycloak.

mrsabath: BTW I liked your suggestion of separating the userCA into a second secret. It's done.

kfox1111 (commented):
I know its not done yet. But some ideas for value naming.

charts/spire/charts/spire-server/values.yaml
```diff
@@ -354,11 +350,30 @@ tornjak:
...
+    # -- Tornjak supports 3 connection types: `http`, `tls`, and `mtls`.
+    # Select only one.
+    connectionType: http
+
+    # -- Name of the secret containing server side key and certificate for TLS verification
+    # (required for `tls` or `mtls` connectionType)
+    serverSecret: tornjak-server-secret
+    # -- Name of the secret containing user CA for mTLS verification
+    # (required for `mtls` connectionType)
+    userSecret: tornjak-user-secret
```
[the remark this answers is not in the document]
kfox1111: Actually, there is a third type in development too that we may want to someday support that makes a lot of sense: https://github.com/kubernetes/enhancements/issues/3257

charts/spire/charts/spire-server/README.md
```diff
@@ -28,6 +28,56 @@ A Helm chart to install the SPIRE server.
 
 * <https://github.com/spiffe/helm-charts/tree/main/charts/spire>
 
+## Tornjak
+
+Tornjak is the UI and Control Plane for SPIRE [https://github.com/spiffe/tornjak](https://github.com/spiffe/tornjak)
```
marcofranssen, line 33: Should we mention the UI part in this README? 

Technically it is part of another chart. Probably better to document this in the spire chart or the tornjak-ui chart. And refer int this documentation to that chart.

charts/spire/charts/spire-server/README.md
```diff
@@ -28,6 +28,56 @@ A Helm chart to install the SPIRE server.
...
+* tls - enables Transport Layer Security (TLS) with HTTPS communication. Tornjak client can validate Tornjak server using TLS Certificate (see instructions below)
+* mtls - mutual TLS. Both, Tornjak client and the server can validate each other using provided certificates (see instructions below)
+
+### Tornjak with TLS
+
+`tornjak.config.connectionType=tls`
+
+TLS connection requires Tornjak to have access to TLS key and certificate.
+Complete instruction on creating your own TLS certificate can be found [here](https://github.com/spiffe/tornjak/blob/main/examples/tls_mtls/README.md).
+TLS Certificate and the private key must be provided to Tornjak via *TLS Secret*. Prior to deploying this Helm chart, create TLS Secret in the deployment namespace (e.g. `spire-server`)
```
marcofranssen, line 47: Shouldn't we just bake this in, similar like the service accounts are implemented?

charts/spire/charts/spire-server/templates/_helpers.tpl
```diff
@@ -178,3 +178,46 @@ Tornjak specific section
...
 {{- end }}
+
+{{/*
+TLS and mTLS Connection Types require tlsSecret to be set before the deployment
+mTLS Connection Types requires userCA.name to be set before the deployment
+*/}}
+{{- define "spire-tornjak.servicename" -}}
+
+{{- if eq .Values.tornjak.config.connectionType "http" -}}
+{{- include "spire-tornjak.backend" . -}}-http
```
marcofranssen, line 189: Do we really need to change the service name when changing the deployment type?

Preferably I would keep the service name the same no matter what deployment type you choose.

In the end this ends up as a dns name.

charts/spire/charts/spire-server/templates/statefulset.yaml
```diff
@@ -160,19 +160,21 @@ spec:
...
             {{- toYaml .Values.tornjak.startupProbe | nindent 12 }}
+          {{- end }}
           args:
             - --spire-config
             - /run/spire/config/server.conf
             - --tornjak-config
             - /run/spire/tornjak-config/server.conf
           ports:
-            - name: tornjak
+            - name: {{ include "spire-tornjak.portname" . }}
```
marcofranssen, line 177: Do we need to change the port name? It will complicate the ingresses as well. 

Would be better if we can just point the ingresses at the tornjak service on the http port.

More logical would be 2 ports on the service:

- http
- https

Tornjak itself should listen on both and do a perm redirect as soon you enable TLS.

That way you don't break existing web behavior, where end users are automatically redirected to the https port as soon it is enabled in tornjak.

charts/spire/charts/spire-server/templates/statefulset.yaml
```diff
@@ -189,6 +191,14 @@ spec:
             - name: spire-data
               mountPath: /run/spire/data
               readOnly: false
+            {{- if or (eq (.Values.tornjak.config.connectionType | toString) "tls") (eq (.Values.tornjak.config.connectionType | toString) "mtls") }}
```
marcofranssen, line 194: I don't think the toString parts are required here.

charts/spire/charts/spire-server/templates/statefulset.yaml
```diff
@@ -189,6 +191,14 @@ spec:
...
+            - name: server-cert
+              mountPath: /opt/spire/server
+            {{- end }}
+            {{- if eq (.Values.tornjak.config.connectionType | toString) "mtls" }}
```
marcofranssen, line 198: I don't think the toString parts are required here

charts/spire/charts/spire-server/templates/statefulset.yaml
```diff
@@ -218,6 +228,24 @@ spec:
           emptyDir: {}
         - name: spire-controller-manager-tmp
           emptyDir: {}
+        {{- if or (eq (.Values.tornjak.config.connectionType | toString) "tls") (eq (.Values.tornjak.config.connectionType | toString) "mtls") }}
```
marcofranssen, line 231: Same as above

charts/spire/charts/spire-server/templates/statefulset.yaml
```diff
@@ -189,6 +191,14 @@ spec:
             - name: spire-data
               mountPath: /run/spire/data
               readOnly: false
+            {{- if or (eq (.Values.tornjak.config.connectionType | toString) "tls") (eq (.Values.tornjak.config.connectionType | toString) "mtls") }}
+            - name: server-cert
+              mountPath: /opt/spire/server
```
marcofranssen, line 196: should we store all certs in the same folder?

That is how it normally works on webservers.

- ca.pem
- server.pem

Also wonder if there should be a user cert here.

Normally in a PKI setup your server and client cert would be signed with the same CA.

The client cert is normally just on the client if I'm not mistaking.

charts/spire/charts/spire-server/templates/tornjak-config.yaml
```diff
@@ -8,11 +8,29 @@ data:
...
+      {{- end }}
+      {{- if eq (.Values.tornjak.config.connectionType | toString) "mtls" }}
+      mtls {
+        enabled = true
+        port = "10000" # container port for mTLS connection
+        cert = "/opt/spire/server/tls.crt" # mTLS server cert
+        key  = "/opt/spire/server/tls.key"  # mTLS server key
+        ca   = "/opt/spire/user/ca.crt"  # mTLS user CA
+      }
+      {{- end }}
```
marcofranssen, line 33: tls and mtls should use the same port.

mtls is only related to the client.

marcofranssen (changes requested):
I think we should implement some plumbing into the chart.

E.g. allow to create the secret via the chart. Ensure there is just a single secret containing all the required certs.

That simplifies the volume mounts.

Furthermore I gave some inline feedback on the service and port. That can be much simpler by doing following

Service structure:

tornjak
  - http
    - 10080
  - https.  only if tls/mtls enabled
    - 10443

secret for certs

tornjak-tls-secret:
  ca.crt
  server.crt
  client.crt

Happy to pick this up in pairing session next week to further explain.

Also wondering if we should just leverage the certManager integration as contributed in the other PR.

We can sign these certs with the same Cert-manager Issuer and CA.

charts/spire/charts/spire-server/templates/_helpers.tpl
mrsabath: You are right. I kept the various names because earlier we had services and ports for each connection, all active at the same, but since we enforce only one connection, there is no need to use different names.

kfox1111:
@marcofranssen, I asked him to do it as two secrets before.

There are multiple reasons not exhaustive:
1. tls certificates in k8s (secret type=tls) has a particular format that includes private key and cert. Other systems such as cert-manager can automatically create certs in this form. So its important that we support it.
2. the user ca is needed for mtls. Its what the client certificate is verified against. if you use cert-manager to setup your user ca as well, its likely a completely separate ca then your service ca. This should come in through an alternate configmap/secret/the new ClusterTrustBundles type (https://github.com/kubernetes/enhancements/issues/3257)

Having a create option though where the user can specify the secrets in values would be ok. That could be in this pr or in another followon feature pr?

marcofranssen:
> @marcofranssen, I asked him to do it as two secrets before.
> 
> There are multiple reasons not exhaustive:
> 
> 1. tls certificates in k8s (secret type=tls) has a particular format that includes private key and cert. Other systems such as cert-manager can automatically create certs in this form. So its important that we support it.

Oh right! Very good point. Then others are just not using this type of secret. I like to use the k8s native types.

> 2. the user ca is needed for mtls. Its what the client certificate is verified against. if you use cert-manager to setup your user ca as well, its likely a completely separate ca then your service ca. This should come in through an alternate configmap/secret/the new ClusterTrustBundles type ([ClusterTrustBundles (previously Trust Anchor Sets) kubernetes/enhancements#3257](https://github.com/kubernetes/enhancements/issues/3257))

Oh then it is probably the naming that confused me. However why would the CA be different?

I would do it like this

- Cluster CA
  - server cert
  - intermediate CA
    - client cert
    - client cert
    - client cert

Isn't the root CA sufficient to build the trustchain? Now you made me doubt though :) 

> Having a create option though where the user can specify the secrets in values would be ok. That could be in this pr or in another followon feature pr?

kfox1111:
User CA's sometimes need to be on the internet so that users can properly get certificates for themselves. Host CA's dont need to be accessible on the internet. So keeping them separate has security benefits.You don't necessarily want a User CA to be able to issue valid Host certs from the User CA as then there is more attack surface.

There are other reasons why you may want separate chains of trust there. Hosts often need to be trusted by browsers or other services not under the sysadmin's control, so having a more traditional chain of trust makes hosts validation easier. For example, with letsencrypt. While user ca validation only happens on your own hosts, so easier to distribute to the set of machines needing that CA for client validation.

So its best to allow treating host chain of trust completely separate from user chain of trust. Some users may combine the two if they like but that should be up to them and their situation.

kfox1111:
> > User CA's sometimes need to be on the internet so that users can properly get certificates for themselves. Host CA's dont need to be accessible on the internet. So keeping them separate has security benefits.You don't necessarily want a User CA to be able to issue valid Host certs from the User CA as then there is more attack surface.
> > There are other reasons why you may want separate chains of trust there. Hosts often need to be trusted by browsers or other services not under the sysadmin's control, so having a more traditional chain of trust makes hosts validation easier. For example, with letsencrypt. While user ca validation only happens on your own hosts, so easier to distribute to the set of machines needing that CA for client validation.
> > So its best to allow treating host chain of trust completely separate from user chain of trust. Some users may combine the two if they like but that should be up to them and their situation.
> 
> So isn't that intermediate as I shown above exactly what you need to just make that available on the internet, but still have it trusted by the upstream cluster CA?

No. If you have your own private CA for the user ca, and a public chain of trust such as letsencrypt for the hosts, it is not possible for the intermediates to be rooted at just one CA. Nor is it desirable as I explained, for security reasons. You don't want machines to trust host certs signed by the user ca so yiou can't trust the a root where there is a shared root between a user and host ca.

Also, an intermediate CA is just a way to add an additional layer of security so you don't have to expose the root ca. The intermediate CA's can have expiry times much less then the root ca's expiration date. But can issue certs much the same as the root.

The user ca **must** be configurable separately then any ca used for the hosts.

kfox1111:
There are a few different options we may want to consider for the tls certs and for the user ca.....

For the host cert for each of the frontend and backend, we could support an enum with:
existing, cert-manager, values

where existing would just use the secret that exists, cert-manager would create a Certificate object to generate the tls secret, and values could create the secret from a key/cert thats specified in values

For now, this pr only supports an existing secret. Or the user can use cert-manager to create it manually. This is probably a good first step and the other options could be added in a follow up pr.

For the userCA, we have some similar options.... some combination of
existing-secret, existing-configmap, cert-manager, values

where existing-secret/configmap is preexisting objects, cert-manager would create a self signed user ca and then reference it, and values would have the ca.pem specified directly in the values and would get put in a secret.

Similar to the previous, I'd recommend the additional options be added in a follow up pr.

charts/spire/charts/spire-server/templates/statefulset.yaml
kfox1111: The client cert is just on the clients, yes. The user ca is coming from a separate object, so better to have a separate mount IMO.

kfox1111 (approved):
Some minor text things that could be changed, but could do it in a followup pr too if everyone else is happy with it the way it is.

mrsabath:
https://github.com/spiffe/helm-charts/pull/338#discussion_r1234512531
Thank you for the comment @kfox1111 
I think the most recent version already has two mount-points: https://github.com/spiffe/helm-charts/blob/tornjak-tls/charts/spire/charts/spire-server/templates/statefulset.yaml#L213-L218

charts/spire/charts/spire-server/templates/_helpers.tpl
```diff
@@ -168,3 +168,54 @@ Tornjak specific section
...
+tornjak-srv-https
+{{- end -}}
+
+{{- define "spire-tornjak.target-portname-http" -}}
+tornjak-http
+{{- end -}}
+
+{{- define "spire-tornjak.target-portname-https" -}}
+tornjak-https
+{{- end -}}
```
marcofranssen, line 221: I would drop these to make it more clear in templates.

charts/spire/charts/spire-server/templates/statefulset.yaml
```diff
@@ -175,20 +175,25 @@ spec:
...
+          {{- end }}
           args:
             - --spire-config
             - /run/spire/config/server.conf
             - --tornjak-config
             - /run/spire/tornjak-config/server.conf
           ports:
-            - name: tornjak
-              containerPort: 10000
+            - name: {{ include "spire-tornjak.target-portname-http" . }}
```
marcofranssen, line 192:
```suggestion
            - name: http
```

charts/spire/charts/spire-server/templates/statefulset.yaml
```diff
@@ -175,20 +175,25 @@ spec:
...
+              containerPort: 10080
+              protocol: TCP
+            - name: {{ include "spire-tornjak.target-portname-https" . }}
```
marcofranssen, line 195:
```suggestion
            - name: https
```

charts/spire/charts/spire-server/templates/tornjak-service.yaml
```diff
@@ -3,20 +3,24 @@ apiVersion: v1
...
    {{- include "spire-server.selectorLabels" . | nindent 4 }}
   ports:
-    - name: {{ include "spire-tornjak.backend" . }}
-      port: {{ .Values.tornjak.service.port }}
-      targetPort: tornjak
+    - name: {{ include "spire-tornjak.service-portname-http" . }}
+      port: {{ .Values.tornjak.config.service.portHttp }}
+      targetPort: {{ include "spire-tornjak.target-portname-http" . }}
+      protocol: TCP
+    - name: {{ include "spire-tornjak.service-portname-https" . }}
```
marcofranssen, line 22:
```suggestion
    - name: https
```

charts/spire/charts/spire-server/values.yaml
```diff
@@ -369,11 +365,34 @@ tornjak:
...
   config:
-    # -- persistent DB for storing Tornjak specific information
+    # -- Persistent DB for storing Tornjak specific information
     dataStore:
       driver: "sqlite3"
       file: "/run/spire/data/tornjak.sqlite3"
+
+    # -- Tornjak supports 3 connection types: `http`, `tls`, and `mtls`.
+    # Select only one.
+    connectionType: http
```
marcofranssen, line 377: I think we can drop this. Could be determined based on if there are tls certs or not.

mrsabath:
@marcofranssen per your suggestion, Tornjak Helm charts currently determine automatically the Tornjak connection type, based on provided information (secrets or ConfigMaps). We will update the Tornjak code (https://github.com/spiffe/tornjak/issues/268) to do this determination on the level of the Tornjak image. Once completed there, we will update these helm charts here with a new configuration format, as a follow-up PR.

charts/spire/charts/spire-server/templates/tornjak-service.yaml
```diff
@@ -3,20 +3,24 @@ apiVersion: v1
...
+  name: {{ include "spire-tornjak.servicename" . }}
+  {{- with .Values.tornjak.config.service.annotations }}
   annotations:
     {{- toYaml . | nindent 8 }}
   {{- end }}
   labels:
     {{- include "spire-server.labels" . | nindent 4 }}
 spec:
-  type: {{ .Values.tornjak.service.type }}
+  type: {{ .Values.tornjak.config.service.type }}
```
marcofranssen, line 14: This nesting into config deviates from the location where all other charts define the k8s service object. Would prefer to keep that where it was.

charts/spire/charts/spire-server/values.yaml
```diff
@@ -372,11 +368,35 @@ tornjak:
...
+      # (required for `mtls` connectionType)
+      type: Secret
+      name: tornjak-user-ca
+
+    # -- Enables the service for a given `connectionType`
+    service:
+      type: ClusterIP
+      portHttp: 10080
+      portHttps: 10443
+      annotations: {}
```
marcofranssen, line 398: The comment is a bit misleading as this is the k8s service. For other native k8s service object values we don't have documentation explaining the service either.

Reducing the indentation level as suggested above, to keep it similar like the remainder of charts.

```suggestion
  service:
    type: ClusterIP
    portHttp: 10080
    portHttps: 10443
    annotations: {}
```

charts/spire/charts/spire-server/README.md
```diff
@@ -28,6 +28,61 @@ A Helm chart to install the SPIRE server.
...
+Tornjak is the UI and Control Plane for SPIRE [https://github.com/spiffe/tornjak](https://github.com/spiffe/tornjak) and it is composed of two components:
+
+* Backend (this chart) - Tornjak APIs that extend SPIRE APIs with Control Plane functionality
+* [Frontend](../tornjak-frontend/README.md) - Tornjak UI
+
+When Tornjak is enabled, it can be reached in one of the three connection types:
+
+* HTTP - simple HTTP connection recommended for demonstration purpose only. It should not be use for production.
+* TLS - enables Transport Layer Security (TLS) with HTTPS communication. Tornjak client can validate Tornjak server using TLS Certificate (see instructions below)
+* mTLS - mutual TLS. Both, Tornjak client and the server can validate each other using provided certificates (see instructions below)
```
marcofranssen, line 42: Here a suggestion to update the docs in such way we should also implement this feature to improve the best UX (User Experience) for both the chart installation as well usage of Tornjak itself.

```suggestion
When Tornjak is enabled, it is exposed on both http and https (if TLS server certs are configured). Tornjak handles a permanent redirect from `http` to `https` to ensure users always use the https endpoint.

In addition to that you can configure a `client certificate authority`, this will make Tornjak backend verify Client certificates signed by this authority to enable mTLS authentication.

> **Warning**: For production ensure to configure the certificates including CA, to protect Tornjak from unauthorized access.
```

charts/spire/charts/spire-server/templates/tornjak-config.yaml
```diff
@@ -8,11 +8,29 @@ data:
...
+        key  = "/opt/spire/server/tls.key"  # TLS server key
+      }
+      {{- end }}
+      {{- if eq (include "spire-tornjak.connectionType" .) "mtls" }}
+      mtls {
+        enabled = true
+        port = "10443" # container port for mTLS connection
+        cert = "/opt/spire/server/tls.crt" # mTLS server cert
+        key  = "/opt/spire/server/tls.key"  # mTLS server key
+        ca   = "/opt/spire/user/ca.crt"  # mTLS user CA
```
marcofranssen, line 31: I would recommend simplifying this to 

```
http {
  port = 10080
}
https {
  port     = 10443
  cert     = …
  key      = …
  clientCA = … (optional setting)
}
```

Implementation wise you will do a permanent redirect if https section is in the config from `http` to `https`.

https section requires mandatory the cert and key.

if https section also has the clientCA, then you will verify that a client Cert is provided.

This way both install and user time the entire connection type is unobtrusive to the end user.

Basically only thing they need to know is the endpoint and all the rest is transparantly resolved.

charts/spire/charts/spire-server/README.md
kfox1111: tls is highly recommended. mtls is more of an authentication feature. When oidc is implemented, that may be preferred to be used over mtls in a lot of cases, so we shouldn't necessarily strongly recommend the user ca part.

marcofranssen: Agree, but for now it is the only option and therefore the current recommendation until we have OIDC available.

kfox1111: Yeah. I guess we can amend it once that's implemented to recommend specifying client CA, or OIDC at that point.

charts/spire/charts/spire-server/values.yaml
```diff
@@ -344,7 +344,7 @@ nodeAttestor:
     enabled: true
     serviceAccountAllowList: []
 
-# tornjak - Tornjak specific configuration
+# tornjak - Tornjak default values
```
marcofranssen, line 347: All values in this file are the defaults, I think it is not worthwhile to mention the values file are defaults.

charts/spire/charts/spire-server/README.md
```diff
@@ -35,11 +35,11 @@ Tornjak is the UI and Control Plane for SPIRE [https://github.com/spiffe/tornjak
...
 
-When Tornjak is enabled, it can be reached in one of the three connection types:
+When Tornjak is enabled, it is exposed on both http and https (if TLS server certs are configured). Tornjak handles a permanent redirect from `http` to `https` to ensure users always use the https endpoint.
 
-* HTTP - simple HTTP connection recommended for demonstration purpose only. It should not be use for production.
-* TLS - enables Transport Layer Security (TLS) with HTTPS communication. Tornjak client can validate Tornjak server using TLS Certificate (see instructions below)
-* mTLS - mutual TLS. Both, Tornjak client and the server can validate each other using provided certificates (see instructions below)
+In addition to that you can configure a `client certificate authority`, this will make Tornjak backend verify Client certificates signed by this authority to enable mTLS authentication.
+
+> **Warning**: For production ensure to configure the certificates including CA, to protect Tornjak from unauthorized access.
```
kfox1111:
```suggestion
> **Warning**: For production, we recommend configuring tls certificates and client CA to protect Tornjak from unauthorized access.
```

kfox1111 (approved):
LGTM

edwbuck:
@marcofranssen You blocked this review with a request for changes, but subsequent changes have moved your requests into the place that deleted files go.  Would you re-review this commit and if you decide it is sufficient, approve it?

charts/spire/charts/spire-server/templates/tests/test-tornjak-connection.yaml
```diff
@@ -1,4 +1,5 @@
 {{- if eq (.Values.tornjak.enabled | toString) "true" }}
+{{- if eq .Values.tornjak.config.connectionType "http" }}
```
marcofranssen, line 2:
```suggestion
{{- if eq (include "spire-tornjak.connectionType" .) "http" }}
```

marcofranssen:
> @marcofranssen Can we come to an agreement on the secrets issue? Basically I can see some scenarios where the secrets are not to be generated through the Helm interface, such as when they are provided by Vault and accessed through a secrets claim that has vault populating the secret.
> 
> I see this as a very good addition that is nearly ready to ship, if we can come to a resolution on the secrets item soon.

This PR isn't blocked on a secrets discussion neither are we ruling out any options to manage secrets. It is blocked because of implementation bugs. Working with Marius to get those resolved, no need to chase us.

charts/spire/charts/spire-server/templates/tornjak-service.yaml
```diff
@@ -15,8 +15,12 @@ spec:
   selector:
    {{- include "spire-server.selectorLabels" . | nindent 4 }}
   ports:
-    - name: {{ include "spire-tornjak.backend" . }}
-      port: {{ .Values.tornjak.service.port }}
-      targetPort: tornjak
+    - name: tornjak-srv-http
+      port: {{ .Values.tornjak.service.portHttp }}
+      targetPort: tornjak-http
+      protocol: TCP
```
kfox1111, line 21: If https is enabled, do we want http exposed out the service if the user specifies type=LoadBalancer? Will it just redirect to https, or will it allow bypassing mtls restrictions?

mrsabath: The idea was to make a permanent redirect from HTTP to HTTPS port, so the HTTP service will be always open but redirecting to HTTPS

marcofranssen: https://github.com/spiffe/tornjak/pull/282/files#diff-7c41928cb703314c59d418bbec523718010421f052461c7e02e88fd3b689433cR612

This is how it is mitigated. This is also a UX feature, so if the consumer forgets to type https in front of the url, this permanent redirect will do it on behalf of the client.

mrsabath:
Minor naming convention updates to be handled by follow-up PR (issue #389 )

marcofranssen (approved):
LGTM

PR #344, opened 2023-06-13
Dropping k8s versions in CI older than 3, as per readme

drewwells:
PRs run 86 tests, this should significantly cut down on that.
correction 125 tests

image


marcofranssen (approved):
LGTM

faisal-memon (approved):

PR #346, opened 2023-06-14
Move spire-lib macro's to their own library chart

kfox1111:


.github/workflows/helm-chart-ci.yaml
```diff
@@ -96,7 +96,7 @@ jobs:
       - name: Run chart-testing (lint)
         run: |
           ct lint --debug ${{ github.base_ref != 'release' && '--check-version-increment=false' || '' }} \
-            --target-branch ${{ github.base_ref }}
+            --target-branch ${{ github.base_ref }} --excluded-charts spire-lib
```
kfox1111, line 99: Lint doesn't support library charts at the moment.

marcofranssen: No need to exclude as ct lint works on the top level charts.

kfox1111: Its something we will need to add once we break out the chart, is related to adding the new chart, and doesn't hurt anything now. I can remove it, but should we just get it out of the way now?

Last I tried, I hit https://github.com/helm/chart-testing/issues/237

kfox1111: It may be the charts/common/values.yaml blank file that is allowing it to work. If I tried that, would you prefer having a blank values.yaml over excluding it?

marcofranssen: Yes that would be better I think

mrsabath (approved):
LGTM

charts/spire/Chart.yaml
```diff
@@ -22,6 +22,9 @@ maintainers:
     email: edwbuck@gmail.com
 kubeVersion: ">=1.21.0-0"
 dependencies:
+  - name: spire-lib
+    repository: file://./charts/spire-lib
+    version: 0.1.0
```
drewwells, line 27: This is not a standard pattern. I suggest not adding another subchart and putting this in the parent chart. https://helm.sh/docs/chart_best_practices/templates/#names-of-defined-templates

This set of subchart would be improved by collapsing into one chart.

marcofranssen: > This is not a standard pattern. I suggest putting this in the parent chart. [helm.sh/docs/chart_best_practices/templates/#names-of-defined-templates](https://helm.sh/docs/chart_best_practices/templates/#names-of-defined-templates)
> 
> This set of subchart would be improved by collapsing into one chart.

Agree, we should hold this off till we break out some charts like spire-server etc. To not waste time keeping this in sync probably better to recreate this PR later. Instead of nesting this chart it can then be a top level chart with its own release cycle.

Before we can breakout charts we first need to fix #324 so our CI can support testing multiple charts.

@drewwells what we are aiming for is something like this. https://github.com/sigstore/helm-charts this will allow us to do advanced usecases like nested spire and such more flexible.

This spire-lib chart would be similar to the sigstore-common chart to share some commonly.

kfox1111: This was an attempt to get the library chart done so the pr that breaks out the subcharts to /charts can have as minimal changes as possible. adding all this code into that breakout pr just makes for a lot more possible merge issues. Its not ideal to have under charts/spire/charts, but is a piece to getting it broken out.

marcofranssen: The breakout can be done chart by chart. Where the first breakout can be these lib functions. Next can be spiffe-csi, because that has very few dependencies, then probably spiffe-oidc.

kfox1111: When I tried it last, it gets confused doing a helm dep up when you have subcharts directly in the charts/spire/charts/ dir. It tries to tar up each of the dependencies and then you get duplicates of the subcharts. I don't think having charts directly under is actually officially supported by helm. So, I think all the charts need to be broken out at once.

PR #347, opened 2023-06-14
Add missing metadata to subcharts

kfox1111:


charts/spire/charts/spiffe-csi-driver/Chart.yaml
```diff
@@ -4,3 +4,18 @@ description: A Helm chart to install the SPIFFE CSI driver.
 type: application
 version: 0.1.0
 appVersion: "0.2.3"
+keywords: ["spiffe", "csi-driver"]
+home: https://github.com/spiffe/helm-charts/tree/main/charts/spire
```
faisal-memon, line 8: Will these URLS need to change once we get everything re organized?

marcofranssen (approved):
LGTM

charts/spire/charts/spiffe-csi-driver/Chart.yaml
kfox1111: hmm... true. should I just make them point at the root of the repo, or leave them as is and we can change it when we break them out?

faisal-memon: Leave as is. I was just confirming.

faisal-memon (approved):

PR #272, opened 2023-05-10
Allow for ConfigMaps to be annotated

faisal-memon:
We use Spinnaker and without the annotation, spinnaker will rename the configmap.

PR #351, opened 2023-06-16
Add configmap annotation to spire-bundle configmap

faisal-memon:
In PR https://github.com/spiffe/helm-charts/pull/272 I added support for configmaps to be annotated. This is needed to deploy with spinnaker. I forgot to add the annotation to the spire-bundle configmap. This pr fills that gap.

### DIFF
--- a/charts/spire/charts/spire-server/templates/bundle-configmap.yaml
+++ b/charts/spire/charts/spire-server/templates/bundle-configmap.yaml
@@ -4,3 +4,7 @@ kind: ConfigMap
 metadata:
   name: {{ include "spire-lib.bundle-configmap" . }}
   namespace: {{ .Values.notifier.k8sbundle.namespace | default $namespace }}
+  {{- with .Values.configMap.annotations }}
+  annotations:
+    {{- toYaml . | nindent 4 }}
+  {{- end }}


charts/spire/charts/spire-server/templates/bundle-configmap.yaml
```diff
@@ -4,3 +4,7 @@ kind: ConfigMap
 metadata:
   name: {{ include "spire-lib.bundle-configmap" . }}
   namespace: {{ .Values.notifier.k8sbundle.namespace | default $namespace }}
+  {{- with .Values.configMap.annotations }}
```
marcofranssen, line 7: These annotations are specific to the bundleConfigmap.

```suggestion
  {{- with .Values.bundleConfigMap.annotations }}
```

marcofranssen (changes requested):
You are missing the update in the valuesFile.

Also see my inline suggestion, to make that possible we need a breaking change, but it will give a better API. Basically aligning this with the way serviceAccounts etc. are defined as well.


We probably should change values.yaml like this

```diff
- bundleConfigmap: spire-bundle
+ bundleConfigmap:
+   name: spire-bundle
+   annotations: {}
```

Furthermore you would require to update the `spire-lib.bundle-configmap` macro and globals sections to make this work.

marcofranssen (approved):
Ignore my earlier comments. I only now saw the PR description with the reference to #272.

My suggestion doesn't make sense now I see the bigger context.

kfox1111 (approved):
I could see it either way. Having annotations for all configmaps in the chart like this or just for bundleConfigMap.

PR #356, opened 2023-06-17
Align the bash image version with other instances for spire-agent

marcofranssen:

### DIFF
--- a/charts/spire/README.md
+++ b/charts/spire/README.md
@@ -234,7 +234,7 @@ Now you can interact with the Spire agent socket from your own application. The 
 | spire-agent.fsGroupFix.image.pullPolicy | string | `"Always"` | The image pull policy |
 | spire-agent.fsGroupFix.image.registry | string | `"cgr.dev"` | The OCI registry to pull the image from |
 | spire-agent.fsGroupFix.image.repository | string | `"chainguard/bash"` | The repository within the registry |
-| spire-agent.fsGroupFix.image.tag | string | `"latest-20230517"` | Overrides the image tag |
+| spire-agent.fsGroupFix.image.tag | string | `"5.2.15"` | Overrides the image tag |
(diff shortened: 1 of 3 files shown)


kfox1111 (approved):
LGTM

kfox1111 (changes requested):
Actually, I think we need more conversation on this. Refound an issue when reviewing the comments you made on the image updater pr.

Its using the bash version here, 5.2.15. if there is a cve in the same version, it shows up with -r2 and the 5.2.15 tag gets overwritten.
```
$ crane ls cgr.dev/chainguard/bash | grep -v sha256 | grep -v latest
5
5.2
5.2.15
5.2.15-r2
```

So, 5.2.15 is a mutable image, not an immutable one. I'd much rather use immutable images that we test against so they don't float on us and have issues with mirroring/inconsistency on users clusters.

faisal-memon (approved):

marcofranssen:
> Actually, I think we need more conversation on this. Refound an issue when reviewing the comments you made on the image updater pr.
> 
> Its using the bash version here, 5.2.15. if there is a cve in the same version, it shows up with -r2 and the 5.2.15 tag gets overwritten.
> 
> ```
> $ crane ls cgr.dev/chainguard/bash | grep -v sha256 | grep -v latest
> 5
> 5.2
> 5.2.15
> 5.2.15-r2
> ```
> 
> So, 5.2.15 is a mutable image, not an immutable one. I'd much rather use immutable images that we test against so they don't float on us and have issues with mirroring/inconsistency on users clusters.
> 
> waitForIt image doesn't even have a versioned tag either.

To get a release out it is important that we align the version with the version we use in the remainder of the chart for the last few releases already. We can the discussion in an issue and decide based on that if we want to change anything in a next release. Consistency is key to have explainable behavior in the chart.

kfox1111 (approved):
We can continue to have this conversation elsewhere.

PR #363, opened 2023-06-22
create CA by default

drewwells:
this is a harmful default. If somebody turns on certManager
    without enabling ca creation, the server will never start.

I upgraded the chart and my server didn't start, tracked it to this default of create enabled: false. Removed it from the test, for a better test of user expectation.


marcofranssen:
Existing users are already managing cert-manager certificates themself. This will break their behavior. Also in my case I need to manage it outside as we have a certain PKI structure on there. We want to leave the default as is and let the user choose to generate the certificate.

PR #374, opened 2023-06-29
Fix bug in cert-manager upstream authority

marcofranssen:
The arguments for default function need to be the other way around


faisal-memon (approved):

kfox1111:
This is another datapoint in why I think having some unit tests would be a good idea. prevent regressions like this.

PR #376, opened 2023-07-01
Fix missing spiffe-csi-driver imagePullSecrets template

LaithLite:
The spiffe-csi-driver sub-chart describes in its `README.md` that it supports an `imagePullSecret` value, however the helm templating functionality is missing in the `daemonset.yaml`.

The following PR adds the missing functionality, replicated using existing logic in other sub-charts.


edwbuck (approved):
Thank you for the fix.

marcofranssen (approved):

PR #385, opened 2023-07-13
Update PSAT requirements documentation 

cccsss01:


kfox1111 (approved):
Very nice. Thanks for the fix. :)

LGTM

cccsss01:
Well, not sure if it's really needed since it requires 1.22 anyways.

charts/spire/README.md
```diff
@@ -27,6 +27,12 @@ A Helm chart for deploying the complete Spire stack including: spire-server, spi
 
 Please note this chart requires `Projected Service Account Tokens` which has to be enabled on your k8s api server.
 
+> **Note** This is enabled by default with newer versions as shown by the existence of
+        - --service-account-issuer
+        - --service-account-key-file
+        - --service-account-signing-key-file
+See https://kubernetes.io/docs/tasks/configure-pod-container/configure-service-account/#serviceaccount-token-volume-projection for more details
```
faisal-memon, line 34:
```suggestion
See [Service Account Token Volume Projection](https://kubernetes.io/docs/tasks/configure-pod-container/configure-service-account/#serviceaccount-token-volume-projection) in the Kubernetes docs for more details.
```

faisal-memon:
> Well, not sure if it's really needed since it requires 1.22 anyways.

@cccsss01 I think you are right. We should probably just remove that whole section to avoid confusion.

marcofranssen:
Change has to be made in https://github.com/spiffe/helm-charts/blob/main/charts/spire/README.md.gotmpl

so when we run helm-docs.sh the README.md is updated.

That is the reason for failing build.

faisal-memon:
@cccsss01 What do you want to do here?

marcofranssen:
We have created a new PR #437 which will move this information to a more clear location for those few usecases where people can't get PSAT to work.

This way we don't confuse the majority of users with instructions that are only required on very old k8s deployments.

faisal-memon:
Will be covered as part of #437 . Thanks @cccsss01 for bring this to our attention.

PR #390, opened 2023-07-18
Add basic unit test framework

kfox1111:


marcofranssen (approved):
LGTM 🚀

.github/workflows/helm-chart-ci.yaml
```diff
@@ -69,6 +69,16 @@ jobs:
             exit 1
           fi
 
+      - name: Setup Go
+        uses: actions/setup-go@v4
```
marcofranssen, line 73:
```suggestion
        uses: actions/setup-go@v4.0.1
        with:
          go-version-file: tests/go.mod
          cache-dependency-path: tests/go.sum
          check-latest: true

```

.github/workflows/helm-chart-ci.yaml
```diff
@@ -69,6 +69,16 @@ jobs:
...
+      - name: Run Unit Tests
+        run: |
+          set +e
+          set -o pipefail
+          go install github.com/onsi/ginkgo/v2/ginkgo@latest
+          cd tests/unit
+          ginkgo
```
marcofranssen, line 80:
```suggestion
        run: ginkgo
        working-directory: ./tests/unit
```

tests/unit/spire_test.go
```diff
@@ -0,0 +1,48 @@
...
+		return nil, err
+	}
+	objs, err := helmengine.Render(chart, v)
+	return objs, err
+}
+
+var _ = Describe("Spire", func() {
+	chart, err := helmloader.Load("../../charts/spire")
+	if err != nil {
+		panic(err)
```
faisal-memon, line 30: Is panic on error normal for this unit test framework? Or is there an `Expect(err).ToBeNil` or something like that?

marcofranssen: Ah yes for other framework I normally use there is a `Expect.NoError(err)`. Probably nicer and might do some nice logging as well.

kfox1111: ./spire_test.go:29:9: Expect.NoError undefined (type func(actual interface{}, extra ...interface{}) "github.com/onsi/gomega/types".Assertion has no field or method NoError)...

kfox1111: Docs make it sound like this should do the trick:
        Expect(err).Should(Succeed())

faisal-memon: How about `Expect(err).ShouldNot(HaveOccurred())`

tests/unit/spire_test.go
```diff
@@ -0,0 +1,46 @@
...
+			objs, err := ValueStringRender(chart, `
+spire-server:
+  upstreamAuthority:
+    certManager:
+      enabled: true
+      issuer_name: abc123
+`)
+			if err != nil {
+				panic(err)
+			}
```
faisal-memon:
```suggestion
			Expect(err).Should(Succeed())
```

PR #391, opened 2023-07-19
support annotations so oidc can be annotated

drewwells:


faisal-memon:
Thanks @drewwells for submitting this fix.

charts/spire/charts/spiffe-oidc-discovery-provider/templates/deployment.yaml
```diff
@@ -6,6 +6,10 @@ metadata:
   namespace: {{ include "spiffe-oidc-discovery-provider.namespace" . }}
   labels:
     {{- include "spiffe-oidc-discovery-provider.labels" . | nindent 4 }}
+  {{- if .Values.annotations }}
+  annotations:
+    {{- toYaml .Values.annotations | nindent 4 }}
+  {{- end }}
```
faisal-memon, line 12: The other templates follow a format like this one here from the [spire server statefulset](https://github.com/spiffe/helm-charts/blob/f461a017012a6268d8d1c996841111627b79497c/charts/spire/charts/spire-server/templates/statefulset.yaml#L31-L33). Would you mind changing this up so it matches?

faisal-memon:
```suggestion
  annotations:
    {{- with .Values.annotations }}
      {{- toYaml . | nindent 4 }}
    {{- end }}
```

kfox1111: @faisal-memon does line 11 there have too much indention now? seems like it has 6 but nindent is set to 4. might break on multiple annotations?

faisal-memon: indentation might be off

kfox1111 (approved):
LGTM

faisal-memon (approved):
Thanks @drewwells for making the requested changes.

PR #397, opened 2023-07-20
Update README.md.gotmpl

cccsss01:


marcofranssen:
Can you please update the existing PR instead of creating a new one. That saves us reviewing time.

PR #400, opened 2023-07-20
Remove misadded lockfile

kfox1111:
The lockfile is causing testing problems down the pipeline. It is a newly introduced to the repo file that was added by mistake. Removing it.


marcofranssen (approved):

faisal-memon (approved):
Do we need to add this to some ignore file?

PR #409, opened 2023-07-26
Removed generation of README.md with the helm_docs tool.

### LINKED ISSUE #405: Remove helm docs from the build / release toolchain
Helm docs has had a number of shortcomings, many of which we have managed to work around.  But, three new items cannot be worked around, without maintaining large sections of the generated Markdown manually.

1. There is no support for suppressing Values.yaml values, which means that the generated table of items is excessively large.
2. There is no support for arbitrary reordering the Values.yaml values, which means that the only orderings that exist are 'in place ordering" and "alphabetical ordering".  This limits the ability to group settings based on functionality, which makes the documentation (a component of the Helm Charts user interface) impossible to design around "areas of functionality".
3. There is no support for adding Values.yaml items that aren't in the Values.yaml into the generated values tables.  This is useful for scenarios where a purposefully missing value (like the trust domain name) is required but will not be in the Values.yaml file (to trigger helm install failures).

After extensive attempts to find solutions to these shortcomings, the only solution that seems viable is either to rewrite helm_docs or to maintain most of the previously generated output manually, replacing the helm_docs generated output.

On July 25th, 2023, the attending members of the maintainers meeting voted unanimously to remove the helm_docs.  There is an additional Issue where some, if not all, of the generation might be implemented by a different markdown generator; but, to facilitate the transition, the scope of this issue is bounded to not include the additional work of adding the new generator.  This issue is the "first" step of a larger overall replacement.

Scope:
1.  Generate the markdown from the helm_docs to maintain released documents as they exist today.
2.  Check those documents in, removing helm_docs and all related helm_docs tooling.
3. Establish a Makefile target test if the original helm docs inputs are newer than the output README.MD.  
    1. If the inputs are newer, the make command should exit in error.  
    2. If the inputs are older, the make command should pass
4. Update review procedures to run the command, to flag if a README.md update may be necessary
    1. If the tooling reports an update may be necessary, the reviewer will preform a diff on the various Values.yaml files
    2. if the various values.yaml files were updated (verified by diff) then the reviewer will fail any review that doesn't include Values.yaml updates.

This is only the minimum to complete this issue.  Improvements on the process are welcome; but, the scope is to transition to manually maintained README.md files.  Scope to transition from manually maintained README.md files to other technologies is constrained to a different Issue, so we can break this effort down into two steps.

edwbuck:
Developers should remove helm-docs from their bin directories, but this commit does not do that; because the helm-docs executable was never under source code control.

Closes #405


Makefile
```diff
@@ -30,7 +33,16 @@ clean-test-leftovers: ## Cleans up any lingering resources in case tests fail ma
 		&>/dev/null || true
 
 .PHONY: test
-test: install-test-deps test-charts test-examples ## Run all chart tests and example tests
+test: unit-test deployment-test ## Run all unit tests and deployment tests
+
+.PHONY: unit-test
+unit-test: README.md ## Run local development tests that must pass prior to submitting a Pull Request
+
```
kfox1111, line 40:
```suggestion
  cd tests/unit; ginkgo

```

kfox1111 (commented):
Should the generated README.md's be checked in at the same time the tool is removed?

kfox1111:
https://github.com/spiffe/helm-charts/blob/main/.github/workflows/helm-chart-ci.yaml#L88-L90 should probably be updated to call:
`make unit-test` instead too, so the readme test is also executed in the gate.

Makefile
edwbuck: Seems a bit out-of-scope for the README.md work, isn't it?

Adding https://github.com/spiffe/helm-charts/issues/410 to capture the request.  Resolving this conversation with the addition of that issue.

kfox1111: perhaps. but the readme check should be incorporated into the gate in this pr. and then you have to tweak one line either way I think.

edwbuck: It's out of scope.  If we don't put some sort of boundary on scope, then eventually anything could go in.  

For example, I would hope that 410 also includes tools to install ginko, and the best time to test those are in 410.

Please proceed past this point, so we can do a "full" and "good" job on the ginko tool chain.

marcofranssen (changes requested):
I don't think it is smart to remove the docs generation. Upstaging docs manually is a big step back.

project/review_process.md
marcofranssen, line 1: This feels like a workaround for CI!

Makefile
```diff
@@ -30,7 +33,16 @@ clean-test-leftovers: ## Cleans up any lingering resources in case tests fail ma
...
+README.md: $(README_SOURCES)
```
marcofranssen, line 41: Workaround for using industry standard tooling for doc generation in Helm projects!

marcofranssen (dismissed):

project/review_process.md
edwbuck: This is part of a four step plan headed by Vivek to migrate to a new documentation tool.  First we remove helm_docs, then we restructure to have sections that follow a person's needs (different image repository, custom volume type, etc.)  Then we use the new markdown generator to generate the new documentation (restoring some automation).

If you attended the last maintainer's meeting, or read the meeting notes, you'll see that all on the call asked for this effort, as the current setup lacks a large number of features that we want in our README.md file (like ability to skip values, arbitrary ordering of the values.yaml entries, more control with the sub-chart presentation, grouping of entries to match specific topics, etc.)

The current auto-generation is basically awful.  It is a massive wall of text that doesn't have any structure.  This is one of three efforts for Vivek to put the new markdown generator in place.

edwbuck: @marcofranssen You are uising the "request changes" feature, but there is no change requested.  Please use the "comment" feature if you aren't actually asking for something.

edwbuck:
Removing the change requested by Marco, as decided in the maintainer's meeting by Faisal, Mariusz, Kevin, and Edwin in the maintainer's meeting of 8/1/2023.

That meeting laid down new guidelines for change requests, namely that they include a described change to a solution, not a call to have solutions be abandoned.  While we all desired Marco's input, Marco is on vacation, and has not responded to the comment in the issues for five days.

kfox1111:
Can you please fix the failing test?

marcofranssen:
As majority agreed in yesterdays meeting we prefer automated documentation can we close this one in favor of #431 that has all ingredients to migrate to another tool.

faisal-memon:
Closing in favor of #431

Issue #413, opened 2023-07-28
Add support for Vault as UpstreamAuthority


LaithLite:
See https://github.com/spiffe/spire/blob/v1.7.0/doc/plugin_server_upstreamauthority_vault.md

Adding the logic similar to:
- https://github.com/spiffe/helm-charts/blob/main/charts/spire/charts/spire-server/templates/configmap.yaml#L61-L73

PR #415, opened 2023-07-31
Add support for Vault UpstreamAuthority plugin - K8s Auth

LaithLite:
This PR enables the usage of Vault as an upstream authority with the usage of Kubernetes Authentication.

This relates to https://github.com/spiffe/helm-charts/issues/413


charts/spire/charts/spire-server/values.yaml
```diff
@@ -235,6 +235,27 @@ upstreamAuthority:
...
+    # -- The URL of the Vault server. (e.g., https://vault.example.com:8443/)
+    vault_addr: ""
+    # -- Name of the Vault namespace. This is only available in the Vault Enterprise.
+    namespace: ""
+    # -- Name of the mount point where PKI secret engine is mounted
+    pki_mount_point: "pki"
+    # -- Path to a CA certificate file used to verify the Vault server certificate. Only PEM format is supported.
+    ca_cert_path: ""
+    # -- If true, the 'ca_cert_path' configuration is omitted and the vault client accepts any server certificates
+    insecure_skip_verify: false
```
faisal-memon, line 249: Should these be in camel case?

faisal-memon:
@LaithLite Thanks for submitting, this will be a useful feature for a lot of people

charts/spire/charts/spire-server/values.yaml
LaithLite: Woops, yep will correct this 👍

charts/spire/charts/spire-server/values.yaml
```diff
@@ -235,6 +235,27 @@ upstreamAuthority:
...
+    pkiMountPoint: "pki"
+    # -- Path to a CA certificate file used to verify the Vault server certificate. Only PEM format is supported.
+    caCertPath: ""
+    # -- If true, the 'caCertPath' configuration is omitted and the vault client accepts any server certificates
+    insecureSkipVerify: false
+
+    k8s_auth:
+      enabled: false
+      # -- Name of the mount point where the Kubernetes auth method is mounted
+      k8s_auth_mount_point: "kubernetes"
```
kfox1111, line 254: Can these be camel cased too please?

LaithLite:
With regards to `ca_cert_path` should the helm chart user:
-  specify their own certificates, volume (via `spire-server.extraVolumes` and volume mount via `spire-server.extraVolumeMounts` to make this option work 

OR

- Make assumptions and generate templating that creates the volume & volumeMount (which mounts a k8s secret/certificate resource) that then refers to a pre-made secret/cert resource by the helm chart user?

faisal-memon:
> With regards to `ca_cert_path` should the helm chart user:
> 
> * specify their own certificates, volume (via `spire-server.extraVolumes` and volume mount via `spire-server.extraVolumeMounts` to make this option work
> 
> OR
> 
> * Make assumptions and generate templating that creates the volume & volumeMount (which mounts a k8s secret/certificate resource) that then refers to a pre-made secret/cert resource by the helm chart user?

I like the 2nd option of those 2. Can also go a step further and pass in the ca cert by a variable and create the secret. We do this with e[xternal database password](https://github.com/spiffe/helm-charts/blob/d2e1606286a19b79f0c22069da08dcb8b6da4e10/charts/spire/charts/spire-server/values.yaml#L125). The first option seems like less work so if you prefer do that, i think thats fine too.

faisal-memon:
> That's messy I think. How do we support the case where the user already has loaded in the cert/secret by some means? (For example: https://external-secrets.io/latest/). We could support both specifying them in values and an external reference I suppose.

Valid point @kfox1111.

faisal-memon:
Hi @LaithLite The secret option seems a little cleaner to me than exposing a volume. @kfox1111 @marcofranssen Any thoughts on this one?

kfox1111:
Was looking for a precedent here... We have https://github.com/spiffe/helm-charts/blob/main/charts/spire/charts/spire-server/values.yaml#L425-L430 already in tree.

That supports both secret and configmap, since some engines may put the ca pem in a configmap (safe as they are public keys only)

marcofranssen:
@LaithLite are you planning on moving this forward? Should one of our team pull it forward? Please let us know.

LaithLite:
> @LaithLite are you planning on moving this forward? Should one of our team pull it forward? Please let us know.

Apologies I have been busy, I am intending to make this PR more active and will move this forward!

kfox1111:
> Apologies I have been busy, I am intending to make this PR more active and will move this forward!

No worries. We just wanted to make sure you weren't stuck.

kfox1111:
Updated example link: https://github.com/spiffe/helm-charts/blob/main/charts/spire/charts/spire-server/values.yaml#L452-L457

charts/spire/charts/spire-server/templates/statefulset.yaml
```diff
@@ -167,7 +167,19 @@ spec:
...
             {{ end }}
-            {{ end }}
+            {{- end }}
+            {{- with .Values.upstreamAuthority.vault }}
+            {{- if eq (.enabled | toString) "true" }}
+            {{- if eq (.k8sAuth.enabled | toString) "true" }}
+            - name: spire-psat
+              mountPath: /run/spire/secrets/kubernetes.io/serviceaccount/{{ .k8sAuth.token.name }}
+            {{- end }}
+            {{- if ne (.insecureSkipVerify | toString) "true" }}
```
marcofranssen, line 177: Shouldn't we just always mount this volume?

charts/spire/charts/spire-server/templates/statefulset.yaml
```diff
@@ -354,6 +366,33 @@ spec:
             path: {{ .Values.persistence.hostPath }}
             type: Directory
         {{- end }}
+        {{- with .Values.upstreamAuthority.vault }}
+        {{- if eq (.enabled | toString) "true" }}
+        {{- if ne (.insecureSkipVerify | toString) "true" }}
```
marcofranssen, line 371: Same question as before.

charts/spire/charts/spire-server/templates/configmap.yaml
```diff
@@ -157,6 +157,38 @@ plugins:
...
+  {{- if eq (.enabled | toString) "true" }}
+  {{- $upstreamAuthorityUsed = add1 $upstreamAuthorityUsed }}
+  UpstreamAuthority:
+    vault:
+      plugin_data:
+        vault_addr: {{ .vaultAddr | quote }}
+        pki_mount_point: {{ .pkiMountPoint | quote }}
+        insecure_skip_verify: {{ .insecureSkipVerify }}
+        {{- if ne (.insecureSkipVerify | toString) "true" }}
+        ca_cert_path: "/run/spire/vault-upstream/tls.crt"
```
marcofranssen, line 170: shouldn't the ca be `ca.crt`

e.g. comparing this to secrets created via cert-manager they have following keys.

```
ca.crt
tls.crt
tls.key
```

In that case tls.crt could be e.g. the hashi vault server certificate when using cert-manager to create a certificate for hashi-vault.

charts/spire/charts/spire-server/values.yaml
```diff
@@ -263,6 +263,38 @@ upstreamAuthority:
...
+    vaultAddr: ""
+    # -- Name of the Vault namespace. This is only available in the Vault Enterprise.
+    namespace: ""
+    # -- Name of the mount point where PKI secret engine is mounted
+    pkiMountPoint: "pki"
+    # -- If true, caCert options are ignored and Spire accepts any server certificates claiming to be Vault
+    insecureSkipVerify: false
+    # -- The CA Certificate file used to verify the Vault server certificate. Only PEM format is supported
+    caCert:
+      # -- Type of resource representing the Vault server certificate, options are 'Secret' or 'Configmap', the item must be named `tls.crt`
```
marcofranssen, line 278: See previous comment. SHouldn't this be `ca.crt`?

marcofranssen (changes requested):
Thanks for the progress on this one. I have a few inline questions and suggestions. Let me know your thoughts on those.

charts/spire/charts/spire-server/templates/statefulset.yaml
LaithLite: Agreed, I will be changing this now

charts/spire/charts/spire-server/templates/configmap.yaml
LaithLite: Ah I did not see this, yes I will update this to make it standard 👍

charts/spire/charts/spire-server/templates/statefulset.yaml
LaithLite: I recently tested the PSAT mounting, it seems I can't mount it in default service account locations such as:
- `/run/spire/secrets/kubernetes.io/serviceaccount`
- `/var/run/spire/secrets/kubernetes.io/serviceaccount`

Would the directory of `/run/spire/secrets/kubernetes.io/psat` be sufficient?

LaithLite: I've decided to change the path to `/var/run/secrets/tokens/`, which is also used by the Spire Agent and its mounted PSAT.

Let me know if you would prefer something else, otherwise feel free to resolve 👍

LaithLite: > Shouldn't we just always mount this volume?

Now that I think about it, the only time we use a PSAT with the Spire Server is for attestation purposes when communicating with Vault. If we left this volume mounted, it serves no other purpose (that I currently see), which may cause confusion for operators.

What are your thoughts?

charts/spire/charts/spire-server/values.yaml
```diff
@@ -327,6 +327,37 @@ upstreamAuthority:
...
+    vaultAddr: ""
+    ## @param upstreamAuthority.vault.namespace Name of the Vault namespace. This is only available in the Vault Enterprise.
+    namespace: ""
+    ## @param upstreamAuthority.vault.pkiMountPoint Name of the mount point where PKI secret engine is mounted
+    pkiMountPoint: "pki"
+      ## @param upstreamAuthority.vault.insecureSkipVerify If true, caCert options are ignored and Spire accepts any server certificates claiming to be Vault
+    insecureSkipVerify: false
+    caCert:
+      ## @param upstreamAuthority.vault.caCert.type Type of resource representing the Vault server certificate, options are 'Secret' or 'Configmap', the item must be named `ca.crt`
+      type: ConfigMap
```
marcofranssen, line 343: shall we default this to Secret instead, that would make it work out of the box when pointing to a cert-manager secret.

charts/spire/charts/spire-server/values.yaml
```diff
@@ -327,6 +327,37 @@ upstreamAuthority:
...
+
+    k8sAuth:
+      ## @param upstreamAuthority.vault.k8sAuth.enabled Enable k8s authentication to Hashicorp Vault
+      enabled: false
+      ## @param upstreamAuthority.vault.k8sAuth.k8sAuthMountPoint Name of the mount point where the Kubernetes auth method is mounted
+      k8sAuthMountPoint: "kubernetes"
+      ## @param upstreamAuthority.vault.k8sAuth.k8sAuthRoleName Required - Name of the Vault role. The plugin authenticates against the named role
+      k8sAuthRoleName: ""
+      token:
+        ## @param upstreamAuthority.vault.k8sAuth.token.audience Intended audience of the PSAT, it must match one of the audiences supported by the Kubernetes API server. If no audience is specified, it defaults to the identifier of API Server.
```
kfox1111, line 355: Should tokens intended only to be handed to the kubernetes api server be handed to vault? PSAT allows you to specify the audience to be specific to vault so that it can't be used to talk to kubernetes too. I think we probably should be defaulting the audience to vault and not recommending they match audiences with the api-server?

LaithLite: As I understand it, the Vault k8s auth method takes a SA JWT, then validates this SA JWT via the k8s TokenReview API, before associating Vault policies & roles for the requested resource.

If we have a custom audience i.e. `audiences: [vault]`, we must also have that same audience configured on the k8s API-Server via provided flags. If not, when Vault calls the `TokenReview` API, it will fail to validate the JWT as it wont be equal to one of the string elements in the `audiences` array.

For this reason, I decided to have template explicitly use the default value, as I am unsure of what audiences a helm operator would initially have. Further, in a scenario where the helm operator does not necessarily manage their k8s cluster, the default may be a useful option to enable easy configuration with Vault.

Reference:
https://developer.hashicorp.com/vault/docs/auth/kubernetes

kfox1111: Thanks for the documentation. It helped a bit...

I see the projection here:
https://github.com/spiffe/helm-charts/pull/415/files#diff-d24749b40abb8bfd0633e64bc366aa9803df1473594e446177244f6bcfb316deR385

but the audience is optional. Which is odd to me for a projected token. I understand reading the docs that vault is doing extra k8s validation so needs it to be in the apiserver audience list of the api server. Is it ever the right thing though to use the apiserver token directly rather then a projected token for the vault service? Feels like you wouldn't be able to easily revoke the vault client should you need to.

So, it still makes sense to me to always project to some client value. If it is the wrong value for the user, they can always configure it themselves? So, an audience of something like 'vault' as a default feels like maybe the right thing to do?

In the case where they don't control the apiserver, they could still set the audience back to 'apiserver' manually to support that case. But more explicitly accepting the tradeoffs that need to be made in that case? Yeah, its a little harder to use, but also more explicit that they are taking the risks?

LaithLite: Agreed, I will default the audience to `vault` now

kfox1111:
Other then the docs needing to get fixed, it looks good to me. Thanks!

kfox1111 (approved):
LGTM

faisal-memon (approved):
Thanks @LaithLite for this contribution.

marcofranssen (approved):
LGTM 🚀

PR #425, opened 2023-08-04
domains can not include URIs fixes #424

### LINKED ISSUE #424: oidc is invalid when jwtIssuer includes protocol
`jwtIssuer` was included as a domain in the oidc configuration. The problem is that https:// is not valid here, so oidc can not start when a jwtIssuer is defined with `https://`

Example
```
      "domains": [
        "spire-oidc-spiffe-oidc-discovery-provider",
        "spire-oidc-spiffe-oidc-discovery-provider.spire-server",
        "spire-oidc-spiffe-oidc-discovery-provider.spire-server.svc.cluster.local",
        "https://spire-env-4.test.infoblox.com",
        "localhost"
      ],
```

```
$ k -n spire-server logs -f spire-spiffe-oidc-discovery-provider-f5874f988-sxxpv
domain "https://spire-env-4.test.infoblox.com" is not a valid domain name: idna: disallowed rune U+003A
```

The issue is that protocols are not allowed and jwtIssuer can be a URI https://datatracker.ietf.org/doc/html/rfc7519#section-4.1.1

drewwells:


kfox1111:
This doesn't fix it, just breaks it in a different way.... (see production test failure) The issue I think is the jwtIssuer can have a protocol, but the list here can't? I think we just need to strip off the protocol if it exists?

drewwells:
> This doesn't fix it, just breaks it in a different way.... (see production test failure) The issue I think is the jwtIssuer can have a protocol, but the list here can't? I think we just need to strip off the protocol if it exists?

jwtIssuer in the test isn't a real URL, so it's not being consumed. It's just calling localhost from what I can tell https://github.com/spiffe/helm-charts/commit/c1b1dd3d88d97030dc70356a3f4be4c21c15f0ff#diff-c49e9f252dd5404974cdbe903c459ac8d149d730526f210eab4a52fa350de636R36

drewwells:
I can't get this chart to install on kind... so it's hard to test these locally. I think something is off with how the production test is working. I did install this version of the chart with our production config successfully.

kfox1111:
The idea was to be able to specify the setting once. not once for the issuer and a second time in the domain list.

It is set by default in the chart here:
https://github.com/spiffe/helm-charts/blob/main/charts/spire/values.yaml#L12

so the domains list is the set of the jwtIssuer and the additional domains and this is breaking that.

So, I think the right thing to do is still to strip off the protocol off the jwtIssuer if it exists when used in the domain list, so we can set it just once.

kfox1111:
> I can't get this chart to install on kind... so it's hard to test these locally. I think something is off with how the production test is working. I did install this version of the chart with our production config successfully.

if you bring up a fresh kind cluster, then from the root of the repo, run:
./examples/production/run-tests.sh

Does that work? (Maybe thats what your doing?)

I don't use kind, but minikube, and it does work for me in that environment. But there shouldn't be much difference between the true.

drewwells:
The rendering has a bug with empty array, fixing it.

charts/spire/charts/spiffe-oidc-discovery-provider/templates/configmap.yaml
```diff
@@ -2,14 +2,17 @@
...
 domains:
   - "{{ include "spiffe-oidc-discovery-provider.fullname" . }}"
   - "{{ include "spiffe-oidc-discovery-provider.fullname" . }}.{{ include "spiffe-oidc-discovery-provider.namespace" . }}"
   - "{{ include "spiffe-oidc-discovery-provider.fullname" . }}.{{ include "spiffe-oidc-discovery-provider.namespace" . }}.svc.{{ include "spire-lib.cluster-domain" . }}"
-  {{- $jwtIssuer := include "spire-lib.jwt-issuer" . }}
-  {{- uniq (concat (list $jwtIssuer) .Values.config.additionalDomains) | toYaml | nindent 2 }}
+  - "{{- default $uri.path $uri.host }}"
+  {{- with .Values.config.additionalDomains }}
+  {{- toYaml . | nindent 2 }}
+  {{- end }}
```
kfox1111, line 15: How about:
```suggestion
  {{- uniq (concat (list $uri.host) .Values.config.additionalDomains) | toYaml | nindent 2 }}
```

Then it still has the deduplication logic.

examples/production/values.yaml
```diff
@@ -2,6 +2,7 @@ global:
   telemetry:
     prometheus:
       enabled: true
+  jwtIssuer: https://oidc-discovery.example.org
```
kfox1111, line 5: Maybe this in charts/spire/values.yaml instead?

charts/spire/charts/spiffe-oidc-discovery-provider/templates/configmap.yaml
drewwells: I don't believe this is a solution looking for a problem. If somebody has added their jwtIssuer manually, then it will have the same domain twice. Does that cause any issues?
Also, $uri.host is only used when a protocol is defined. A non-protocol string comes in as `$url.path`

examples/production/values.yaml
drewwells: I figured this would create less pushback and ensure my charts aren't broken in future upgrades :D

charts/spire/charts/spiffe-oidc-discovery-provider/templates/configmap.yaml
kfox1111: 🤷 I was asked to put the uniq in there when I added the code in my own pr. I'd expect the same pushback if its removed?

examples/production/values.yaml
kfox1111: Its already being set to oidc-discovery.example.org in charts/spire/values. if its wrong to not have the https://, then it should probably be corrected there?

kfox1111 (approved):
LGTM

marcofranssen (approved):
LGTM 🚀

PR #432, opened 2023-08-08
this job is incompatible with helm template

drewwells:
If the chart is rendered with `helm template | kubectl apply -f -`, this job deletes the deployment that was just created. We should be leaving the lifecycle of objects up to helm rather than forcing behavior with a job.


faisal-memon:
#434 is opened to put the guard around the hooks. @drewwells Is it ok to close this now?

PR #434, opened 2023-08-08
Allow job hooks to be disabled

faisal-memon:
Github wouldn't let me reopen #203 so creating a new pr. Here is the description from the original PR.

This eliminates a bunch of code for folks that don't need the hooks:

```
$ helm template charts/spire/ | wc -l
1150
$ helm template charts/spire/ --set spire-server.controllerManager.upgradeHooks.enabled=false | wc -l
868
```
It also runs faster in some environments.

### DIFF
--- a/charts/spire/charts/spiffe-oidc-discovery-provider/values.yaml
+++ b/charts/spire/charts/spiffe-oidc-discovery-provider/values.yaml
@@ -136,7 +136,7 @@ serviceAccount:
   name: ""
 
 helmHooks:
-  # -- Enable jobs to run at upgrade time to work around issues with deleting
+  # -- Enable Helm hooks to  autofix common issues (should be disabled when using `helm template`)
   enabled: true
 
 autoscaling:
--- a/charts/spire/charts/spiffe-oidc-discovery-provider/values.yaml
+++ b/charts/spire/charts/spiffe-oidc-discovery-provider/values.yaml
@@ -135,6 +135,10 @@ serviceAccount:
   # If not set and create is true, a name is generated using the fullname template
   name: ""
 
+upgradeHooks:
+  # -- Enable jobs to run at upgrade time to work around issues with deleting
+  enabled: true
+
 autoscaling:
   enabled: false
   minReplicas: 1
--- a/charts/spire/charts/spiffe-oidc-discovery-provider/values.yaml
+++ b/charts/spire/charts/spiffe-oidc-discovery-provider/values.yaml
@@ -135,7 +135,7 @@ serviceAccount:
   # If not set and create is true, a name is generated using the fullname template
   name: ""
 
-upgradeHooks:
+helmHooks:
   # -- Enable jobs to run at upgrade time to work around issues with deleting
   enabled: true
 
--- a/charts/spire/charts/spiffe-oidc-discovery-provider/values.yaml
+++ b/charts/spire/charts/spiffe-oidc-discovery-provider/values.yaml
@@ -135,8 +135,8 @@ serviceAccount:
   # If not set and create is true, a name is generated using the fullname template
   name: ""
 
-helmHooks:
-  # -- Enable Helm hooks to  autofix common issues (should be disabled when using `helm template`)
+deleteHook:
+  # -- Enable Helm hooks to autofix common delete issues (should be disabled when using `helm template`)
   enabled: true
 
 autoscaling:
--- a/charts/spire/charts/spire-server/templates/controller-manager-webhook.yaml
+++ b/charts/spire/charts/spire-server/templates/controller-manager-webhook.yaml
@@ -10,7 +10,7 @@ webhooks:
         name: {{ include "spire-controller-manager.fullname" . }}-webhook
         namespace: {{ include "spire-server.namespace" . }}
         path: /validate-spire-spiffe-io-v1alpha1-clusterfederatedtrustdomain
-    {{- if eq (.Values.controllerManager.upgradeHooks.enabled | toString) "true" }}
+    {{- if eq (.Values.controllerManager.installAndUpgradeHook.enabled | toString) "true" }}
     failurePolicy: Ignore # Actual value to be set by post install/upgrade hooks
     {{- else }}
     failurePolicy: {{ .Values.controllerManager.validatingWebhookConfiguration.failurePolicy }}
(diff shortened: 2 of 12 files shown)


kfox1111:
Maybe since there are delete hooks now too, we s/upgradeHooks/helmHooks/?

drewwells (approved):
LGTM, no jobs created. helm template is working! 🥳

kfox1111:
We could add a global for it as well, so it could be specified per chart, or all at once in one global.spire.enableHooks.

faisal-memon:
> We could add a global for it as well, so it could be specified per chart, or all at once in one global.spire.enableHooks.

Yes, a global is a good idea.

charts/spire/charts/spiffe-oidc-discovery-provider/values.yaml
```diff
@@ -135,6 +135,10 @@ serviceAccount:
   # If not set and create is true, a name is generated using the fullname template
   name: ""
 
+upgradeHooks:
```
marcofranssen, line 138: These are not just upgrade hooks. They are a combination of pre-delete and post-install and post-upgrade hooks.

```suggestion
helmHooks:
```

charts/spire/charts/spiffe-oidc-discovery-provider/values.yaml
```diff
@@ -135,6 +135,10 @@ serviceAccount:
...
+  # -- Enable jobs to run at upgrade time to work around issues with deleting
```
marcofranssen, line 139:
```suggestion
  # -- Enable Helm hooks to  autofix common issues (should be disabled when using `helm template`)
```

charts/spire/charts/spire-server/values.yaml
```diff
@@ -243,6 +243,9 @@ notifier:
 
 controllerManager:
   enabled: false
+  upgradeHooks:
+    # -- Enable jobs to run at upgrade time to work around issues with custom resources
```
marcofranssen, line 247:
```suggestion
    # -- Enable Helm hooks to  autofix common issues (should be disabled when using `helm template`)
```

charts/spire/charts/spire-server/templates/post-upgrade-hook.yaml
```diff
@@ -1,3 +1,4 @@
+{{- if eq (.Values.controllerManager.upgradeHooks.enabled | toString) "true" }}
```
marcofranssen, line 1:
```suggestion
{{- if eq (.Values.controllerManager.helmHooks.enabled | toString) "true" }}
```

charts/spire/charts/spire-server/templates/post-install-hook.yaml
```diff
@@ -1,3 +1,4 @@
+{{- if eq (.Values.controllerManager.upgradeHooks.enabled | toString) "true" }}
```
marcofranssen, line 1:
```suggestion
{{- if eq (.Values.controllerManager.helmHooks.enabled | toString) "true" }}
```

charts/spire/charts/spire-server/templates/controller-manager-webhook.yaml
```diff
@@ -10,7 +10,11 @@ webhooks:
         name: {{ include "spire-controller-manager.fullname" . }}-webhook
         namespace: {{ include "spire-server.namespace" . }}
         path: /validate-spire-spiffe-io-v1alpha1-clusterfederatedtrustdomain
+    {{- if eq (.Values.controllerManager.upgradeHooks.enabled | toString) "true" }}
```
marcofranssen, line 13:
```suggestion
    {{- if eq (.Values.controllerManager.helmHooks.enabled | toString) "true" }}
```

drewwells:
For what it is worth, `upgradeHooks` is a better name. These are not normal jobs, they are a series of hacks to force upgrades around issues with the csi driver. If anything, I'd add more information about the csi issue in the description, and link to k8s defect if such a thing exists.

kfox1111:
> All of the above.
> 
> ```
> hooks:
>   enabled: true
>   upgradeHooks: true
>   preDeleteHooks: true
>   postInstallHooks: true
> ```

how would enabled play with the others? kind of need a trinary for that.

drewwells:
`{{ if .Values.hooks.enabled AND .Values.hooks.upgradeHooks.enabled }}`

So you can turn off all hooks or just a section of them. Can default to on, but I feel confident that all of these hooks can be deleted :).

kfox1111:
> `{{ if .Values.hooks.enabled AND .Values.hooks.upgradeHooks.enabled }}`
> 
> So you can turn off all hooks or just a section of them. Can default to on, but I feel confident that all of these hooks can be deleted :).

Hmm... Yeah, I like the usability of that model. But its not the most obvious from just the values themselves, so a user could get quite confused on what they need to set?

kfox1111:
Thinking about it some more... how about:
```
hooks:
  disableAll: false
  enable:
    upgrade: true
    preDelete: true
    postInstall: true
```

Then its obvious if you want to disable them all, you set hooks.disableAll=true. If its not true, then each are configured independently and its obvious they are enabled by default?

drewwells:
`enable/enabled` keys are more common across helm charts because they're easier to understand. There's more cognitive overhead thinking about disabling a disableAll flag. Double negative guard statement :)

kfox1111:
enable is more common, yeah.

but its unclear the behavior of enable: true with upgradeHooks: false at a glance.

I'd guess it would enable all the hooks, but it wouldn't have that behavior... it only enables the other flags. or put another way, setting it to false is the only setting that has a direct affect of disabling all the hooks.

There something else we could use to make it more clear?

drewwells:
They can read the template code to see what it does. `enable` makes any hooks possible. `enable` + `upgradeHook` is required for upgradeHooks to be activated

faisal-memon:
Is there a use case for those fine grained controls? For spinnaker you just need to disable them all.

marcofranssen:
> Is there a use case for those fine grained controls? For spinnaker you just need to disable them all.

Good question. If there is no usecase I prefer to merge as currently implemented.

kfox1111:
> Is there a use case for those fine grained controls? For spinnaker you just need to disable them all.

For my pipelines, I can see benefit in disabling just the upgrade hooks, as those run all the time and are costly to the runtime length of the pipeline. The delete hooks only happen once on delete, so I'm ok leaving those enabled.

faisal-memon:
Updated to address comments. I combined install and upgrade hooks together since they basically are addressing different manifestations of the same root issue. So there are 2 globals: `installAndUpgradeHooks` and `deleteHooks`. I hope that works for everyone.

marcofranssen (approved):
LGTM 🚀

PR #435, opened 2023-08-08
Adds AWS KMS KeyManager support

### LINKED ISSUE #403: Add support for AWS KMS as KeyManager
Would be nice for the helm charts to have support for aws_kms KeyManager.

https://github.com/spiffe/spire/blob/main/doc/plugin_server_keymanager_aws_kms.md

I'd like to work on this if the team see this as a good addition.

mchurichi:
This adds toggles for both `disk` and `aws_kms` key managers. It keeps `disk` as the default one for backward compatibility, but one and only one must be enabled, so `disk` should be explicitly disabled if `aws_kms` is enabled.

Closes #403


charts/spire/README.md
```diff
@@ -362,6 +362,12 @@ Now you can interact with the Spire agent socket from your own application. The
 | spire-server.ingress.tls | list | `[]` |  |
 | spire-server.initContainers | list | `[]` |  |
 | spire-server.jwtIssuer | string | `"oidc-discovery.example.org"` | The JWT issuer domain |
+| spire-server.keyManager.awsKMS.enabled | bool | `false` |  |
+| spire-server.keyManager.awsKMS.keyMetadataFile | string | `"/run/spire/data/aws-kms-key-metadata"` |  |
```
kfox1111, line 366: Not sure we want the keyMetadataFile configurable. It exists in our own volumes, so all the user can do is misconfigure it to be outside the volume?

charts/spire/README.md
```diff
@@ -362,6 +362,12 @@ Now you can interact with the Spire agent socket from your own application. The
...
+| spire-server.keyManager.awsKMS.keyPolicyFile | string | `""` |  |
+| spire-server.keyManager.awsKMS.region | string | `""` |  |
+| spire-server.keyManager.disk.enabled | bool | `true` |  |
+| spire-server.keyManager.disk.keysPath | string | `"/run/spire/data/keys.json"` |  |
```
kfox1111, line 370: Not sure we want the keysPath configurable. It exists in our own volumes, so all the user can do is misconfigure it to be outside the volume?

charts/spire/README.md
```diff
@@ -362,6 +362,12 @@ Now you can interact with the Spire agent socket from your own application. The
 | spire-server.ingress.tls | list | `[]` |  |
 | spire-server.initContainers | list | `[]` |  |
 | spire-server.jwtIssuer | string | `"oidc-discovery.example.org"` | The JWT issuer domain |
+| spire-server.keyManager.awsKMS.enabled | bool | `false` |  |
+| spire-server.keyManager.awsKMS.keyMetadataFile | string | `"/run/spire/data/aws-kms-key-metadata"` |  |
+| spire-server.keyManager.awsKMS.keyPolicyFile | string | `""` |  |
```
kfox1111, line 367: Should the keyPolicyFile be a configmap reference instead so the user can upload it?

charts/spire/charts/spire-server/templates/configmap.yaml
```diff
@@ -46,11 +47,41 @@ plugins:
...
+  {{- with .Values.keyManager.awsKMS }}
+  {{- if eq (.enabled | toString) "true" }}
+  {{- $keyManagerUsed = add1 $keyManagerUsed }}
+  KeyManager:
+    - aws_kms:
+        plugin_data:
+          region: {{ .region | quote }}
+          key_metadata_file: {{ .keyMetadataFile | quote }}
+          {{- if ne .accessKeyID "" }}
+          access_key_id: {{ .accessKeyID | quote }}
```
kfox1111, line 70: This is placing the potentially senstive key/secret into the configmap. They should be in a secret instead. Here, you could do:
```suggestion
          access_key_id: "${AWSKM_ACCESS_KEY_ID}"
```

charts/spire/charts/spire-server/templates/configmap.yaml
```diff
@@ -46,11 +47,41 @@ plugins:
...
+          {{- end }}
+          {{- if ne .secretAccessKey "" }}
+          secret_access_key: {{ .secretAccessKey | quote }}
```
kfox1111, line 73:
```suggestion
          secret_access_key: "${AWSKM_SECRET_ACCESS_KEY}"
```

kfox1111: Then in secrets.yaml in the same folder, add:
```
  AWSKM_ACCESS_KEY_ID: {{ .accessKeyID | quote }}
  AWSKM_SECRET_ACCESS_KEY: {{ .secretAccessKey | quote }}
```

charts/spire/README.md
mchurichi: that makes sense, we can set a good value and hide this detail

charts/spire/README.md
mchurichi: Do you mean it can be just the name of a configmap instead of a file path? something like this

```yaml
awsKMS:
  keyPolicyConfigMap: my-policy
```
then a user can
```yaml
kind: ConfigMap
metadata:
  name: my-policy
data:
  policy.json: # policy here
```
then we mount the configmap, if set, on a volume and hard-code the `keyPolicyFile` in the config of the spire server?

That makes sense, but we'd need to establish a well-known file name (i.e `policy.json`) so we can safely mount it, right?

kfox1111: Or maybe, something like:
```
awsKMS:
  keyPolicy:
    existingConfigMap: ""
    policy: ""
```

If they specify existingConfigMap, then include policy.json from it.
If policy, make a configmap for them with the policy content?

kfox1111 (approved):
This looks great. Thank you!

mchurichi:
Thanks for the swiftness in getting this reviewed! I think I addressed all of the comments.

charts/spire/charts/spire-server/values.yaml
```diff
@@ -175,6 +175,23 @@ ca_subject:
...
+    # -- Policy to use when creating keys. If no policy is specified, a default policy will be used.
+    keyPolicy:
+      # -- Key policy in JSON format.
+      policy: ""
+      # -- Name of a ConfigMap that has a `policy.json` file with the key policy in JSON format.
+      existingConfigMap: ""
+    # -- Access key ID for the AWS account. If empty, the default credential chain will be used.
+    accessKeyID: ""
+    # -- Secret access key for the AWS account. If empty, the default credential chain will be used.
+    secretAccessKey: ""
```
marcofranssen, line 193: I'm not very familiar with this plugin, but a more secure approach to integrate with AWS is to use a role that can be assumed by adding an annotation on the k8s deployment. This way we don't have to expose secrets like the access keys in the cluster + we can define that key policy on the role.

mchurichi: The plugins works with either an access key or assuming a role as you describe. Access key fields are optional.

https://github.com/spiffe/spire/blob/main/doc/plugin_server_keymanager_aws_kms.md#aws-kms-access

mchurichi: Seems like the service account needs to be annotated to assume the IAM role. I wonder if it makes sense to add a convenience field here `assumeRoleARN` so users don't need to do it in  `serviceAccount.annotations`.

https://docs.aws.amazon.com/eks/latest/userguide/associate-service-account-role.html

marcofranssen: Could we document the annotation as preferred option. We can keep the accesskeys approach as a fallback for people who can't use a role.

I would prefer to stick with just the annotations.

On the accesskeys we could enhance the documentation to tell people about the assumeRoleARN annotation as preferred approach.

mchurichi: Sounds good. I just added an extra sentence to make it clear.

faisal-memon (approved):

charts/spire/charts/spire-server/values.yaml
mchurichi: @marcofranssen Are we good with these changes?

marcofranssen (approved):
LGTM 🚀

charts/spire/charts/spire-server/values.yaml
marcofranssen: Thanks, this is great, love the inclusion of that link. 🎉

PR #441, opened 2023-08-10
fix some issues found with kube-score

drewwells:
Statefulsets are supposed to have headless services assigned to them, see https://kubernetes.io/docs/concepts/workloads/controllers/statefulset/#limitations https://kubernetes.io/docs/concepts/services-networking/service/#headless-services

The others changes are probably easier to see why kube-score complains

### DIFF
--- a/charts/spire/charts/spire-server/templates/post-upgrade-hook.yaml
+++ b/charts/spire/charts/spire-server/templates/post-upgrade-hook.yaml
@@ -51,6 +51,9 @@ metadata:
   annotations:
     "helm.sh/hook": post-upgrade
     "helm.sh/hook-delete-policy": before-hook-creation, hook-succeeded, hook-failed
+    {{- with .Values.podAnnotations }}
+      {{- toYaml . | nindent 6 }}
+    {{- end }}
 spec:
   template:
     metadata:
--- a/charts/spire/charts/spire-server/templates/statefulset.yaml
+++ b/charts/spire/charts/spire-server/templates/statefulset.yaml
@@ -51,8 +51,10 @@ spec:
       initContainers:
       {{- if and .Values.upstreamAuthority.certManager.enabled .Values.upstreamAuthority.certManager.ca.create }}
         - name: wait
+          {{- with .Values.tools.kubectl.image.securityContext }}
           securityContext:
-            {{- toYaml .Values.securityContext | nindent 12 }}
+            {{- toYaml . | nindent 12 }}
+          {{- end }}
           image: {{ template "spire-lib.kubectl-image" (dict "appVersion" $.Chart.AppVersion "image" .Values.tools.kubectl.image "global" .Values.global "KubeVersion" .Capabilities.KubeVersion.Version) }}
           args:
             - wait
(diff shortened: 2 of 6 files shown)


drewwells:
Seems like headless service is causing an issue with the test. I haven't been able to reproduce yet
```
 Args:
      -t
      30
      -h
      spire-eltibqgbgf-server.spire-eltibqgbgf
      -p
      8081
    State:          Running
      Started:      Thu, 10 Aug 2023 15:28:21 +0000
    Ready:          False
    Restart Count:  0
    Environment:    <none>
    Mounts:
      /var/run/secrets/kubernetes.io/serviceaccount from kube-api-access-lcgtz (ro)
```

kfox1111:
I asked the api folks a while ago about required headless service and they said it really was optional but couldn't change it to such without breaking compatibility. It doesn't enforce it and a lot of folks never use it. So its extremely unlikely to ever break in the future.

drewwells:
> I asked the api folks a while ago about required headless service and they said it really was optional but couldn't change it to such without breaking compatibility. It doesn't enforce it and a lot of folks never use it. So its extremely unlikely to ever break in the future.

Do you have a link to that in documentation? I can push that up to kube-score and maybe this PR will pass without serviceName change

kfox1111:
https://github.com/kubernetes/kubernetes/issues/69608

charts/spire/charts/spire-server/templates/post-upgrade-hook.yaml
```diff
@@ -51,6 +51,9 @@ metadata:
   annotations:
     "helm.sh/hook": post-upgrade
     "helm.sh/hook-delete-policy": before-hook-creation, hook-succeeded, hook-failed
+    {{- with .Values.podAnnotations }}
```
marcofranssen, line 54: I wonder if these annotations need to be separately configurable from the workload itself? 🤔 

Not completely sure though.

charts/spire/charts/spire-server/templates/statefulset.yaml
```diff
@@ -51,8 +51,10 @@ spec:
       initContainers:
       {{- if and .Values.upstreamAuthority.certManager.enabled .Values.upstreamAuthority.certManager.ca.create }}
         - name: wait
+          {{- with .Values.tools.kubectl.image.securityContext }}
```
marcofranssen, line 54:
```suggestion
          {{- with .Values.tools.kubectl.securityContext }}
```

should be one level up.

charts/spire/charts/spire-server/values.yaml
```diff
@@ -34,6 +34,8 @@ serviceAccount:
   # If not set and create is true, a name is generated using the fullname template
   name: ""
 
+jobAnnotations: {}
```
marcofranssen, line 37: From a practical point of view. 

- When will we use this?
- Do we need to be able to annotate differently per job?
- If so do we need post-upgrade-job annotations and pre-delete job annotations etc?

This is an open question, no thought or ideas on that myself. Trying to ensure we aren't adding things that need changes soon again, so we can keep the API stable.

marcofranssen (changes requested):
Few inline questions and suggestions.

faisal-memon:
@drewwells What do you think of putting the headless service behind a flag? Seems like not everyone will need it.

marcofranssen:
In that case just deploy it. No need for more flags

Issue #442, opened 2023-08-10
ACTION REQUIRED: Changes to pulling Chainguard Images


chainguard-alerter:
Hey there [Chainguard](https://github.com/chainguard-dev) here.

We noticed that you are using [Chainguard Images](https://github.com/chainguard-images/images), thank you! We wanted to make you aware of an upcoming change that will impact your project.

Starting August 16, 2023 public users will no longer be able to pull images from our registry (cgr.dev/chainguard) by tags other than `latest` or `latest-dev`. Please see [the announcement](https://www.chainguard.dev/unchained/scaling-chainguard-images-with-a-growing-catalog-and-proactive-security-updates) for more information.

You are currently using the following.

In https://github.com/spiffe/helm-charts/blob/a167ce68901a2b08d44367aa3a6da7612a64afc6/charts/spire/charts/tornjak-frontend/values.yaml:

- cgr.dev/chainguard/bash:5.2.15

In https://github.com/spiffe/helm-charts/blob/a167ce68901a2b08d44367aa3a6da7612a64afc6/charts/spire/charts/spire-server/values.yaml:

- cgr.dev/chainguard/bash:5.2.15

In https://github.com/spiffe/helm-charts/blob/a167ce68901a2b08d44367aa3a6da7612a64afc6/charts/spire/charts/spire-agent/values.yaml:

- cgr.dev/chainguard/wait-for-it:latest-20230517
- cgr.dev/chainguard/bash:5.2.15

In https://github.com/spiffe/helm-charts/blob/a167ce68901a2b08d44367aa3a6da7612a64afc6/charts/spire/charts/spire-agent/README.md:

- cgr.dev/chainguard/wait-for-it:latest-20230517
- cgr.dev/chainguard/bash:5.2.15

In https://github.com/spiffe/helm-charts/blob/a167ce68901a2b08d44367aa3a6da7612a64afc6/charts/spire/charts/spiffe-oidc-discovery-provider/values.yaml:

- cgr.dev/chainguard/bash:5.2.15

In https://github.com/spiffe/helm-charts/blob/a167ce68901a2b08d44367aa3a6da7612a64afc6/charts/spire/README.md:

- cgr.dev/chainguard/wait-for-it:latest-20230517
- cgr.dev/chainguard/bash:5.2.15

[...]

PR #443, opened 2023-08-10
Fix chainguard image references as per issue 442

kfox1111:
fixes: https://github.com/spiffe/helm-charts/issues/442

### DIFF
--- a/charts/spire/README.md
+++ b/charts/spire/README.md
@@ -237,7 +237,7 @@ Now you can interact with the Spire agent socket from your own application. The 
 | spire-agent.fsGroupFix.image.pullPolicy | string | `"Always"` | The image pull policy |
 | spire-agent.fsGroupFix.image.registry | string | `"cgr.dev"` | The OCI registry to pull the image from |
 | spire-agent.fsGroupFix.image.repository | string | `"chainguard/bash"` | The repository within the registry |
-| spire-agent.fsGroupFix.image.tag | string | `"5.2.15"` | Overrides the image tag |
+| spire-agent.fsGroupFix.image.tag | string | `"latest@sha256:96ab1600d945b4a99c8610b5c8b31e346da63dc20573a26bb0777dd0190db5d4"` | Overrides the image tag |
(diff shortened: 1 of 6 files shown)


faisal-memon (approved):
Thanks @kfox1111 for fixing this quickly.

marcofranssen (changes requested):
We agreed the following. Took from meeting notes.

```
- Immutable image references - Decided to go with Version@shasum except where version is not available and then latest@shasum will be used.
```

Both @mrsabath and I raised our concerns on the latest tag which we both have blocked in our clusters. E.g. we have kyverno policies that block latest tags e.g.

Can you please lock the bash image on the `5.2.15@sha256:.....`.

kfox1111:
> We agreed the following. Took from meeting notes.
> 
> ```
> - Immutable image references - Decided to go with Version@shasum except where version is not available and then latest@shasum will be used.
> ```
> 
> Both @mrsabath and I raised our concerns on the latest tag which we both have blocked in our clusters. E.g. we have kyverno policies that block latest tags e.g.
> 
> Can you please lock the bash image on the `5.2.15@sha256:.....`.

Can you please read the issue chainguard created? They have made the decision for us and are forcing the issue. Either you'll need to adapt your policy to allow latest with shas, or we will have to switch away from chainguard entirely.

marcofranssen:
> > We agreed the following. Took from meeting notes.
> 
> > 
> 
> > ```
> 
> > - Immutable image references - Decided to go with Version@shasum except where version is not available and then latest@shasum will be used.
> 
> > ```
> 
> > 
> 
> > Both @mrsabath and I raised our concerns on the latest tag which we both have blocked in our clusters. E.g. we have kyverno policies that block latest tags e.g.
> 
> > 
> 
> > Can you please lock the bash image on the `5.2.15@sha256:.....`.
> 
> 
> 
> Can you please read the issue chainguard created? They have made the decision for us and are forcing the issue. Either you'll need to adapt your policy to allow latest with shas, or we will have to switch away from chainguard entirely.

Do you have a link to the issue?

marcofranssen:
Oh 🤦  found it.

PR #444, opened 2023-08-11
option to set KeyManager memory in spire server

grameshtwilio:
This supports the Keymanager memory plugin for the spire server In addition to the default disk plugin

### DIFF
--- a/charts/spire/charts/spire-server/templates/configmap.yaml
+++ b/charts/spire/charts/spire-server/templates/configmap.yaml
@@ -54,7 +54,7 @@ plugins:
   KeyManager:
     - disk:
         plugin_data:
-          keys_path: {{ .keysPath | quote }}
+          keys_path: "/run/spire/data/keys.json"
   {{- end }}
   {{- end }}
 
--- a/charts/spire/charts/spire-server/templates/configmap.yaml
+++ b/charts/spire/charts/spire-server/templates/configmap.yaml
@@ -1,5 +1,6 @@
 {{- define "spire-server.yaml-config" -}}
 {{- $upstreamAuthorityUsed := 0 }}
+{{- $keyManagerUsed := 0 }}
 {{- $root := . }}
 server:
   bind_address: "0.0.0.0"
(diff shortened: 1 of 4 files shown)


kfox1111:
Thanks for working on this use case.

Can the same effect be done with https://github.com/spiffe/helm-charts/blob/main/charts/spire/charts/spire-server/values.yaml#L105  type=emptyDir already?

grameshtwilio:
@kfox1111 that's right, it should have the same effect. Do you think this could be useful when other KeyManager plugins e.g. aws/gcp kms configurations need to be supported? If not, I will close this PR. Thanks!

kfox1111:
Yeah, I could see it being useful there.... Thinking about it a little more, I think it also may be useful to disable the keymanager while still having the persistence enabled for something like tornjak.

charts/spire/charts/spire-server/templates/configmap.yaml
```diff
@@ -47,10 +48,28 @@ plugins:
...
   {{- end }}
 
+  {{- with .Values.keyManager.disk }}
+  {{- if eq (.enabled | toString) "true" }}
+  {{- $keyManagerUsed = add1 $keyManagerUsed }}
   KeyManager:
     - disk:
         plugin_data:
-          keys_path: "/run/spire/data/keys.json"
+          keys_path: {{ .keysPath | quote }}
```
kfox1111, line 57: Could you please undo the change to the keys_path? It is specific to the PVC and its likely changing it will just enable the user to break the plugin or future upgrades.

grameshtwilio: done!

aarontams (approved):

charts/spire/charts/spire-server/values.yaml
```diff
@@ -175,6 +175,13 @@ ca_subject:
   organization: Example
   common_name: example.org
 
+keyManager:
+  disk:
+    enabled: false
```
kfox1111, line 180: Looks like the defaults are backwards here, and there is an extra newline at the bottom.

grameshtwilio: sorry about that!

kfox1111: No worries. Thanks for fixing it. :)

kfox1111 (approved):
LGTM. Thanks! :)

marcofranssen (approved):

PR #446, opened 2023-08-14
support overriding annotations in jobs

drewwells:
Should be a no-op 
```
helm template charts/spire -f .../src/github.com/spiffe/helm-charts/examples/production/values.yaml --debug | less > before
helm template charts/spire -f .../src/github.com/spiffe/helm-charts/examples/production/values.yaml --debug | less > after
```

```
% diff before after                                           
% echo $?     
0
```

### DIFF
--- a/charts/spire/README.md
+++ b/charts/spire/README.md
@@ -191,6 +191,8 @@ Now you can interact with the Spire agent socket from your own application. The 
 | spiffe-oidc-discovery-provider.insecureScheme.nginx.image.tag | string | `"1.24.0-alpine"` | Overrides the image tag |
 | spiffe-oidc-discovery-provider.insecureScheme.nginx.image.version | string | `""` | This value is deprecated in favor of tag. (Will be removed in a future release) |
 | spiffe-oidc-discovery-provider.insecureScheme.nginx.resources | object | `{}` |  |
+| spiffe-oidc-discovery-provider.jobAnnotations."helm.sh/hook" | string | `"pre-delete"` |  |
+| spiffe-oidc-discovery-provider.jobAnnotations."helm.sh/hook-delete-policy" | string | `"before-hook-creation, hook-succeeded, hook-failed"` |  |
 | spiffe-oidc-discovery-provider.jwtIssuer | string | `"https://oidc-discovery.example.org"` |  |
 | spiffe-oidc-discovery-provider.livenessProbe.initialDelaySeconds | int | `5` | Initial delay seconds for livenessProbe |
 | spiffe-oidc-discovery-provider.livenessProbe.periodSeconds | int | `5` | Period seconds for livenessProbe |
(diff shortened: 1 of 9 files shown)


faisal-memon:
Thanks @drewwells for the contribution. Now that you can disable jobs is this still needed? #434

kfox1111:
I can see adding additional labels, but is changing the existing annotations ever going to be useful outside of breaking things?
```
"helm.sh/hook": "pre-delete"
"helm.sh/hook-delete-policy": "before-hook-creation, hook-succeeded, hook-failed"
```

kfox1111:
> > I can see adding additional labels, but is changing the existing annotations ever going to be useful outside of breaking things?
> > ```
> > "helm.sh/hook": "pre-delete"
> > "helm.sh/hook-delete-policy": "before-hook-creation, hook-succeeded, hook-failed"
> > ```
> 
> It needs to support annotations. I got blocked on this due to spinnaker: https://spinnaker.io/docs/guides/user/kubernetes-v2/best-practices/#version-your-configmaps-and-secrets

not sure how secrets/configmaps docs work with job hooks. But the merged patch to allow disabling hooks should fix the root issue?  Any way I can think of this pr being configured in spinnaker would break things as its deleting things you don't want deleted?

We could have it leave the existing annotations always alone and add additional annotations from the user should they be needed?

drewwells:
In the docs, Spinnaker versions all the configmaps and secrets. It also edits deployment and job specs to change the mount to the new configmap/secret versioned name. You have to annotate the job to turn off this behavior.

Im not understanding why there are questions about making the annotations configurable. Unless we're deleting jobs from the chart, we should be able to mutate their annotations.

drewwells:
See the search results for all the objects that support annotations, jobs needs to be treated as first class objects or removed from the chart. Having them semi-supported in some use cases causes number of issues that made using spire quite difficult for me. https://github.com/search?q=repo%3Aspiffe%2Fhelm-charts%20annotations&type=code

kfox1111:
Just trying to understand the use case better so we're not forgetting things or choosing the wrong api.

I'm convinced we should support additional annotations on these objects. Just not sure making the existing ones overridable is a good idea.

drewwells:
six of one or half dozen of another... 
I'd discourage trying to make too many guard rails in the chart. Offer great defaults and maximum flexibility. It lets people configure the chart to their heart is content and we don't have to make change chart changes to support it. People will find ways to break the helm chart no matter how perfectly crafted our tpl statements are.

charts/spire/README.md
```diff
@@ -191,6 +191,8 @@ Now you can interact with the Spire agent socket from your own application. The
 | spiffe-oidc-discovery-provider.insecureScheme.nginx.image.tag | string | `"1.24.0-alpine"` | Overrides the image tag |
 | spiffe-oidc-discovery-provider.insecureScheme.nginx.image.version | string | `""` | This value is deprecated in favor of tag. (Will be removed in a future release) |
 | spiffe-oidc-discovery-provider.insecureScheme.nginx.resources | object | `{}` |  |
+| spiffe-oidc-discovery-provider.jobAnnotations."helm.sh/hook" | string | `"pre-delete"` |  |
+| spiffe-oidc-discovery-provider.jobAnnotations."helm.sh/hook-delete-policy" | string | `"before-hook-creation, hook-succeeded, hook-failed"` |  |
```
kfox1111, line 195: can these be updated to be jobPreDeleteAnnotations in case we have new types of jobs in the future?

faisal-memon:
> I'm convinced we should support additional annotations on these objects. Just not sure making the existing ones overridable is a good idea.

I agree with @kfox1111 on this one. Those labels are auto generated from the [policy specified along with the job](https://github.com/spiffe/helm-charts/blob/main/charts/spire/charts/spire-server/templates/post-install-hook.yaml#L12-L13), so this duplicates it.  Supporting additional annotations is fine.

charts/spire/README.md
faisal-memon: What do you think of putting under the existing `deleteHook` and `installAndUpgradeHook` configurables?

faisal-memon:
> six of one or half dozen of another... I'd discourage trying to make too many guard rails in the chart. Offer great defaults and maximum flexibility. It lets people configure the chart to their heart is content and we don't have to make change chart changes to support it. People will find ways to break the helm chart no matter how perfectly crafted our tpl statements are.

Valid point too.

drewwells:
I don't agree, closing.

PR #458, opened 2023-08-18
Fix initContainers spire-server statefulset

marcofranssen:
To allow for inplace upgrades of spire-server statefulset we are not allowed to
make changes to all fields of the statefulset spec.

When bumping from 0.8.1 to the latest version the `initContainers:` field is added in the spec and therefore does not allow for updating the statefulset.

This fix prevents the empty initContainers block when bumping from a previous chart version.


kfox1111:
We probably need at least one upgrade test.

I just tested it and adding an init container works fine for a statefulset. So, I don't think this is needed?

charts/spire/charts/spire-server/templates/statefulset.yaml
```diff
@@ -48,6 +48,7 @@ spec:
       shareProcessNamespace: true
       securityContext:
         {{- toYaml .Values.podSecurityContext | nindent 8 }}
+      {{- if or (gt (len .Values.initContainers) 0) (and .Values.upstreamAuthority.certManager.enabled .Values.upstreamAuthority.certManager.ca.create) }}
```
faisal-memon, line 51: Can the if check on line 53 be removed with this change? Seem duplicate.

marcofranssen: No still need it, if not running cert-manager with the create. It should only do the extra unit containers.

faisal-memon (approved):

kfox1111 (approved):
I don't think Kubernetes cares about an empty initContainers section. It seems to work as is.

That being said, I don't think this hurts either.

LGTM

Issue #465, opened 2023-08-22
New image has to be added to update-tags script


marcofranssen:
See for a newly added image.

https://github.com/spiffe/helm-charts/pull/423/files#diff-7be84ce176a12ba3b77f98055916d01a5c97e1688a3a32e361495839b5ef69caR237-R248

PR #466, opened 2023-08-22
Scan for updates to new images

kfox1111:
Fixes: https://github.com/spiffe/helm-charts/issues/465

Also adds support for registries that are blank (IE, docker.io) and fixes a bug with the pip install that installs to the wrong place.


.github/scripts/update-tags.sh
```diff
@@ -46,18 +46,21 @@ jq -r '. | keys[]' "$IMAGEJSON" | while read -r CHART; do
     REGISTRY=$(yq e ".${QUERY}.registry" "$VALUES")
     REPOSITORY=$(yq e ".${QUERY}.repository" "$VALUES")
     VERSION=$(yq e ".${QUERY}.tag" "$VALUES")
```
kfox1111, line 48: These changes allow blank Registries (IE, docker.io). First time we've seen one in the chart.

.github/scripts/update-tags.sh
```diff
@@ -46,18 +46,21 @@ jq -r '. | keys[]' "$IMAGEJSON" | while read -r CHART; do
...
+    if [[ "$REGISTRY" != "" ]]; then
+      REGISTRY="$REGISTRY/"
```
marcofranssen, line 50: 👍

marcofranssen (approved):
LGTM 🚀

faisal-memon (approved):
Thanks @kfox1111 for all the automation work.

PR #467, opened 2023-08-22
Split steps in check-versions wf for easier debugging

marcofranssen:


faisal-memon (approved):
Thanks @marcofranssen for the cleanup.

PR #471, opened 2023-08-30
Switch mysql and postgresql tests to HA Production configs

kfox1111:


examples/external-mysql/run-tests.sh
```diff
@@ -12,31 +12,35 @@ source "${SCRIPTPATH}/../../.github/scripts/parse-versions.sh"
...
-  kubectl delete ns mysql 2>/dev/null || true
+  kubectl delete ns spire-server 2>/dev/null || true
+  kubectl delete ns spire-system 2>/dev/null || true
 }
 
 trap 'trap - SIGTERM && teardown' SIGINT SIGTERM EXIT
 
-"${helm_install[@]}" mysql mysql --version "$VERSION_MYSQL" --repo "$HELM_REPO_MYSQL" \
-  --namespace mysql \
+kubectl create namespace spire-system --dry-run=client -o yaml | kubectl apply -f -
```
faisal-memon, line 24: Why not just `kubectl create namespace spire-system`?

kfox1111: Idempotent. If it exists, it leaves it alone. If it doesn't, it creates it.

examples/external-mysql/run-tests.sh
```diff
@@ -12,31 +12,35 @@ source "${SCRIPTPATH}/../../.github/scripts/parse-versions.sh"
...
+kubectl create namespace spire-server --dry-run=client -o yaml | kubectl apply -f -
+kubectl label namespace spire-server pod-security.kubernetes.io/enforce=restricted || true
+
+helm upgrade --install --create-namespace mysql mysql --version "$VERSION_MYSQL" --repo "$HELM_REPO_MYSQL" \
+  --namespace spire-server \
   --values "${DEPS}/mysql.yaml" \
   --wait
 
-"${helm_install[@]}" --namespace "${ns}" --values "${SCRIPTPATH}/values.yaml" \
+helm upgrade --install --create-namespace --namespace "spire-server" \
```
faisal-memon, line 34: The name space is already created above, probably do we need the extra `--create-namespace` here?

kfox1111: yeah, I can remove that.

examples/external-mysql/run-tests.sh
faisal-memon: `kubectl create` is not idempotent?

examples/external-mysql/run-tests.sh
kfox1111: Done

examples/external-mysql/run-tests.sh
kfox1111: No. it fails if it already exists. :/

```
$ kubectl create namespace foo
namespace/foo created
$ kubectl create namespace foo
Error from server (AlreadyExists): namespaces "foo" already exists
$ echo $?
1
```

create to apply is a common pattern to work around it.

examples/external-mysql/run-tests.sh
```diff
@@ -12,31 +12,35 @@ source "${SCRIPTPATH}/../../.github/scripts/parse-versions.sh"
...
+kubectl label namespace spire-server pod-security.kubernetes.io/enforce=restricted || true
+
+helm upgrade --install mysql mysql --version "$VERSION_MYSQL" --repo "$HELM_REPO_MYSQL" \
+  --namespace spire-server \
   --values "${DEPS}/mysql.yaml" \
   --wait
 
-"${helm_install[@]}" --namespace "${ns}" --values "${SCRIPTPATH}/values.yaml" \
+helm upgrade --install --namespace "spire-server" \
+  --values "${SCRIPTPATH}/values.yaml,${SCRIPTPATH}/../production/values.yaml,${SCRIPTPATH}/../production/values-node-pod-antiaffinity.yaml" \
```
faisal-memon, line 35: Previously used spire-system but now you are switching to spire-server. Is spire-system still being used here?

kfox1111: All the privileged system level bits go in spire-system. all the rest of the spire-server bits go into spire-server.

examples/external-mysql/run-tests.sh
faisal-memon: Got it, thanks.

examples/external-mysql/run-tests.sh
faisal-memon: Youre using the `namespaceOverride` for the multi namespace override. Was missing that part.

mrsabath (approved):
Nice addition @kfox1111 Thank you! 
LGTM

PR #477, opened 2023-09-06
Use deploy key in actions that create PRs

### LINKED ISSUE #334: Test not running for automated PRs
When creating automated PRs from the cron jobs that checks versions, the tests arent being run automatically. We have to push a commit to the branch to force the tests to run.

faisal-memon:
Following steps in this [medium article](https://medium.com/prompt/trigger-another-github-workflow-without-using-a-personal-access-token-f594c21373ef) to hopefully fix issue where tests don't run for auto created PRs. Alternative approach is to create a bot account and use a PAT from that. 

fixes #334 (hopefully)

### DIFF
--- a/.github/workflows/check-versions.yaml
+++ b/.github/workflows/check-versions.yaml
@@ -21,6 +21,8 @@ jobs:
     steps:
       - name: Checkout
         uses: actions/checkout@v4.0.0
+        with:
+          ssh-key: "${{ secrets.COMMIT_KEY }}"
 
       - name: Set up Helm
         uses: azure/setup-helm@v3.5
--- a/.github/workflows/check-versions.yaml
+++ b/.github/workflows/check-versions.yaml
@@ -64,7 +64,6 @@ jobs:
         id: cpr
         uses: peter-evans/create-pull-request@v5.0.2
         with:
-          token: ${{ secrets.GITHUB_TOKEN }}
           title: Bump test chart dependencies
           branch: bump-test-chart-deps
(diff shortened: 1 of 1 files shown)


kfox1111 (approved):
I think this might work. Won't know for sure until we merge I think.

kfox1111:
> Why are we doing this? This is an anti pattern as it circumvents the permissions system in GH workflows.
> 
> Normally you just need to do following:
> 
> ```
> permissions:
>   contents: write
> ```
> 
> This will allow the workflow to make a commit so we don't have to manage our own keys.

Because the github workflows are broken with regard to submitting pr's that actually have their tests run automatically. :/

marcofranssen:
Did some additional research.

https://github.com/peter-evans/create-pull-request/blob/main/docs/concepts-guidelines.md#triggering-further-workflow-runs

1. Seems if we want to do with deploy keys all workflows need to have the push event added as well in this PR. That also means it is a different workflow run and therefore also a different mandatory check, which complicates what we are doing.

2. Another good option to me is the GitHub app option which also works on the pull_request event, so we don't have to change the workflow triggers and the mandotory checks in branch protection.

As a workaround to trigger the workflow they also suggest to `close` and `reopen` the PR which also triggers the workflows, which is 2 manual clicks on these PRs. This would also allow us to stick with current merge checks as we don't have to change workflow triggers and such.

faisal-memon:
Thanks @marcofranssen for doing the research. Seems like this option won't work. Will look into the github app option.

PR #480, opened 2023-09-08
Allow configuration of priorityClassName on spire-server statefulset

InverseIntegral:
In #58 it was made possible to configure the `priorityClassName` for `Daemonsets`. With this PR it's now possible to the same for the `Statefulset` that is deployed as part of `spire-server`.

### DIFF
--- a/examples/production/values.yaml
+++ b/examples/production/values.yaml
@@ -23,6 +23,7 @@ spire-server:
       drop: [ALL]
     seccompProfile:
       type: RuntimeDefault
+  priorityClassName: system-cluster-critical
 
   logLevel: info
 
--- a/charts/spire/charts/spire-server/README.md
+++ b/charts/spire/charts/spire-server/README.md
@@ -81,7 +81,7 @@ curl  --cacert CA/rootCA.crt --key client.key --cert client.crt https://localhos
 
 In order to run Tornjak with simple HTTP Connection only, make sure you don't create any `Secrets` or `ConfigMaps` listed above.
 
-## Parameters 
+## Parameters
 
 ### Chart parameters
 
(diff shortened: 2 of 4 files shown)


examples/production/values.yaml
```diff
@@ -23,6 +23,7 @@ spire-server:
       drop: [ALL]
     seccompProfile:
       type: RuntimeDefault
+  priorityClassName: system-node-critical
```
marcofranssen, line 26: Why would the server be critical to a node? A common pattern is to do this for DaemonSets to make room for them on a node, I don't understand why we would want to do this for this statefulset which in essence does not require to run on a specific node.

What is the usecase for this? What are your trying to achieve?

InverseIntegral: Maybe the example that I chose was not optimal. Instead, I should have chosen `system-cluster-critical`.

The idea is that `spire-server` is critical to our cluster and should not be evicted before other, less critical, pods. If we were unable to set the `priorityClassName`, it could happen that the `spire-server` gets evicted and it would no longer be possible to start new pods since they need the `spire-server` to get their SVIDs.

marcofranssen: 💡 that makes more sense, thanks for clarifying the usecase.

Could you update the PR with that change?

marcofranssen:
Please note we recently switched to a different documentation tool. Could you resolve the conflicts and rerun the docs script?

Sorry for the inconvenience.

examples/production/values.yaml
InverseIntegral: Sure thing, it should be updated now :+1:

InverseIntegral:
@marcofranssen the documentation should now also be up-to-date using the new documentation tooling

kfox1111 (approved):
LGTM! Thanks! :)

marcofranssen (approved):
Thanks for the effort. LGTM

PR #492, opened 2023-09-21
Add support for experimental flags

faisal-memon:
Add support for the experimental section of the config. Needed for developers. This PR doesnt add support for the `auth_opa_policy_engine` experimental config.


marcofranssen (approved):
LGTM 🚀

charts/spire/charts/spire-server/values.yaml
```diff
@@ -628,6 +628,12 @@ customPlugins:
   upstreamAuthority: {}
   notifier: {}
 
+experimental:
+  ## @param experimental.cacheReloadInterval The amount of time between two reloads of the in-memory entry cache.
+  cacheReloadInterval: 5s
```
kfox1111, line 633: This causes it to always be set in the config file. Seems like it should't be set in the config file until explicitly set by the user since experimental features most likely shoulnd't be enabled in prod?

kfox1111 (commented):
Maybe a comment in the generated notes like:
https://github.com/spiffe/helm-charts/blob/main/charts/spire/templates/NOTES.txt#L6-L7

stating something like warning, experimental flags detected?

charts/spire/charts/spire-server/values.yaml
faisal-memon: Good point. What do you think of an `enabled` flag for this section?

kfox1111: Yeah, I think that would probably work. Makes it explicit that the user is intentionally using it.

faisal-memon:
> Maybe a comment in the generated notes like: https://github.com/spiffe/helm-charts/blob/main/charts/spire/templates/NOTES.txt#L6-L7
> 
> stating something like warning, experimental flags detected?

warning added

charts/spire/charts/spire-server/values.yaml
faisal-memon: enabled flag added

kfox1111 (approved):
LGTM

